PR #3, opened 2024-08-29
NOSTASK: fix bug with nil telemetry

Uburro:


helm/slurm-cluster/values.yaml
```diff
@@ -178,7 +178,16 @@ slurmNodes:
...
+telemetry: {}
+  # jobsTelemetry:
+  #   otelCollectorHttpHost: vmsingle-slurm.monitoring-system.svc.cluster.local
+  #   otelCollectorPath: /opentelemetry/api/v1/push
+  #   otelCollectorPort: 8429
+  #   sendJobsEvents: true
+  #   sendOtelMetrics: true
+  # prometheus:
+  #   enabled: true
+  #   imageSlurmExporter: cr.nemax.nebius.cloud/crnlu9nio77sg3p8n5bi/exporter:1.9.2-jammy-slurm24.05.2-185df658
```
dstaroff, line 194: This field doesn't seem to correspond to anything

helm/slurm-cluster/values.yaml
```diff
@@ -178,7 +178,16 @@ slurmNodes:
 #        - name: "mlcommons-sd-bench-data"
 #          mountPath: "/mlperf-sd"
 #          volumeSourceName: "mlperf-sd"
-
+telemetry: {}
+  # jobsTelemetry:
+  #   otelCollectorHttpHost: vmsingle-slurm.monitoring-system.svc.cluster.local
+  #   otelCollectorPath: /opentelemetry/api/v1/push
+  #   otelCollectorPort: 8429
```
dstaroff, line 185: Will it work without specifying `openTelemetryCollector`?

Uburro: yes, it will

PR #6, opened 2024-09-03
MSP-2748 Allow NodePort configuration

dstaroff:


helm/slurm-cluster/templates/slurm-cluster-cr.yaml
```diff
@@ -128,12 +128,13 @@ spec:
       sshRootPublicKeys:
         {{- default list .Values.slurmNodes.login.sshRootPublicKeys | toYaml | nindent 8 }}
       sshdServiceType: {{ default "LoadBalancer" .Values.slurmNodes.login.sshdServiceType | quote }}
-      sshdServiceAnnotations:
-        external-dns.alpha.kubernetes.io/hostname: "{{ include "slurm-cluster.name" . }}.slurm-operator.msp-dev.nemax.nebius.cloud."
+      sshdServiceAnnotations: {}
+      {{- /*external-dns.alpha.kubernetes.io/hostname: "{{ include "slurm-cluster.name" . }}.slurm-operator.msp-dev.nemax.nebius.cloud."*/ -}}
```
rdjjke, line 132: Why is it commented? Should it be conditional?

dstaroff: Right now there is no possibility to create DNS records in Newbius. Especially, via service annotations.
We will tidy this up when it's possible

api/v1/slurmcluster_types.go
```diff
@@ -405,6 +406,12 @@ type SlurmNodeLogin struct {
 	// +kubebuilder:validation:Optional
 	SshdServiceLoadBalancerIP string `json:"sshdServiceLoadBalancerIP,omitempty"`
 
+	// SshdServiceNodePort represents the port to be opened on nodes in case of NodePort type of service
+	//
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=0
```
asteny, line 412: Why port is 0? What does it mean?

dstaroff: It's used in validation to be sure it was set if service type is `NodePort`

helm/slurm-cluster/templates/slurm-cluster-cr.yaml
rdjjke: Ah, OK, got it

PR #9, opened 2024-09-03
MSP-2766: add helm chart for crds

Uburro:
https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-2-separate-charts

### DIFF
--- a/helm/slurm-operator-crds/Chart.yaml
+++ b/helm/slurm-operator-crds/Chart.yaml
@@ -13,9 +13,9 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 0.1.0
+version: 1.11.3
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -127,6 +128,7 @@ helm: kustomize helmify yq ## Update slurm-operator Helm chart
 	rm -rf $(CHART_OPERATOR_PATH)
 	$(KUSTOMIZE) build config/default | $(HELMIFY) --crd-dir $(CHART_OPERATOR_PATH)
 	rm -f $(CHART_PATH)/operatorAppVersion
+	cp -r $(CHART_OPERATOR_PATH)/crds/* $(CHART_OPERATOR_CRDS_PATH)/templates/
```
dstaroff, line 131: Should we also remove CRD from `slurm-operator` chart?

Uburro: No, if we use FluxCD or ArgoCD, we won't encounter the same issues as with Helm. This is for those who use Helm without wrappers

PR #13, opened 2024-09-05
MSP-2775: move creating cgroupv2 from init container to entrypoint worker

Uburro:
move creating cgroupv2 from init container to entrypoint worker


helm/slurm-operator-crds/Chart.yaml
```diff
@@ -13,9 +13,9 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
-version: 1.11.3
+version: 1.11.3-9a6c3bf3
```
asteny: you should remove hash

Uburro: done

PR #14, opened 2024-09-05
MSP-2764: add cel validation for jobtelemetry and shared memory

Uburro:


api/v1/slurmcluster_types.go
```diff
@@ -333,6 +333,7 @@ type SlurmNodeControllerVolumes struct {
 // If CgroupVersion is "v2", InitContainer must not be null
```
asteny: I think that comments should be remover and all comments about InitContainer

Uburro: it has already removed in previous PR

asteny: so, you need to rebase?

Uburro: done =)

PR #16, opened 2024-09-06
MSP-2788: moved pod template from common to prometheus package

Uburro:
pod.go use specific values for exporter. Not for commom pod template

### DIFF
--- a/internal/render/prometheus/exporter.go
+++ b/internal/render/prometheus/exporter.go
@@ -55,7 +55,7 @@ func RenderDeploymentExporter(
 			Selector: &metav1.LabelSelector{
 				MatchLabels: matchLabels,
 			},
-			Template: common.RenderPodTemplateSpec(
+			Template: RenderPodTemplateSpec(
 				clusterName,
 				&valuesExporter.ContainerMunge,
 				valuesExporter,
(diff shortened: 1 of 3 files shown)


internal/render/prometheus/exporter.go
```diff
@@ -55,7 +55,7 @@ func RenderDeploymentExporter(
 			Selector: &metav1.LabelSelector{
 				MatchLabels: matchLabels,
 			},
-			Template: common.RenderPodTemplateSpec(
```
asteny, line 58: I'm not sure, that I understand why you are moving it?

internal/render/prometheus/pod_test.go
```diff
@@ -1,4 +1,4 @@
-package common_test
+package prometheus_test
```
asteny, line 1: it should be prometheus, I think...
Am I right?

Uburro: No, I am specifically trying to move tests to a separate package whenever possible. @dstaroff  and I discussed this.

internal/render/prometheus/exporter.go
Uburro: pod.go use specific values for exporter. Not for common pod template.

Uburro: RenderPodTemplateSpec is func in pod.go. it has moved.

PR #23, opened 2024-09-11
MSP-2788: add accounting to consts and values

Uburro:


api/v1/slurmcluster_types.go
```diff
@@ -295,6 +300,70 @@ type SlurmNodes struct {
...
+	// +kubebuilder:default=3306
+	Port int32 `json:"port"`
+	// Key defines the key of username and password in the secret
+	//
+	// +kubebuilder:validation:Optional
+	User string `json:"user"`
+	// Secret for connection string to the SlurmDBD database
+	//
+	// +kubebuilder:validation:Optional
+	Secret SecretName `json:"secret"`
```
asteny: Maybe just Secret?

api/v1/slurmcluster_types.go
```diff
@@ -295,6 +300,70 @@ type SlurmNodes struct {
...
+}
+
+// SecretName defines the name of the secret
```
asteny: It not defines only name, so Secret?

api/v1/slurmcluster_types.go
```diff
@@ -329,7 +398,6 @@ type SlurmNodeControllerVolumes struct {
 }
 
 // SlurmNodeWorker defines the configuration for the Slurm worker node
-// +kubebuilder:validation:XValidation:rule="self.slurmd.resources.memory > self.volumes.sharedMemorySize",message="resources memory of slurmd must be bigger than sharedMemorySize"
```
asteny, line 332: Why u delete this validation?

internal/naming/naming.go
```diff
@@ -80,6 +80,14 @@ func BuildStatefulSetName(componentType consts.ComponentType, clusterName string
 	}.String()
 }
 
+func BuildDeploymentName(componentType consts.ComponentType, clusterName string) string {
```
asteny, line 83: U do not use clusterName here

internal/consts/accounting.go
```diff
@@ -0,0 +1,11 @@
+package consts
+
+const (
+	SecretSlurmdbdConfigs       = "slurmdbd-configs"
+	HostnameAccounting          = "accounting"
+	DbdPort                     = "6819"
```
asteny: Why we have 2 vars with port and using only one?

internal/consts/annotation.go
```diff
@@ -4,5 +4,6 @@ const (
 	AnnotationVersions = "versions"
 
 	AnnotationApparmorKey             = "container.apparmor.security.beta.kubernetes.io"
+	DefaultContainerAnnotationName    = "kubectl.kubernetes.io/default-container"
```
asteny, line 7: U want use this annotation in all our containers?

Uburro: Yes, right now default container is munge. It's inconvenient to specify a special container every time.

api/v1/slurmcluster_types.go
Uburro: Cel validate int, not resources of memory, this leads to an error

internal/consts/annotation.go
asteny: Does that mean that next PR will be about changing annotations for all containers?

Uburro: Not next, but after accounting merged - yes

PR #50, opened 2024-09-20
[MSP-2905] Support clusters without GPUs

asteny:


images/worker/slurmd_entrypoint.sh
```diff
@@ -47,15 +42,26 @@ echo "Complement jail rootfs"
...
+    # See Slurm docs: https://slurm.schedmd.com/gres.html#AutoDetect
+    GRES="$(nvidia-smi --query-gpu=name --format=csv,noheader | sed -e 's/ /_/g' -e 's/.*/\L&/' | sort | uniq -c | awk '{print "gpu:" $2 ":" $1}' | paste -sd ',' -)"
+    echo "Detected GRES is $GRES"
+
+    echo "Create NVML symlink with the name expected by Slurm"
+    pushd /usr/lib/x86_64-linux-gnu
+        ln -s libnvidia-ml.so.1 libnvidia-ml.so
+    popd
+else
+    echo "Detect available CPUs"
```
rdjjke: The message looks wrong here

rdjjke: Ah, OK, just read it as "GPUs", not "CPUs".
Anyway, it's not only about CPUs.

What about printing "Skip GPU detection"?

internal/render/common/configmap.go
```diff
@@ -164,9 +165,13 @@ func generateSpankConfig() renderutils.ConfigFile {
 	return res
 }
 
-func generateGresConfig() renderutils.ConfigFile {
+func generateGresConfig(clusterType consts.ClusterType) renderutils.ConfigFile {
 	res := &renderutils.PropertiesConfig{}
-	res.AddProperty("AutoDetect", "nvml")
+	res.AddComment("Gres config")
+	if clusterType == consts.ClusterTypeGPU {
+		res.AddProperty("GresTypes", "gpu")
```
rdjjke: "GresTypes" option should be in "slurm.conf", not "gres.conf"

helm/slurm-cluster/templates/slurm-cluster-cr.yaml
```diff
@@ -14,6 +14,7 @@ metadata:
 spec:
   crVersion: {{ .Chart.Version }}
   pause: {{ .Values.pause }}
+  clusterType: {{ .Values.clusterType }}
```
rdjjke, line 17: Maybe we can just use the number of "nvidia.com/gpu" instead of introducing a separate field?

rdjjke: I'm asking because we'll definitely add support for separate "NodeSets", where nodes in each set may have different resources. So we'll have to drop this field.

asteny: I think number of "nvidia.com/gpu" is very complicated logic... and "nvidia.com/gpu" number can be valid only for Nebius... Lets drop that field, with NodeSets?

rdjjke: OK

rdjjke (approved):

Uburro (approved):

PR #56, opened 2024-09-23
[MSP-2928] Push docker images into github

asteny:


helm/slurm-operator/values.yaml
```diff
@@ -34,7 +34,7 @@ controllerManager:
       isPrometheusCrdInstalled: "false"
       slurmOperatorWatchNamespaces: '*'
     image:
-      repository: cr.nemax.nebius.cloud/crnlu9nio77sg3p8n5bi/slurm-operator
+      repository: ghcr.io/nebius/soperator/slurm-operator
```
dstaroff, line 37: Do we want to use `soperator/slurm-operator` name instead of `soperator/soperator`?

dstaroff (changes requested):

helm/slurm-operator/values.yaml
asteny: That's a lot of changes (((

dstaroff: ok

dstaroff (approved):

PR #58, opened 2024-09-23
MSP-2790: add pv for filestore storage of accounting 

Uburro:


api/v1/slurmcluster_types.go
```diff
@@ -353,16 +353,17 @@ type PasswordSecretKeyRef struct {
 	Key string `json:"key"`
 }
 
-type MariaDbOpeator struct {
+type MariaDbOperator struct {
 	// +kubebuilder:validation:Optional
```
dstaroff, line 357: This validation marker does nothing in terms of validation. It's just a decorative stuff, and redundant in case of controller-gen usage.
https://book.kubebuilder.io/reference/markers#difference-between--optional-and--kubebuildervalidationoptional

If you want a field to be optional, you have to specify `omitempty` in json tags as well

api/v1/slurmcluster_types.go
```diff
@@ -310,10 +310,10 @@ type SlurmNodeAccounting struct {
 	//
 	// +kubebuilder:validation:Optional
 	ExternalDB ExternalDB `json:"externalDB"`
-	// MariaDbOpeator represents the MariaDB CRD configuration
+	// MariaDbOperator represents the MariaDB CRD configuration
 	//
 	// +kubebuilder:validation:Optional
-	MariaDbOpeator MariaDbOpeator `json:"mariaDbOpeator"`
+	MariaDbOperator MariaDbOperator `json:"MariaDbOperator"`
```
dstaroff, line 316: JSON tag became Pascal cased after the replace

asteny (approved):

helm/slurm-cluster-storage/values.yaml
```diff
@@ -17,6 +17,11 @@ volume:
     name: "controller-spool"
     size: "30Gi"
     filestoreDeviceName: "controller-spool"
+  accounting:
+    enabled: true
```
dstaroff, line 21: Why is accounting enabled by default during storage creation while it's by default disabled during cluster creation?

helm/slurm-cluster-storage/templates/_helpers.tpl
```diff
@@ -106,6 +106,40 @@
...
+
+{{/* Controller spool size */}}
+{{- define "slurm-cluster-storage.volume.accounting.size" -}}
+    {{- required "Controller spool volume size is required." .Values.volume.accounting.size -}}
+{{- end }}
+
+{{/* Controller spool device name */}}
+{{- define "slurm-cluster-storage.volume.accounting.device" -}}
+    {{- required "Controller spool Filestore device name is required." .Values.volume.accounting.filestoreDeviceName | trim | kebabcase -}}
+{{- end }}
```
dstaroff, line 137: Comments and `required` messages refer to controller spool instead of accounting

dstaroff (changes requested):

PR #60, opened 2024-09-23
Msp 2791 rebase

Uburro:


helm/slurm-cluster-storage/templates/_helpers.tpl
```diff
@@ -106,6 +106,40 @@
 ---
 */}}
 
+{{/* Controller accounting volume */}}
```
dstaroff: `Controller` prefix is still present. I suggest replacing it with `Accounting database`

dstaroff (changes requested):

dstaroff (approved):

PR #121, opened 2024-10-16
Fixes for version check and yq installation

asteny:


.github/workflows/github_release.yml
```diff
@@ -79,6 +79,13 @@ jobs:
                   "target": "$1"
                 }
               ],
+              "tag_resolver": {
+                "method": "semver",
+                "filter": {
+                  "pattern": "^\d+\.\d+\.\d+",
```
dstaroff: Maybe we can also add `v` at the beginning?
```suggestion
                  "pattern": "^v?\d+\.\d+\.\d+",
```

dstaroff: However, it's not needed if we will use `v`-less tags

asteny: Do you want continue using tags with v?

dstaroff (approved):

PR #125, opened 2024-10-17
FIX: Use empty string as default for `sshdServiceLoadBalancerIP`

dstaroff:


Makefile
```diff
@@ -210,6 +210,9 @@ sync-version: yq ## Sync versions from file
 	@echo ')'                                               >> $(GO_CONST_VERSION_FILE)
 	@# endregion internal/consts
 
+.PHONY: sync-version-from-scratch
+sync-version-from-scratch: generate manifests helm sync-version ## Regenerates all resources and syncs versions to them
```
asteny, line 214: Do we need helmify here?
`make generate manifests helm helmify sync-version`

asteny (approved):

Makefile
dstaroff: `helmify` target ensures that you have [helmify](https://github.com/arttor/helmify) installed.
Then it's used in `helm` target to transform Kustomize output into Helm chart.
BTW, `helmify` target is a dependency of `helm` target.

PR #131, opened 2024-10-18
feature: slurm partition configuration

asteny:


api/v1/slurmcluster_types.go
```diff
@@ -68,6 +68,25 @@ type SlurmClusterSpec struct {
...
+}
+
+type PartitionConfiguration struct {
+	// ConfigType
+	// +kubebuilder:validation:Enum=default;custom
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default="default"
+	ConfigType string `json:"configType,omitempty"`
+	// RawConfig define partition configuration as raw multi string started with PartitionName
+	// Example:
```
rdjjke: This example is odd. Aren't PriorityTiers used for overlapping partitions?

config/crd/bases/slurm.nebius.ai_slurmclusters.yaml
```diff
@@ -1083,6 +1083,26 @@ spec:
...
+                    default: default
+                    description: ConfigType
+                    enum:
+                    - default
+                    - custom
+                    type: string
+                  rawConfig:
+                    description: |-
+                      RawConfig define partition configuration as raw multi string started with PartitionName
+                      Example:
```
rdjjke: Same comment as [here](https://github.com/nebius/soperator/pull/131/files#diff-9b6e2902200568f146ded160389e94de6f824f6e4a7c2f89ac86ab14c0062048R85)

helm/slurm-cluster/values.yaml
```diff
@@ -5,6 +5,15 @@ annotations: {}
...
 # Slurm cluster type. Can be now gpu or cpu
 clusterType: gpu
+# partitionConfiguration define partition configuration of slurm worker nodes
+# https://slurm.schedmd.com/slurm.conf.html#SECTION_PARTITION-CONFIGURATION
+partitionConfiguration:
+  # Could be default or custom
+  configType: "default"
+  # configuration as raw multi string started with PartitionName
+  # Example:
+  # PartitionName=low_priority Nodes=worker-[0-9] Default=YES MaxTime=INFINITE State=UP PriorityTier=1
```
rdjjke: Why the default partition is called "low_priority"?

helm/soperator-crds/templates/slurmcluster-crd.yaml
```diff
@@ -1082,6 +1082,26 @@ spec:
...
+                    description: ConfigType
+                    enum:
+                    - default
+                    - custom
+                    type: string
+                  rawConfig:
+                    description: |-
+                      RawConfig define partition configuration as raw multi string started with PartitionName
+                      Example:
+                      PartitionName=low_priority Nodes=worker-[0-9] Default=YES MaxTime=INFINITE State=UP PriorityTier=1
```
rdjjke: Same comment as [here](https://github.com/nebius/soperator/pull/131/files#diff-9b6e2902200568f146ded160389e94de6f824f6e4a7c2f89ac86ab14c0062048R85)

helm/soperator/crds/slurmcluster-crd.yaml
```diff
@@ -1082,6 +1082,26 @@ spec:
...
+                    description: ConfigType
+                    enum:
+                    - default
+                    - custom
+                    type: string
+                  rawConfig:
+                    description: |-
+                      RawConfig define partition configuration as raw multi string started with PartitionName
+                      Example:
+                      PartitionName=low_priority Nodes=worker-[0-9] Default=YES MaxTime=INFINITE State=UP PriorityTier=1
```
rdjjke: Same comment as [here](https://github.com/nebius/soperator/pull/131/files#diff-9b6e2902200568f146ded160389e94de6f824f6e4a7c2f89ac86ab14c0062048R85)

internal/values/validate.go
```diff
@@ -18,6 +20,21 @@ import (
 // Returns true if valid. Otherwise, false.
 func (c *SlurmCluster) Validate(ctx context.Context) error {
 	logger := log.FromContext(ctx)
+	// PartitionConfiguration
+	{
+		if c.PartitionConfiguration.ConfigType == "custom" {
+			scanner := bufio.NewScanner(strings.NewReader(c.PartitionConfiguration.RawConfig))
+			for scanner.Scan() {
+				line := scanner.Text()
+				if !strings.HasPrefix(line, "PartitionName") {
```
rdjjke, line 27: Shouldn't we skip empty lines? (or lines consisting of blank characters)

internal/render/common/configmap.go
```diff
@@ -113,7 +115,21 @@ func generateSlurmConfig(cluster *values.SlurmCluster) renderutils.ConfigFile {
...
+	res.AddProperty("JobRequeue", 1)
+	res.AddProperty("PreemptMode", "REQUEUE")
+	res.AddProperty("PreemptType", "preempt/partition_prio")
+	switch cluster.PartitionConfiguration.ConfigType {
+	case "default":
+		res.AddProperty("PartitionName", "main Nodes=ALL Default=YES MaxTime=INFINITE State=UP OverSubscribe=YES")
+	case "custom":
+		scanner := bufio.NewScanner(strings.NewReader(cluster.PartitionConfiguration.RawConfig))
+		for scanner.Scan() {
+			line := scanner.Text()
```
rdjjke: Same comment as [here](https://github.com/nebius/soperator/pull/131/files#diff-82653c79336d19d9bddd13c6b6749ad837799b6bd778a83d6d72e5fb96104262R29)

rdjjke (changes requested):

internal/values/validate.go
asteny: why do u want to have empty lines here?

rdjjke: If the partition configuration is long, the user may want to split it into blocks separated by empty lines

rdjjke: Maybe we should also ignore comment lines

helm/slurm-cluster/values.yaml
asteny: Because in this example u have 2 partitions high and low, and by default u will use low. Which names do u want see here?

api/v1/slurmcluster_types.go
asteny: Ok, I changed it

helm/slurm-cluster/values.yaml
rdjjke: Because this is an example of the "default" config. I expect a name like "default" or "main" (we have "main" now while it's not configurable)

asteny: But that's example of not default config) If configType - default doesn't matter what u will put on rawConfig. Ok I'll change comment here: `Example for custom ConfigType`

internal/values/validate.go
asteny: Changed. Now only if string started with PartitionName we will put it in config

internal/render/common/configmap.go
```diff
@@ -113,7 +115,24 @@ func generateSlurmConfig(cluster *values.SlurmCluster) renderutils.ConfigFile {
...
+	res.AddProperty("PreemptType", "preempt/partition_prio")
+	switch cluster.PartitionConfiguration.ConfigType {
+	case "default":
+		res.AddProperty("PartitionName", "main Nodes=ALL Default=YES MaxTime=INFINITE State=UP OverSubscribe=YES")
+	case "custom":
+		scanner := bufio.NewScanner(strings.NewReader(cluster.PartitionConfiguration.RawConfig))
+		for scanner.Scan() {
+			line := strings.TrimSpace(scanner.Text())
+
+			if strings.HasPrefix(line, "PartitionName") {
```
rdjjke, line 127: Why there is no validation anymore?

I think the logic should be such:
- Lines that start with "PartitionName" are partition configuration that is inserted into the slurm.conf
- Empty lines, lines of blank chars, and lines starting with "#" are ignored.

We should fail if a user tries to configure "PreemptMode" or anything else except "PartionName=....". Now such settings are just silently ignored, and I don't think it's a good idea.

api/v1/slurmcluster_types.go
```diff
@@ -68,6 +68,25 @@ type SlurmClusterSpec struct {
...
+	// +kubebuilder:validation:Enum=default;custom
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default="default"
+	ConfigType string `json:"configType,omitempty"`
+	// RawConfig define partition configuration as raw multi string started with PartitionName
+	// Example for custom ConfigType:
+	// PartitionName=low_priority Nodes=worker-[0-15] Default=YES MaxTime=INFINITE State=UP PriorityTier=1
+	// PartitionName=high_priority  Nodes=worker-[10-20] Default=NO MaxTime=INFINITE State=UP PriorityTier=2
+	// +kubebuilder:validation:Optional
+	RawConfig string `json:"rawConfig,omitempty"`
```
dstaroff: Maybe it is better to have it as `[]string` instead of multi-line string?

internal/render/common/configmap.go
asteny: You wrote exactly that I did in code) 

Why we should fail? We wrote example that config lines should have PartionName in the start of the line. And only them we get from config. How you want validate another string?

dstaroff (changes requested):

internal/render/common/configmap.go
rdjjke: OK, but what if the user writes "PartirionName=..." (with a mistake) -- this line will be ignored.

api/v1/slurmcluster_types.go
dstaroff: It gives:
- You won't need `scanner` anymore
  - As simple as `for range` loop
  - Line processing becomes more clear
- You can just add comment lines between config strings to separate blocks like it's said [here](https://github.com/nebius/soperator/pull/131#discussion_r1806632926)

internal/render/common/configmap.go
asteny: So, what will be if user write #PartirionName=..? 
Next question: where is better place for fail? In operator (here you need read logs of operator in order to understand what happen and where is an error) or in slurm (read that in sinfo command or in slurm log)?

rdjjke (approved):

rdjjke (approved):

Uburro (approved):

PR #151, opened 2024-10-29
perf(images): use multi-stage to reduce populate_jail image size

CrackedPoly:
After using multiple stages in docker build, we can reduce populate_jail image from ~20G to ~10G.

You should merge this as the size is laaaaarge.


images/populate_jail/populate_jail.dockerfile
```diff
@@ -1,14 +1,18 @@
 ARG BASE_IMAGE=ubuntu:jammy
 
+# Frist stage: untap jail_rootfs.tar
```
asteny: First

asteny (changes requested):
Hello! Thanks for your contribution. Please fix comment and lets merge it. I changed the branch to dev, because we have release process based on dev branch. We collecting changes in dev branch and then merge them all in main and build stable version of soperator. So, you changes will be in next release.

images/populate_jail/populate_jail.dockerfile
CrackedPoly: Done.

asteny (approved):

PR #152, opened 2024-10-30
feat: support ImagePullPolicy from manifest

CrackedPoly:
Hi people, I have added an extra field `ImagePullPolicy` in specs, so we can avoid always pulling images from the cloud.


Uburro:
LGTM 
@CrackedPoly thanks!

Uburro (approved):

PR #161, opened 2024-11-04
hotfix: fix required Limit must be set for non overcommitable 

Uburro:
fix required value Limit must be set for non overcommitable  resource and if limits empty  pass to empty resource list

### DIFF
--- a/internal/render/common/limits.go
+++ b/internal/render/common/limits.go
@@ -1,0 +1,18 @@
+package common
+
+import (
+	corev1 "k8s.io/api/core/v1"
+)
+
+func CopyNonCPULimits(resourceList corev1.ResourceList) corev1.ResourceList {
+	// Create a copy of all the resources except CPU.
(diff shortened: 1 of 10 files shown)


asteny (dismissed):

internal/render/common/limits.go
```diff
@@ -0,0 +1,18 @@
+package common
+
+import (
+	corev1 "k8s.io/api/core/v1"
+)
+
+func CopyNonCPULimits(resourceList corev1.ResourceList) corev1.ResourceList {
```
rdjjke: What about naming it "ExcludeCPU" or "CopyNonCPUResources? "CopyLimits" is used a little bit wrongly here IMO, because it takes not "limits", but "resources".

The fact that excluding CPU from resources may be useful for setting container limits can be mentioned in the comment to this function (not in a comment inside the function body).

rdjjke (approved):

PR #164, opened 2024-11-04
MSP-3261: add apparmor profile

Uburro:
Add an AppArmor profile to prevent the modification of libnvidia libraries to versions not installed by the NVIDIA operator.


internal/render/worker/statefulset.go
```diff
@@ -110,3 +103,21 @@ func RenderStatefulSet(
...
 	}, nil
 }
+
+// BuildAnnotations returns a map of annotations for the Slurm worker pods
+func BuildAnnotations(clusterName string) map[string]string {
+	annotations := map[string]string{
+		consts.DefaultContainerAnnotationName: consts.ContainerNameSlurmd,
+	}
+
+	// Always set the annotation for ContainerNameMunge to unconfined
```
asteny, line 113: May be lets do that for all containers and remove that comment?

asteny:
Lets bump version to 1.15.1?

internal/render/worker/statefulset.go
Uburro: it has already unconfined for all containers except Slurmd in apapparmor profile case

PR #169, opened 2024-11-07
MSP-2609: add flux for local development soperator

Uburro:
values for resources and url will be add in next PR


flux/clusters/single-node/slurm-vars.yaml
```diff
@@ -0,0 +1,162 @@
...
+        requiredDuringSchedulingIgnoredDuringExecution:
+          nodeSelectorTerms:
+            - matchExpressions:
+                - key: kubernetes.io/os
+                  operator: In
+                  values:
+                    - linux
+
+volumeSources:
+- name: jail
```
asteny, line 21: tab?)

flux/clusters/single-node/apps-variables.ini
```diff
@@ -0,0 +1,20 @@
+soperator_suspended=false
+slurm_cluster_suspended=false
+cert_manager_suspended=false
+monitoring_suspended=false
+
+# version of the soperator to install
+soperator_version=1.15.1-8b29c649
```
asteny, line 7: May be we should add some version sync string to the Make file in order to update version here

asteny (changes requested):

PR #190, opened 2024-11-20
Set UseInfiniband defaul - true

asteny:


helm/slurm-cluster/values.yaml
```diff
@@ -105,9 +105,9 @@ periodicChecks:
       # NCCL timeout in its special format. By default, 20 minutes
       timeout: "20:00"
       # Threshold for benchmark result that must be guaranteed. CronJob will fail if the result is less than the threshold
-      thresholdMoreThan: "400"
+      thresholdMoreThan: "0"
```
rdjjke, line 108: It shouldn't be zero by default. Because it means that in any result, all nodes will be drained

rdjjke: Oh sorry I’m wrong. It’s actually vise versa

rdjjke (approved):

dstaroff (approved):

PR #212, opened 2024-12-03
MSP-3191: add webhook protect delete secret mariadb

Uburro:


api/v1/slurmcluster_types.go
```diff
@@ -441,6 +441,13 @@ type MariaDbOperator struct {
 	// +kubebuilder:validation:Optional
 	Enabled bool `json:"enabled"`
 
+	// If enabled, secret cannot be deleted until custom resource slurmcluster is deleted
+	//
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default="false"
```
asteny, line 447: May be it should be true?

Uburro: If set to true by default, it will break old clusters where the secret was generated by the MariaDB operator.

asteny: But without enabling the option, nobody will use it. May be we should manually recreate secrets for our clients? And write this information to release

Uburro: >  nobody will use it. May be we should manually recreate secrets for our clients? And write this information to release

why nobody? we will in new clusters

PR #228, opened 2024-12-09
Set compute instance name as slurm InstanceId

asteny:


images/worker/slurmd_entrypoint.sh
```diff
@@ -89,12 +89,15 @@ else
     echo "Skipping GPU detection"
 fi
 
+echo "Delete previous node config from memory if exists"
```
rdjjke: This is actually not from a memory, but from the controller's state, which is stored on a disk

rdjjke (dismissed):

Uburro (approved):

dstaroff (approved):

PR #231, opened 2024-12-11
MSP-3272: Support GDRCopy + preinstall more tools

rdjjke:


Uburro (dismissed):

Uburro (approved):

images/common/scripts/install_gdrcopy.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
asteny, line 1: https://github.com/NVIDIA/gdrcopy?tab=readme-ov-file#deb-package

rdjjke: It requires a driver package to be installed. But we can't do this in K8s, because this driver is propagated by nvidia GPU operator

asteny: Same as from source: ```Compiling the gdrdrv driver requires the NVIDIA driver source code, which is typically installed at /usr/src/nvidia-<version>. ```  https://github.com/NVIDIA/gdrcopy?tab=readme-ov-file#notes

rdjjke: This is about compiling the gdrdrv driver and that it needs source codes. But we don't compile/install the driver, it's propagated by GPU operator. We install gdrcopy libraries & test executables.
Deb packages for these libs & exes have a dependency on the package with gdrdrv driver, that we can't install inside k8s containers.

rdjjke: So, I tried to install it as you suggested, but it doesn't work.

rdjjke: This issue is about docker, not K8s, but anyway: https://github.com/NVIDIA/gdrcopy/issues/197

asteny (approved):

PR #240, opened 2024-12-11
MSP-3578: add ssh to slurm work

Uburro:


api/v1/slurmcluster_types.go
```diff
@@ -635,6 +639,18 @@ type SlurmNodeWorker struct {
 	CgroupVersion string `json:"cgroupVersion,omitempty"`
 }
 
+type SlurmNodeWorkerSshd struct {
+	// Enabled defines whether the SSH sidecar is enabled on the worker node
+	//
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=true
+	Enabled bool `json:"enabled,omitempty"`
+	// Sshd represents the SSH daemon service configuration
```
asteny: There is no Sshd configuration here...

internal/render/common/volume.go
```diff
@@ -300,3 +300,120 @@ func RenderVolumeMountRESTJWTKey() corev1.VolumeMount {
 }
 
 // endregion REST JWT key
+
+// region SSH
+
+// region configs
+
+// renderVolumeSshConfigs renders [corev1.Volume] containing SSH configs contents
```
asteny, line 308: sshd

internal/render/common/volume.go
```diff
@@ -300,3 +300,120 @@ func RenderVolumeMountRESTJWTKey() corev1.VolumeMount {
...
+			ConfigMap: &corev1.ConfigMapVolumeSource{
+				LocalObjectReference: corev1.LocalObjectReference{
+					Name: naming.BuildConfigMapSSHConfigsName(clusterName),
+				},
+			},
+		},
+	}
+}
+
+// renderVolumeMountSshConfigs renders [corev1.VolumeMount] defining the mounting path for SSH configs
```
asteny: sshd

internal/render/common/volume.go
```diff
@@ -300,3 +300,120 @@ func RenderVolumeMountRESTJWTKey() corev1.VolumeMount {
...
+		MountPath: consts.VolumeMountPathSSHConfigs,
+		ReadOnly:  true,
+	}
+}
+
+// endregion configs
+
+// region root keys
+
+// renderVolumeSshRootKeys renders [corev1.Volume] containing SSH root keys contents
```
asteny: sshd

internal/render/common/volume.go
```diff
@@ -300,3 +300,120 @@ func RenderVolumeMountRESTJWTKey() corev1.VolumeMount {
...
+				LocalObjectReference: corev1.LocalObjectReference{
+					Name: naming.BuildConfigMapSshRootPublicKeysName(clusterName),
+				},
+				DefaultMode: ptr.To(int32(384)),
+			},
+		},
+	}
+}
+
+// renderVolumeMountSshRootKeys renders [corev1.VolumeMount] defining the mounting path for SSH root keys
```
asteny: sshd

internal/render/worker/container.go
```diff
@@ -174,3 +175,57 @@ func renderRealMemorySlurmd(resources corev1.ResourceRequirements) int64 {
 	memoryInMebibytes := memoryInBytes / 1_048_576 // 1 MiB = 1,048,576 bytes
 	return memoryInMebibytes
 }
+
+func renderContainerSshd(
```
asteny: I think that this part same as in login node... Can we use it one time from common?

asteny (changes requested):

internal/render/worker/container.go
Uburro: they have different arguments.  i do not think we have to spend  time for this way

PR #249, opened 2024-12-13
run enroot containers without root privileges

asteny:


internal/render/worker/daemonset.go
```diff
@@ -0,0 +1,66 @@
...
+		Spec: appsv1.DaemonSetSpec{
+			Selector: &metav1.LabelSelector{
+				MatchLabels: matchLabels,
+			},
+			Template: corev1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: labels,
+				},
+				Spec: corev1.PodSpec{
+					ServiceAccountName: naming.BuildServiceAccountWorkerName(clusterName),
```
Uburro: it is not necessary, and better remove it

Uburro (approved):

PR #269, opened 2024-12-17
MSP-3724: fix Supervisor entrypoint

Uburro:


images/worker/supervisord.sh
rdjjke, line 1: You should also change the default supervisord config map in the Soperator code

Uburro: https://github.com/nebius/soperator/pull/269/files#diff-7d981bdd8096c8bbdf86b9656bbd373645365ae39266a32f734c50af94138e73L210

what do you mean?

internal/render/worker/configmap.go
rdjjke, line 1:
```
res.AddLine("command=/opt/bin/slurm/slurmd_entrypoint.sh")
```
Here should be the slurmd binary

Uburro: https://github.com/nebius/soperator/pull/269/files#diff-9e90dcbf1100350ca32f5496062b222a6484fa33168142c891138e1d8619d4abL95

i do not think it's good idea

Uburro: maybe will we remove it `2>&1 | tee >(multilog s100000000 n5 /var/log/slurm/multilog)` ?

Uburro: if yes - it will be readable more

rdjjke: Ah, OK, I got how you made this.

> maybe will we remove it 2>&1 | tee >(multilog s100000000 n5 /var/log/slurm/multilog) ?
Hmm, I don't think we should.

So, now slurm_entrypoint.sh uses some env vars. Are you sure supervisord transfers them to the slurmd entrypoint?

rdjjke: I don't think it does at least for the `GRES` env var as it's not exported

images/worker/supervisord.sh
rdjjke, line 1: Can we name it `supervisord_entrypoint.sh` to make it the same with all other containers?

internal/render/worker/configmap.go
Uburro: https://supervisord.org/configuration.html#environment-variables

Uburro: https://github.com/nebius/soperator/pull/269/commits/6dd4f8819755670c8a226792935650e6b1b79ea7
fixed

PR #301, opened 2025-01-07
MSP-3569: delete accounting pod, svc and mariadb when accounting false

Uburro:


internal/controller/clustercontroller/accounting.go
```diff
@@ -52,6 +48,11 @@ func (r SlurmClusterReconciler) ReconcileAccounting(
 					stepLogger := log.FromContext(stepCtx)
 					stepLogger.Info("Reconciling")
 
+					if !isAccountingEnabled || (!isExternalDBEnabled && !isMariaDBEnabled) {
```
rdjjke: NIT: This logical expression is quite hard to read. Can we simplify it somehow?

For example, we can add a new variable out of `reconcileAccountingImpl` func:
```
isDBEnabled := isExternalDBEnabled || isMariaDBEnabled
```

And then make all checks look so:
```
if !isAccountingEnabled && !isDBEnabled {
```

rdjjke (dismissed):

PR #334, opened 2025-01-20
MSP-3992: fix bug with empty version of annotation

Uburro:
When creating the StatefulSet, it was discovered that the StatefulSet is recreated twice, and during an update, the pod with the `-0` index is not updated. This happens because maps in Go are not merged.
image


internal/controller/reconciler/k8s_statefulset.go
```diff
@@ -48,7 +48,11 @@ func (r *StatefulSetReconciler) patch(existing, desired client.Object) (client.P
 		res := client.MergeFrom(dst.DeepCopy())
 
 		dst.Spec.Template.ObjectMeta.Labels = src.Spec.Template.ObjectMeta.Labels
```
asteny, line 50: do we have such issue with annotations and maybe some different maps?

dstaroff (approved):

internal/controller/reconciler/k8s_statefulset.go
Uburro: just annotations

PR #335, opened 2025-01-21
Move chroot plugin inside containers and bind mount it

asteny:


Uburro (dismissed):

images/controller/slurmctld.dockerfile
```diff
@@ -53,12 +53,12 @@ RUN for pkg in slurm-smd-client slurm-smd-dev slurm-smd-libnss-slurm slurm-smd-l
 
 RUN apt install -y /tmp/*.deb && rm -rf /tmp/*.deb
 
-# Install slurm plugins
```
rdjjke, line 56: Isn't it needed to install Pyxis here as well?
If scripts for plugin installation is splitter now, maybe we can create another one that calls all separate plugin installation scripts?
Just not to forget to install one of the plugins somewhere.

images/controller/slurmctld_entrypoint.sh
```diff
@@ -19,6 +19,10 @@ for file in /mnt/slurm-configs/*; do
     touch "/etc/slurm/$filename" && mount --bind "$file" "/etc/slurm/$filename"
 done
 
+echo "Bind-mount slurm chroot plugin from container at jail"
```
rdjjke: Maybe it's better to move it to complement_jail.sh?
Just because other similar bind-mounts are done there.

images/controller/slurmctld.dockerfile
asteny: removed

images/controller/slurmctld_entrypoint.sh
asteny: moved

rdjjke (approved):

PR #360, opened 2025-01-27
MSP-4037: add simple check controller

Uburro:


cmd/checkcontroller/main.go
```diff
@@ -0,0 +1,209 @@
...
+	metricsserver "sigs.k8s.io/controller-runtime/pkg/metrics/server"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+
+	slurmv1 "nebius.ai/slurm-operator/api/v1"
+	"nebius.ai/slurm-operator/internal/checkcontroller"
+	//+kubebuilder:scaffold:imports
+)
+
+var (
+	controllerName = "checkcontroller"
```
asteny: can u use checkcontroller.controllerName ?

internal/checkcontroller/reconcile.go
```diff
@@ -0,0 +1,518 @@
...
+	logger.V(1).Info("Updating conditions on nodes")
+	err = r.updateKubeNodeConditionSlurm(ctx, unschedulableNodes)
+	if err != nil {
+		return ctrl.Result{RequeueAfter: r.reconcileTimeout}, fmt.Errorf("error updating conditions on nodes: %w", err)
+	}
+
+	return ctrl.Result{RequeueAfter: r.reconcileTimeout}, nil
+}
+
+func (r *CheckControllerReconciler) processMatchingNodes(ctx context.Context, nodes []string, namespace string) (map[string]*corev1.Node, []error) {
```
itechdima: maybe instead of `[]error` use just `error`

and in the end of func, you can use `return unschedulableNodes , errors.Join(errs...)`

internal/checkcontroller/reconcile.go
```diff
@@ -0,0 +1,518 @@
...
+
+	logger.V(1).Info("Draining node")
+	if err := r.DrainNode(ctx, node); err != nil {
+		return fmt.Errorf("failed to drain node: %w", err)
+	}
+
+	logger.V(1).Info("Marking node as unschedulable")
+	node.Spec.Unschedulable = true
+	if err := r.Update(ctx, node); err != nil {
+		return fmt.Errorf("failed to update nodeto unschedulable: %w", err)
```
itechdima:
```suggestion
		return fmt.Errorf("failed to update node to unschedulable: %w", err)
```

internal/checkcontroller/reconcile.go
```diff
@@ -0,0 +1,518 @@
...
+	return nil
+}
+
+// CreateRebooterPodIfNeeded creates a rebooter pod for the node with the given name if it does not already exist.
+func (r *CheckControllerReconciler) CreateRebooterPodIfNeeded(ctx context.Context, workerName, nodeName, namespace string) error {
+	logger := log.FromContext(ctx).WithName("CreateRebooterPodIfNeeded").WithValues("workerName", workerName, "nodeName", nodeName)
+
+	logger.V(1).Info("Checking for existing rebooter pod")
+	existingPod, err := r.GetRebooterPod(ctx, nodeName, namespace)
+	if err == nil && existingPod != nil {
```
itechdima: what about case `err != nil && !errors.Is(ErrNotFound, err)`?
I mean, you tried to `GetRebooterPod`, it returned an error, but rebooter pod already exists

internal/checkcontroller/reconcile.go
```diff
@@ -0,0 +1,518 @@
...
+				Type:    conditionType,
+				Status:  status,
+				Message: message,
+				LastTransitionTime: metav1.Time{
+					Time: time.Now(),
+				},
+				LastHeartbeatTime: metav1.Time{
+					Time: time.Now(),
+				},
+			}
```
itechdima: lets maybe create new var for that, to avoid duplication in the next lines (266-276)
```
newCond := corev1.NodeCondition{
				Type:    conditionType,
				Status:  status,
				Message: message,
				LastTransitionTime: metav1.Time{
					Time: time.Now(),
				},
				LastHeartbeatTime: metav1.Time{
					Time: time.Now(),
				},
			}
```
and reuse it in line 250 and 266?

internal/checkcontroller/reconcile.go
```diff
@@ -0,0 +1,518 @@
...
+
+// GetPodNodeName gets the node name of the pod with the given name in the given namespace.
+func (r *CheckControllerReconciler) GetPodNodeName(ctx context.Context, nodeName, namespace string) (string, error) {
+	pod := &corev1.Pod{}
+	err := r.Get(ctx, client.ObjectKey{Name: nodeName, Namespace: namespace}, pod)
+	if err != nil {
+		return "", fmt.Errorf("error getting pod: %w", err)
+	}
+	if pod.Spec.NodeName == "" {
+		return "", fmt.Errorf("pod has no node name")
```
itechdima: for future debugging:
```suggestion
		return "", fmt.Errorf("pod %s has no node name", podName)
```

internal/checkcontroller/reconcile.go
```diff
@@ -0,0 +1,518 @@
...
+		},
+		LastHeartbeatTime: metav1.Time{
+			Time: time.Now(),
+		},
+	})
+	return r.Status().Update(ctx, node)
+}
+
+// GetPodNodeName gets the node name of the pod with the given name in the given namespace.
+func (r *CheckControllerReconciler) GetPodNodeName(ctx context.Context, nodeName, namespace string) (string, error) {
```
itechdima:
```suggestion
func (r *CheckControllerReconciler) GetPodNodeName(ctx context.Context, podName, namespace string) (string, error) {
```

internal/checkcontroller/reconcile.go
```diff
@@ -0,0 +1,518 @@
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      pod.Name,
+			Namespace: pod.Namespace,
+		},
+	}
+
+	if err := r.Client.SubResource("eviction").Create(ctx, pod, eviction); err != nil {
+		if apierrors.IsTooManyRequests(err) {
+			logger.V(1).Info("Pod eviction is rate-limited, retrying", "podName", pod.Name)
+			return nil
```
itechdima:
```suggestion
			return err
```

internal/checkcontroller/reconcile.go
```diff
@@ -0,0 +1,518 @@
...
+) (bool, error) {
+
+	for _, condition := range node.Status.Conditions {
+		if condition.Type == nodeConditionType && condition.Status == conditionStatus {
+			return true, nil
+		}
+	}
+
+	return false, nil
+}
```
itechdima:
```suggestion
func (r *CheckControllerReconciler) CheckNodeCondition(
	ctx context.Context, node *corev1.Node, nodeConditionType corev1.NodeConditionType, conditionStatus corev1.ConditionStatus,
) bool {

	for _, condition := range node.Status.Conditions {
		if condition.Type == nodeConditionType && condition.Status == conditionStatus {
			return true
		}
	}

	return false
}
```
probably we don't need an error here?

asteny (approved):

internal/checkcontroller/reconcile.go
```diff
@@ -0,0 +1,511 @@
...
+		if err := r.rebootNode(ctx, node, workerName, namespace); err != nil {
+			errs = append(errs, fmt.Errorf("error rebooting node: %w", err))
+		}
+	}
+
+	return unschedulableNodes, errors.Join(errs...)
+}
+
+// checkIfNodeNeedsReboot checks if the node with the given name needs to be rebooted.
+func (r *CheckControllerReconciler) checkIfNodeNeedsReboot(ctx context.Context, node *corev1.Node) (bool, error) {
```
itechdima: always `false`

itechdima (approved):

PR #372, opened 2025-01-28
MSP-4080: add CheckNodeCondition to rebooter

Uburro:


internal/rebooter/reconcile.go
```diff
@@ -27,23 +27,108 @@ var (
...
+		}
+	}
+
+	if r.checkIfNodeNeedsDrain(ctx, node) {
+		logger.Info("Node needs drain")
+
+		if err := r.DrainNode(ctx, node); err != nil {
+			return ctrl.Result{RequeueAfter: r.reconcileTimeout}, fmt.Errorf("failed to drain node %s: %w", r.nodeName, err)
+		}
+	}
```
itechdima: now it looks like if node has 2 conditions `Drain` and `Reboot`, it'll first be rebooted, and then drained, but the remaining drain looks like not needed

maybe it'll be done later in implementation of `RebootNode` that node Conditions (`Drain` and `Reboot`) will be cleaned somehow?

rdjjke (dismissed):

itechdima (approved):

PR #376, opened 2025-01-30
Update pyxis version with `container_image_save` and `expose_enroot_logs` enagled

### LINKED ISSUE #396: Update pyxis version with `container_image_save` and `expose_enroot_logs` enabled

itechdima:
Use new pyxis with options to save squashfs files and expose enroot logs.
Closes #396

### DIFF
--- a/images/common/scripts/complement_jail.sh
+++ b/images/common/scripts/complement_jail.sh
@@ -110,6 +110,9 @@ pushd "${jaildir}"
         flock etc/complement_jail_setcap_enroot_aufs2ovlfs.lock -c "setcap cap_sys_admin,cap_mknod+pe usr/bin/enroot-aufs2ovlfs"
     fi
 
+    echo "Shared folder for pyxis output /var/cache/enroot-container-images"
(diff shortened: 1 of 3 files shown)


Uburro (dismissed):

rdjjke (dismissed):

images/common/scripts/complement_jail.sh
```diff
@@ -110,6 +110,9 @@ pushd "${jaildir}"
         flock etc/complement_jail_setcap_enroot_aufs2ovlfs.lock -c "setcap cap_sys_admin,cap_mknod+pe usr/bin/enroot-aufs2ovlfs"
     fi
 
+    echo "Shared folder for pyxis output /var/cache/enroot-container-images"
```
asteny, line 113: Do we need such folder on login node? If not you can move that part to the worker side (`if [ -z "$worker" ]`)

itechdima: I would keep it, so in case you want to see what images you have, you don't need to ssh to worker, you can just simply list it from login node

PR #377, opened 2025-01-30
MSP-4080: add rebooting node check

Uburro:


internal/rebooter/reconcile.go
```diff
@@ -134,20 +138,40 @@ func (r *RebooterReconciler) checkIfNodeNeedsDrain(ctx context.Context, node *co
...
 	if r.CheckNodeCondition(ctx, node, consts.SlurmNodeReboot, corev1.ConditionTrue) {
-		logger.V(1).Info("Need to reboot node")
-		return true
+		logger.Info("Need to reboot node")
+		if nodeCondSlurmReboot := r.GetNodeConditions(ctx, node, consts.SlurmNodeReboot); nodeCondSlurmReboot != nil {
+			if r.IsUptimeGreaterThanLastTransition(ctx, nodeCondSlurmReboot.LastTransitionTime) {
+				logger.Info("Uptime is greater than last transition time")
+				return true
+			}
+		}
```
itechdima: NIT: I think there's even no need for `if r.CheckNodeCondition(ctx, node, consts.SlurmNodeReboot, corev1.ConditionTrue) {` line, just get node condition and check it, to not iterate over conditions twice (but it's not crucial anyway)
```suggestion
		nodeCondSlurmReboot := r.GetNodeConditions(ctx, node, consts.SlurmNodeReboot)
		if nodeCondSlurmReboot == nil {
			return false;
		}
		logger.WithValues("nodeCondition", nodeCondSlurmReboot).Info("Checking if node needs reboot)
		return nodeCondition.Status == corev1.ConditionTrue && 
		  r.IsUptimeGreaterThanLastTransition(ctx, nodeCondSlurmReboot.LastTransitionTime)
```

itechdima (approved):
just one NIT :)

PR #389, opened 2025-02-05
Pre-create enroot credentials, exclude enroot bind-mounts from motd, make Docker mount /dev/infiniband, store /tmp on disk and add tmpfs /mnt/memory

rdjjke:
See commit messages for details.


Uburro (approved):

images/jail/jail.dockerfile
```diff
@@ -114,6 +114,9 @@ RUN chmod +x /opt/bin/install_container_toolkit.sh && \
     /opt/bin/install_container_toolkit.sh && \
     rm /opt/bin/install_container_toolkit.sh
 
+# Copy NVIDIA Container Toolkit config
+COPY common/nvidia-container-runtime/config.toml /etc/nvidia-container-runtime/config.toml
```
asteny, line 187: Maybe it should be in worker container and bind mount it?

rdjjke: I thought about it, but then users won't be able to adjust this config for their needs (we don't make it configurable via a ConfigMap or something).

So I decided to just copy it to jail.

In the slurmd image its content is crucial (we use nvidia-container-runtime to propagate GPU drivers to jail), but inside the jail this config only allows users to run Docker containers without manually mounting `/dev/infiniband`. But they may want to change something there.

Uburro (approved):

asteny (approved):

Uburro (approved):

Uburro (dismissed):

itechdima (approved):

PR #390, opened 2025-02-05
Install nvtop as deb package from repo and bind mount it from container to the jail filesystem

asteny:


images/common/scripts/complement_jail.sh
```diff
@@ -138,8 +139,13 @@ pushd "${jaildir}"
         done
     fi
 
+    # For $worker node only
     if [ -n "$worker" ]; then
         echo "Update linker cache inside the jail"
         flock etc/complement_jail_ldconfig.lock -c "chroot \"${jaildir}\" /usr/sbin/ldconfig"
+
+        echo "Bind-mount nvtop from container to the jail"
```
rdjjke: Now nvtop package doesn't depend on nvidia driver (and we added a mock package for it anyway).
So why can't we just pre-install nvtop to jail and don't do any bind-mounts.
It's not necessarily to have it inside the slurmd image at all.

asteny: But what we should do if there is no jail filesystem?

rdjjke: In this case, we'll have the other image for slurmd anyway. And there, nvtop will be pre-installed.

rdjjke (approved):

PR #426, opened 2025-02-12
JWT issuance origin [MSP-3383]

dstaroff:
This PR implements JWT issuance process by introducing a dedicated package for managing token signing according to [Slurm specification](https://slurm.schedmd.com/jwt.html#compatibility). It also features a token cache as a replacement for storing token in secret and managing its expiration. Unit tests are added to validate token registry and caching behavior.
Alongside, it adjusts the Kubernetes service configuration to correct misaligned labels and selectors for the `slurmrestd` service.


internal/jwt/token.go
```diff
@@ -0,0 +1,211 @@
...
+	cluster  types.NamespacedName
+	username string
+
+	annotations map[string]string
+}
+
+// String implements fmt.Stringer interface.
+// Use it to get the key for the token.
+func (m TokenMeta) String() string {
+	res := fmt.Sprintf("%s/%s:%s", m.cluster, m.username, m.username)
```
asteny: Why there are  `m.username` 2 times?

rdjjke:
Is it actually "bug", "dependencies", "feature", "fix", "go", and "test"?

internal/jwt/token.go
rdjjke: The second username is probably the group

dstaroff: It supposed to be `namespace/name:user`. Fixed it

dstaroff:
> Is it actually "bug", "dependencies", "feature", "fix", "go", and "test"?

Yes.

- `bug` and `fix` are almost the same
- `dependencies` is for PRs changing `go.mod`
- `feature` because it's a new feature
- `go` is for PRs changing Go code
- `test` because there are new unit-tests being added

itechdima (approved):

PR #476, opened 2025-02-20
add sync versions for soperatorchecks

itechdima:


asteny (dismissed):

internal/soperatorchecks/soperatorchecks.go
```diff
@@ -0,0 +1,122 @@
...
+		LastHeartbeatTime: metav1.Time{
+			Time: time.Now(),
+		},
+	}
+}
+
+func getK8SNode(ctx context.Context, c client.Client, nodeName string) (*corev1.Node, error) {
+	node := &corev1.Node{}
+	if err := c.Get(ctx, client.ObjectKey{Name: nodeName}, node); err != nil {
+		return nil, fmt.Errorf("get node: %w", err)
```
Uburro, line 108: could be not found?

internal/soperatorchecks/slurm_nodes_controller.go
```diff
@@ -172,25 +220,48 @@ func (c *slurmWorkersController) processKillTaskFailed(
...
-			newNodeCondition(
-				consts.SlurmNodeDrain,
-				corev1.ConditionTrue,
-				consts.ReasonNodeDraining,
-				"",
-			),
-		)
+func (c *SlurmNodesController) processK8SNodesMaintenance(ctx context.Context) error {
+	var (
+		limit     int64 = 64
```
Uburro: what is the magic limit?

internal/soperatorchecks/slurm_nodes_controller.go
```diff
@@ -5,66 +5,114 @@ import (
...
 		}
 	}
 
-	return errors.Join(errs...)
+	if err := errors.Join(errs...); err != nil {
+		logger.V(1).Error(err, "Process degraded nodes produced an error")
+		return ctrl.Result{}, err
+	}
+
+	return ctrl.Result{RequeueAfter: c.reconcileTimeout}, nil
```
Uburro, line 113: add comment for future us why we always return RequeueAfter, please

internal/soperatorchecks/soperatorchecks.go
itechdima: yep, and we want an error in that case, so I don't ignore it

internal/soperatorchecks/slurm_nodes_controller.go
itechdima: just random limit :) 
will move to some const

Uburro (approved):

PR #496, opened 2025-02-25
480: do not run nccl test if some proccess running on gpus

Uburro:
[#480](https://github.com/nebius/soperator/issues/480)

### DIFF
--- a/images/worker/gpubench/main.go
+++ b/images/worker/gpubench/main.go
@@ -96,10 +96,21 @@ func init() {
 }
 
 func main() {
-	_ = fs.Parse(os.Args[1:])
+	fs.Parse(os.Args[1:])
 	log.Info(fmt.Sprintf("Starting %s", nameNCCL))
 	if *debugLog {
 		debugFlags()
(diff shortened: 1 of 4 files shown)


asteny (dismissed):

itechdima (dismissed):

images/worker/gpubench/main.go
```diff
@@ -220,6 +232,31 @@ func debugFlags() {
 	log.Debugf("log_format: %s", *logFormat)
 }
 
+func isRunningProcessOnGPU() (bool, error) {
+	cmd := exec.Command(
+		"chroot",
+		"/run/nvidia/driver",
+		"nvidia-smi",
```
rdjjke, line 239: It's not necessarily to parse nvidia-smi output in ASCII table format.
There are many arguments in `nvidia-smi`. For example, you can use the following:
```
nvidia-smi --query-compute-apps=gpu_serial,process_name --format=csv,noheader
```

asteny (approved):

PR #505, opened 2025-02-28
Implement NodeSet CRD

dstaroff:


api/v1alpha1/nodeset_types.go
```diff
@@ -0,0 +1,267 @@
...
+	// +kubebuilder:validation:Optional
+	ConfigMapRefSSHD string `json:"sshd,omitempty"`
+}
+
+// GPUSpec defines the settings related to GPU support
+type GPUSpec struct {
+	// Supported indicates whether GPU support is enabled for the NodeSet.
+	//
+	// +kubebuilder:validation:Optional
+	Supported bool `json:"supported,omitempty"`
```
rdjjke: Maybe we can name it "enabled"? We already use this word in many places. It's not important, but just to keep it consistent.

api/v1alpha1/nodeset_types.go
```diff
@@ -0,0 +1,267 @@
...
+
+	// ReadOnly defines whether the mounting point should be read-only
+	//
+	// +kubebuilder:validation:Optional
+	ReadOnly bool `json:"readOnly,omitempty"`
+
+	// VolumeSource defines the volume source for the sub-mount
+	//
+	// +kubebuilder:validation:Optional
+	VolumeSource *corev1.VolumeSource `json:"volumeSource,omitempty"`
```
rdjjke, line 257: In the current spec, this is `VolumeSourceName` -- the reference to the VolumeSource defined in a separate spec.
How will it work this way?

api/v1alpha1/nodeset_types.go
dstaroff: I wasn't sure in naming of this field.
Yes, `Enabled` would look consistent with other entities

api/v1alpha1/nodeset_types.go
dstaroff: Here we won't reference `VolumeSource`s defined in Cluster CR.
AFAIR, we concluded to do so on design meeting as it's easier to just define it one more time

rdjjke: Yes, sorry, I confused it a little: I thought that fields like SubPath and ReadOnly are part of VolumeSource.

In general, there's no need in having a separate VolumeSource field, but we can think about it in the future.

For now it would be easier to switch to DaemonSets

rdjjke (approved):

PR #506, opened 2025-02-28
504: customisable slurm config

itechdima:


api/v1/slurmcluster_types.go
```diff
@@ -156,6 +162,24 @@ type SlurmConfig struct {
...
+	// - "patch": Merges the raw configuration with existing SlurmConfig values (default).
+	// - "override": Applies only the raw configuration, ignoring values provided in SlurmConf.
+	//
+	// +kubebuilder:validation:Enum=patch;override
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=patch
+	Strategy string `json:"strategy"`
+	// RawContent contains the raw Slurm configuration.
+	//
+	// +kubebuilder:validation:Required
```
Uburro: let's without Required here?

itechdima: I want it to be here required
`SlurmConfigRaw` field is optional itself. So, if that field is passed, I want the `rawContent` to be in place

asteny (commented):
Could you add labels to the PR? Without labels in new release your PR will bi in section Other

asteny (approved):

Issue #487, opened 2025-02-24
Implement `NodeSetController`


dstaroff:
It should be able to:
- Reconcile `NodeSet` CRs as particular StatefulSets
- Provide `NodeFeature`s for worker nodes it creates
- Provide conditions on `NodeSet`s

PR #507, opened 2025-03-03
NodeSet Controller origin

dstaroff:
This PR is a first stage of #487

### DIFF
--- a/hack/boilerplate.go.txt
+++ b/hack/boilerplate.go.txt
@@ -1,5 +1,5 @@
 /*
-Copyright 2024 Nebius B.V.
+Copyright 2025 Nebius B.V.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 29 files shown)


hack/boilerplate.go.txt
```diff
@@ -1,5 +1,5 @@
 /*
```
asteny, line 1: Why we have such file here? We have license in root directory

internal/controller/nodesetcontroller/controller.go
```diff
@@ -0,0 +1,83 @@
+/*
+Copyright 2025 Nebius B.V.
```
asteny, line 2: Will we update this file every year at 1 January?)

hack/boilerplate.go.txt
dstaroff: @Uburro ?

internal/controller/nodesetcontroller/controller.go
dstaroff: I'd like to have a copyright update machinery connected to the repo

asteny: I'm not sure that we need update it and have it here. In root file we have such string https://github.com/nebius/soperator/blob/dev/LICENSE#L189 And we should not update it

asteny (approved):

Uburro (dismissed):

asteny (approved):

asteny (approved):

PR #543, opened 2025-03-10
508:  SConfigController | Added controller logic

andrei-pokhila:
Changes:
- Added “sConfigController” object to the spec of the kind **SlurmCluster**
- Watching for ConfigMaps with label “**sconficontroller.slurm.nebius.ai=true**” in the namespace with value of “**--cluster-namespace**” flag
- Removed config mounts for login & nccl-benchmark & worker & controller. Now they have configs in **/etc/slurm** (jail). Also removed their dependency from this ConfigMaps
- Now we have **two** SSHDConfigs ConfigMaps instead of one. One is for the **login** node, and one is for the **worker**. (fix of label reconciliation cycle)
- **Added reconciliation of ConfigMaps labels**. 

Important:
- We are **not watching for Deletion** event of ConfigMaps. If ConfigMap was deleted (or config file was deleted from ConfigMap) we won't delete configs from fs
- **Rest and Accounting services are still mounted with SlurmConfig ConfigMap**. They will restart after config change (we dont want to put jail into them)

Side Changes:
- Hardcoded version of mariadb for cluster-manager to version “(http://docker-registry1.mariadb.com/library/mariadb:11.4.3)”
- Fixed bug with flags not parsed in soperatorchecks


config/rbac/sconfigcontroller/role.yaml
```diff
@@ -0,0 +1,14 @@
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
Uburro: it should be kind: Role and it created soperator

internal/render/sconfigcontroller/deployment.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	return &appsv1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      naming.BuildDeploymentName(consts.ComponentTypeSConfigController),
+			Namespace: namespace,
+			Labels:    labels,
+		},
+		Spec: appsv1.DeploymentSpec{
+			// replicas should be 1.
```
Uburro: why?

internal/render/sconfigcontroller/pod.go
```diff
@@ -0,0 +1,64 @@
...
+	affinity = nodeFilter.Affinity
+	nodeSelector = nodeFilter.NodeSelector
+
+	return &corev1.PodTemplateSpec{
+		ObjectMeta: metav1.ObjectMeta{
+			Labels: matchLabels,
+			Annotations: map[string]string{
+				fmt.Sprintf(
+					"%s/%s", consts.AnnotationApparmorKey, consts.ContainerNameSConfigController,
+				): sConfigController.Container.AppArmorProfile,
```
Uburro: for  your controller appamor is not needed

internal/render/sconfigcontroller/container.go
```diff
@@ -0,0 +1,43 @@
...
+			ContainerPort: container.Port,
+			Protocol:      corev1.ProtocolTCP,
+		}},
+		VolumeMounts: []corev1.VolumeMount{
+			common.RenderVolumeMountJail(),
+		},
+		SecurityContext: &corev1.SecurityContext{
+			Capabilities: &corev1.Capabilities{
+				Add: []corev1.Capability{
+					consts.ContainerSecurityContextCapabilitySysAdmin,
```
Uburro: ContainerSecurityContextCapabilitySysAdmin these privileges are not required.

cmd/sconfigcontroller/main.go
```diff
@@ -126,9 +118,10 @@ func main() {
 		"If set, HTTP/2 will be enabled for the metrics and webhook servers")
 	flag.StringVar(&logFormat, "log-format", "json", "Log format: plain or json")
 	flag.StringVar(&logLevel, "log-level", "debug", "Log level: debug, info, warn, error, dpanic, panic, fatal")
-	flag.DurationVar(&reconcileTimeout, "reconcile-timeout", 5*time.Minute, "The maximum duration allowed for a single reconcile")
 	flag.IntVar(&maxConcurrency, "max-concurrent-reconciles", 1, "Configures number of concurrent reconciles. It should improve performance for clusters with many objects.")
 	flag.DurationVar(&cacheSyncTimeout, "cache-sync-timeout", 5*time.Minute, "The maximum duration allowed for caching sync")
+	flag.StringVar(&configsPath, "configs-path", "/mnt/jail/slurm", "Path where to store configs")
+	flag.StringVar(&configsNamespace, "configs-namespace", "soperator", "Namespace of the soperator configs")
```
Uburro: default instead of soperator. Same as a default in helm

api/v1/slurmcluster_types.go
```diff
@@ -190,6 +195,12 @@ type MPIConfig struct {
 	PMIxEnv string `json:"pmixEnv,omitempty"`
 }
 
+type SConfigController struct {
+	SlurmAPIServer string        `json:"slurmAPIServer,omitempty"`
```
Uburro: SlurmAPIServer name generated by soperator, why we set this variable here?

internal/render/rest/service.go
```diff
@@ -30,3 +32,7 @@ func RenderService(namespace, clusterName string, rest *values.SlurmREST) (*core
 		},
 	}, nil
 }
+
+func GetServiceURL(namespace string, rest *values.SlurmREST) string {
+	return fmt.Sprintf("http://%s.%s:%d", rest.Service.Name, namespace, rest.ContainerREST.Port)
```
Uburro: add`fmt.Sprintf("http://%s.%s.svc:%d"` please. We have issue from our client that they take a config from cross namespace.

Uburro (dismissed):

internal/controller/sconfigcontroller/file_store.go
```diff
@@ -0,0 +1,33 @@
...
+	if _, err = file.Write([]byte(content)); err != nil {
+		return fmt.Errorf("write file: %w", err)
+	}
+
+	if err = file.Close(); err != nil {
+		return fmt.Errorf("closing file: %w", err)
+	}
+
+	return nil
+}
```
itechdima: to not leave file open if `Write` failed
```suggestion
func (s *FileStore) Add(name string, content string) error {
	file, err := os.Create(fmt.Sprintf("%s/%s", s.path, name))
	if err != nil {
		return fmt.Errorf("open file: %w", err)
	}
	defer file.Close()

	if _, err = file.Write([]byte(content)); err != nil {
		return fmt.Errorf("write file: %w", err)
	}

	return nil
}
```

internal/render/worker/configmap.go
```diff
@@ -143,3 +144,54 @@ func generateDefaultSupervisordConfig() renderutils.ConfigFile {
...
+			Namespace: cluster.Namespace,
+			Labels:    common.RenderLabels(componentType, cluster.Name),
+		},
+		Data: map[string]string{
+			consts.ConfigMapKeySshdConfig: generateSshdConfig(cluster).Render(),
+		},
+	}, nil
+}
+
+func generateSshdConfig(cluster *values.SlurmCluster) renderutils.ConfigFile {
```
itechdima, line 167: isn't it a duplications of https://github.com/nebius/soperator/blob/9b181e9fefd2450ca36671debe0268cb52a5f9f0/internal/render/login/configmap.go#L61
?
maybe add common exported function?

andrei-pokhila: We've decided to make two different ConfigMaps. Previously it was one common function for both login/worker.
If we have two different ConfigMaps, I decided to make two different functions which generate content. Even if the have the same code inside.
Thats was the thought process behind. If you insist than I'll change

itechdima (approved):

Uburro (approved):

PR #565, opened 2025-03-19
Enable gpus metrics in exporter

nandexsp:
In order to enable the gpu metrics is necessary to launch the exporter with the -gpus -acct flag. In the https://github.com/vpenso/prometheus-slurm-exporter/ repo we can found this note: 
```
NOTE: since version 0.19, GPU accounting has to be explicitly enabled adding the -gpus-acct option to the command line otherwise it will not be activated.
```
Thanks

### DIFF
--- a/images/exporter/exporter.dockerfile
+++ b/images/exporter/exporter.dockerfile
@@ -90,4 +90,4 @@ RUN ldconfig
 
 COPY --from=exporter_builder /app/prometheus-slurm-exporter /opt/bin/
 
-ENTRYPOINT ["/opt/bin/prometheus-slurm-exporter"]
+ENTRYPOINT ["/opt/bin/prometheus-slurm-exporter", "--gpus-acct"]
--- a/images/exporter/exporter.dockerfile
+++ b/images/exporter/exporter.dockerfile
@@ -90,4 +90,4 @@ RUN ldconfig
 
 COPY --from=exporter_builder /app/prometheus-slurm-exporter /opt/bin/
 
-ENTRYPOINT ["/opt/bin/prometheus-slurm-exporter", "--gpus-acct"]
+ENTRYPOINT ["/opt/bin/prometheus-slurm-exporter", "-gpus-acct"]


Uburro:
it failed in my test env if gres/gpu disabled
https://github.com/vpenso/prometheus-slurm-exporter/blob/master/README.md#state-of-the-gpus

Uburro:
could you show me accountingStorageTRES, please?

nandexsp:
> could you show me accountingStorageTRES, please?

nando@login-0:~$ scontrol show config | grep AccountingStorageTRES
AccountingStorageTRES   = cpu,mem,energy,node,billing,fs/disk,vmem,pages,gres/gpu,gres/gpumem,gres/gpuutil

nandexsp:
the other info that I got is from slurm-ai-acct-db-metrics:9104/metrics, where can I found the exporter, or is the same endpoint??

nandexsp:
> > > could you show me accountingStorageTRES, please?
> > 
> > 
> > nando@login-0:~$ scontrol show config | grep AccountingStorageTRES AccountingStorageTRES = cpu,mem,energy,node,billing,fs/disk,vmem,pages
> 
> if you set this one cpu,mem,energy,node,billing,fs/disk,vmem,pages
> 
> is that still works?

Will check ir in 12 hours more less

nandexsp:
> > > could you show me accountingStorageTRES, please?
> > 
> > 
> > nando@login-0:~$ scontrol show config | grep AccountingStorageTRES AccountingStorageTRES = cpu,mem,energy,node,billing,fs/disk,vmem,pages
> 
> if you set this one cpu,mem,energy,node,billing,fs/disk,vmem,pages
> 
> is that still works?

Are not the same?

Uburro:
Anyway I will check it tomorrow again

nandexsp:
without the GPU flags in the TRES is not working. 

```
Empty reply from server
```

what solution could you recommend for users that wants to use GPUs? 

because two exporters are not the solution

Uburro:
> without the GPU flags in the TRES is not working.
> 
> ```
> Empty reply from server
> ```
> 
> what solution could you recommend for users that wants to use GPUs?
> 
> because two exporters are not the solution

use [podTemplateRef](https://github.com/nebius/soperator/blob/dev/api/v1/slurmcluster_types.go#L896) or adding ARGS  to spec of the exporters when `-gpus-acct` needed

nandexsp:
HOw can I deploy this? I was trying to use the slinky one but is not ready for gpu metrics

Uburro:
`podTemplateRef: my-exporter`

```
apiVersion: v1
kind: PodTemplate
metadata:
  name: my-exporter
  labels:
    app: my-app
template:
  spec:
    containers:
      - name: "exporter"
        image: "cr.eu-north1.nebius.cloud/soperator/exporter:1.19.0-jammy-slurm24.05.5"
        command: ["/opt/bin/prometheus-slurm-exporter", "-gpus-acct"]
        resources:
          limits:
            memory: "256Mi"
          requests:
            cpu: "250m"
            memory: "128Mi"
```

PR #591, opened 2025-03-25
terraform apply/destroy scenario

itechdima:


test/e2e/e2e_test.go
```diff
@@ -1,11 +1,164 @@
...
+		Vars:         tfVars,
+		RetryableTerraformErrors: map[string]string{
+			"Context deadline exceeded": "retry on context deadline exceeded",
+		},
+		MaxRetries: 5,
+	}
+
+	terraform.Init(t, terraformOptions)
+	terraform.WorkspaceSelectOrNew(t, terraformOptions, "e2e-test")
+	terraform.Destroy(t, terraformOptions)
```
Uburro, line 45: is not mistake? Below `defer terraform.Destroy(t, terraformOptions)`

itechdima: I want to destroy first in case there're some leftovers from previous test run

PR #598, opened 2025-03-26
use vars instead of secrets, fix syntax

itechdima:


Uburro (dismissed):

.github/workflows/e2e_test.yml
```diff
@@ -59,7 +59,7 @@ jobs:
         run: |
           git clone --branch "${{ env.TERRAFORM_REPO_VERSION }}" "${{ env.TERRAFORM_REPO }}" terraform-repo
           
-          cd terraform-repo/${{ .github.event.inputs.path_to_installation }}
+          cd terraform-repo/${{ github.event.inputs.path_to_installation }}
```
asteny, line 67: Why you dont want use here actions/checkout ?

itechdima: updated to use checkout action

asteny (approved):

PR #607, opened 2025-03-26
fix env vars in e2e tests

itechdima:


asteny (approved):

test/e2e/e2e_test.go
```diff
@@ -44,8 +44,18 @@ func TestTerraform(t *testing.T) {
...
 
-	defer terraform.Destroy(t, terraformOptions)
-	terraform.Apply(t, terraformOptions)
+	var applyError error
+	defer func() {
+		if applyError != nil {
+			t.Errorf("Failed to apply: %v", err)
+		}
+
+		terraform.Destroy(t, terraformOptions)
```
Uburro, line 53: DestroyE?
cause how we recognise failed destroy or apply or both of them?

itechdima: I firstly log apply error and then destroy
in case both of them are failed - we'll see all errors

itechdima: rollback here:
https://github.com/nebius/soperator/pull/608

Issue #623, opened 2025-04-01
Use Advanced StatefulSet


dstaroff:
Regular K8s' StatefulSet has some fields immutable, which doesn't suit our needs.
Advanced StatefulSet from OpenKruise is much more preferred in our case.

Instead of installing OpenKruise operator, we can try to install the schema from the Soperator.

dstaroff:
Injection of Kruise operator parts responsive for Advanced StatefulSets didn't work out. There were few collisions in dependency versions led to incompatibility and failing build.
So, we chose to make installed Kruise operator a dependency of Soperator.

dstaroff:
Replacement of StatefulSets for controllers, logins, and workers, with Advanced StatefulSets is done. Proper implementation of Kruise Helm chart dependency is still left.

PR #672, opened 2025-04-09
#623 Advanced stateful sets

dstaroff:
This PR adds integration with Kruise’s Advanced StatefulSets - enables Soperator to leverage Kruise’s extended StatefulSet capabilities.

This feature requires that the Kruise controller is deployed and properly configured in the cluster, which is stated in Soperator Helm chart's `README.md`.

### DIFF
--- a/internal/check/consts.go
+++ b/internal/check/consts.go
@@ -1,0 +1,6 @@
+package check
+
+const (
+	ForceTrue  = true
+	ForceFalse = false
+)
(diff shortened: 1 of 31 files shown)


Uburro (approved):

internal/check/consts.go
```diff
@@ -0,0 +1,6 @@
+package check
+
+const (
+	ForceTrue  = true
```
asteny, line 4: 🤔

dstaroff: έ𝓝Ｔ€ᖇℙ尺Ɨ𝕊乇

PR #673, opened 2025-04-09
Issue-654 Customize slurm healthcheck script and interval

ChessProfessor:


internal/render/common/configmap.go
```diff
@@ -102,9 +102,9 @@ func generateSlurmConfig(cluster *values.SlurmCluster, topologyConfig corev1.Con
 	res.AddComment("")
 	res.AddComment("HEALTH CHECKS")
 	res.AddComment("https://slurm.schedmd.com/slurm.conf.html#OPT_HealthCheckInterval")
-	res.AddProperty("HealthCheckInterval", 30)
+	res.AddProperty("HealthCheckInterval", cluster.HealthCheckConfig.HealthCheckInterval)
 	if cluster.ClusterType == consts.ClusterTypeGPU {
```
asteny: We need to think about that option. Maybe we do not need it if we customize healthcheck program

ChessProfessor: We can get rid of default value for `HealthCheckProgram` and if it's empty do not run checks at all, wdyt?

ChessProfessor: But it means we have to migrate all current clusters to correct spec to not break health checks for them

ChessProfessor: Or we can have optional HealthCheckConfig with required HealthCheckInterval & HealthCheckProgram in it. If it's not presented we'll have the same logic as now (every 30 sec gpu checks if cluster type gpu), otherwise both parameters will be configured by client

asteny (approved):

api/v1/slurmcluster_types.go
```diff
@@ -235,6 +240,32 @@ type WorkerFeature struct {
...
+	// +kubebuilder:validation:MinItems=1
+	HealthCheckNodeState []HealthCheckNodeState `json:"healthCheckNodeState,omitempty"`
+}
+
+type HealthCheckNodeState struct {
+	// State identifies node state on which HealthCheckProgram should be executed.
+	//
+	// +kubebuilder:validation:Enum=ALLOC;ANY;CYCLE;IDLE;NONDRAINED_IDLE;MIXED
+	// +kubebuilder:validation:Required
+	State string `json:"state,omitempty"`
```
Uburro: if `Required` it should not be `omitempty`

ChessProfessor: Fixed

helm/slurm-cluster/templates/slurm-cluster-cr.yaml
```diff
@@ -32,6 +32,15 @@ spec:
       hostlistExpr: {{ .hostlistExpr | quote }}
       nodesetName: {{ .nodesetName | quote }}
     {{- end }}
+  {{- if .Values.healthCheckConfig }}
+  healthCheckConfig:
```
Uburro, line 36: {{ .Values.healthCheckConfig | toYaml | indent 2 }} ?
 bacause it's optional of value it could be empty

Uburro (approved):

theyoprst (approved):

PR #677, opened 2025-04-09
Allow launching background processes in login container

angelbejarano:
There are use cases where some commands will need to be executed after login container has been started.

Proposal is making use of /etc/rc.local file from the SysV init startup scripts days and run it in background when it exists in the jail filesystem.


angelbejarano:
Does not custom entrypoint require a non-standard image?

Uburro:
> Does not custom entrypoint require a non-standard image?

it is not required. You can submount it to your container and [execute](https://github.com/kubernetes/kubernetes/issues/71356#issuecomment-441169334).

Uburro (commented):
I think this is PR could be cloused ?

Issue #680, opened 2025-04-10
Setup linters properly


theyoprst:
Linters help to find certain bugs and ensure some level of code quality. Right now we turned on a minimum set of linters. We need to turn on at least popular ones (e.g. govet) and set up it properly. But-catching linters should be prioritized.

This article summarize most of the linters (beware: 2y old):
https://olegk.dev/go-linters-configuration-the-right-version

Another source of inspiration: golangci-lint configuration of Nebius managed kubernetes team (closed source).

theyoprst:
## Progress Update: depguard linter enabled

✅ **Completed**: Enable `depguard` linter and migrate from pkg/errors

**Created subissue**: #1001 - Enable depguard linter and migrate from pkg/errors to standard library

### What was accomplished:
- ✅ Migrated entire codebase from `github.com/pkg/errors` to standard library `fmt.Errorf()`
- ✅ Enabled `depguard` linter to prevent future pkg/errors usage
- ✅ All tests passing, builds successful

This addresses the dependency management and code quality aspects of the linter setup initiative. The `depguard` linter now actively prevents introduction of the deprecated `pkg/errors` package.

**Next steps for #680**: Consider enabling other high-value linters like `govet`, `staticcheck`, etc.

PR #702, opened 2025-04-15
Turn on errcheck linter.

theyoprst:
Issue #680

### DIFF
--- a/internal/rebooter/reconcile.go
+++ b/internal/rebooter/reconcile.go
@@ -86,7 +86,10 @@ func (r *RebooterReconciler) Reconcile(ctx context.Context, req ctrl.Request) (c
 		return ctrl.Result{}, err
 	}
 
-	nodeActions := r.GetActions(ctx, node)
+	nodeActions, err := r.GetActions(ctx, node)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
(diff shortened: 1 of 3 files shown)


internal/rebooter/reconcile.go
```diff
@@ -135,8 +135,10 @@ func (r *RebooterReconciler) GetActions(ctx context.Context, node *corev1.Node)
 	if nodeRebootCondition != nil {
 		if !r.IsUptimeGreaterThanLastTransition(ctx, nodeRebootCondition.LastTransitionTime) {
 			logger.Info("Node does not need to be rebooted")
-			r.setNodeCondition(ctx, node, consts.SlurmNodeReboot, corev1.ConditionTrue, consts.ReasonNodeRebooted, consts.MessageRebooted)
-			return actions
+			err := r.setNodeCondition(ctx, node, consts.SlurmNodeReboot, corev1.ConditionTrue, consts.ReasonNodeRebooted, consts.MessageRebooted)
+			if err != nil {
+				logger.Error(err, "Failed to set 'rebooted' node condition")
+			}
```
itechdima, line 144: missing `return actions`

and maybe yes, better to return an error, so we won't miss this event, it'll be retired

theyoprst: Done

Uburro (approved):

PR #703, opened 2025-04-15
#639 Change ActiveCheck CR status based on k8s jobs

ChessProfessor:


api/v1alpha1/extra_types.go
```diff
@@ -35,59 +35,8 @@ type StatusMetadata struct {
...
-	// +kubebuilder:validation:MaxLength=1024
-	// +kubebuilder:validation:MinLength=1
-	Reason string `json:"reason"`
-	// Message is a human readable message indicating details about the transition.
-	// This may be an empty string.
-	// +optional
-	// +kubebuilder:validation:MaxLength=32768
-	Message string `json:"message,omitempty"`
+	// Jobs maps child job identifier to its status
+	Jobs map[string]batchv1.JobStatus `json:"jobs,omitempty"`
```
Uburro: why is map of string and batchv1.JobStatus, not just batchv1.JobStatus. It has several string and array of condion

internal/controller/soperatorchecks/activecheck_jobs_controller.go
```diff
@@ -0,0 +1,155 @@
...
+			DeleteFunc: func(e event.DeleteEvent) bool {
+				return false
+			},
+			GenericFunc: func(e event.GenericEvent) bool {
+				configMap, ok := e.Object.(*corev1.ConfigMap)
+				if !ok {
+					return false
+				}
+
+				return isValidConfigMap(configMap)
```
Uburro: it is not configMap

internal/controller/soperatorchecks/activecheck_jobs_controller.go
```diff
@@ -0,0 +1,155 @@
...
+			},
+		})).
+		WithOptions(controllerconfig.ControllerOptions(maxConcurrency, cacheSyncTimeout)).
+		Complete(r)
+}
+
+func isValidConfigMap(cm *corev1.ConfigMap) bool {
+	if v, ok := cm.Annotations[consts.LabelComponentKey]; ok {
```
Uburro: service discovery should be by label, not annotation

internal/controller/soperatorchecks/activecheck_jobs_controller.go
```diff
@@ -0,0 +1,155 @@
...
+		if apierrors.IsNotFound(err) {
+			logger.Info("ActiveCheckJob resource not found. Ignoring since object must be deleted.")
+			return ctrl.Result{}, nil
+		}
+		logger.Error(err, "Failed to get ActiveCheckJob")
+		return ctrl.Result{}, err
+	}
+
+	activeCheckNamespacedName := job.Annotations[consts.AnnotationActiveCheckKey]
+	parts := strings.SplitN(activeCheckNamespacedName, "/", 2)
```
Uburro, line 121: why do not use  `req.NamespacedName`?

internal/controller/soperatorchecks/activecheck_jobs_controller.go
```diff
@@ -0,0 +1,155 @@
...
+	if len(parts) != 2 {
+		return ctrl.Result{}, fmt.Errorf("invalid input format")
+	}
+	activeCheckNamespace := parts[0]
```
Uburro: why do not use `req.NamespacedName`?

internal/controller/soperatorchecks/activecheck_jobs_controller.go
```diff
@@ -0,0 +1,155 @@
...
+	activeCheckName := parts[1]
```
Uburro: it should be value of this annotation

api/v1alpha1/extra_types.go
ChessProfessor: I'm sorry I don't get what you mean
I did it to easily map Job to JobStatus using its name. We can use []batchv1.JobStatus, but it doesn't contain its name

internal/controller/soperatorchecks/activecheck_jobs_controller.go
ChessProfessor: You mean having only ActiveCheck name in annotations and get namespace from req.Namespace? I initially did it like that, but then decided that storing namespace is more safe. I can revert it back if you say so

api/v1alpha1/extra_types.go
Uburro: JobStatus batchv1.JobStatus `json:"-"` ?

importing struct `batchv1.JobStatus` to `StatusMetadata` and wiork with that

ChessProfessor: There might be several jobs running concurrently for one CronJob. I added deleting of finished jobs from ActiveCheck status to not make it huge

itechdima: Maybe it's better to add it [here](https://github.com/nebius/soperator/blob/f34a2f70968aea53f4de6794035895bc9dda1efb/api/v1alpha1/activecheck_types.go#L111-L113) instead of common struct for StatusMetadata, since it's only related to the checks

api/v1alpha1/extra_types.go
```diff
@@ -35,59 +35,8 @@ type StatusMetadata struct {
...
-	// +kubebuilder:validation:Required
-	// +kubebuilder:validation:MaxLength=1024
-	// +kubebuilder:validation:MinLength=1
-	Reason string `json:"reason"`
-	// Message is a human readable message indicating details about the transition.
-	// This may be an empty string.
-	// +optional
-	// +kubebuilder:validation:MaxLength=32768
-	Message string `json:"message,omitempty"`
+	// Jobs maps child job identifier to its status
```
theyoprst: Also I think using maps in API is almost always a bad idea: you need to specify clearly what the key does mean, and in this case, it's not clear to me, what is "job identifier".

I propose list of jobs instead, it's more self-documented approach, like this:

```go
struct JobStatusMetadata {
  Namespace string
  Name string
  Status batchv1.JobStatus
}

...

Jobs []JobStatusMetadata `json:"jobs,omitempty"`
```

api/v1alpha1/activecheck_types.go
```diff
@@ -107,9 +107,34 @@ type K8sJobSpec struct {
...
+
+	LastK8sJobCompletionTime *metav1.Time            `json:"lastK8sJobEndTime"`
+	LastK8sJobName           string                  `json:"lastK8SJobName"`
+	LastK8sJobStatus         ActiveCheckK8sJobStatus `json:"lastK8sJobStatus"`
+}
+
+// ActiveCheckK8sJobStatus defines status for ActiveCheck k8s job.
+type ActiveCheckK8sJobStatus string
+
+const (
```
Uburro: it should be in packege const

internal/controller/soperatorchecks/activecheck_jobs_controller.go
```diff
@@ -0,0 +1,199 @@
...
+			logger.Info("CronJob resource not found. Ignoring since object must be deleted.")
+			return ctrl.Result{}, nil
+		}
+		logger.Error(err, "Failed to get CronJob")
+		return ctrl.Result{}, err
+	}
+
+	activeCheck.Status = slurmv1alpha1.ActiveCheckStatus{
+		K8sJobsStatus: slurmv1alpha1.ActiveCheckK8sJobsStatus{
+			LastTransitionTime: metav1.Now(),
```
Uburro: it should be changed when has a drift beetween actual state and disered

api/v1alpha1/activecheck_types.go
ChessProfessor: I don't think so, it's more like enum than consts
For example the same approach is used for `UpdateStatus` in `extra_types.go`

internal/controller/soperatorchecks/activecheck_jobs_controller.go
ChessProfessor: Makes sense, let me see how to do that easily

api/v1alpha1/activecheck_types.go
Uburro: https://github.com/nebius/soperator/blob/dev/internal/consts/conditions.go

internal/controller/soperatorchecks/soperatorchecks.go
```diff
@@ -19,7 +19,8 @@ import (
 //+kubebuilder:rbac:groups=core,resources=nodes/status,verbs=get;update;patch;watch;list
 //+kubebuilder:rbac:groups=slurm.nebius.ai,resources=slurmclusters,verbs=get;watch;list
 //+kubebuilder:rbac:groups=coordination.k8s.io,resources=leases,verbs=get;watch;list;update;create
-// +kubebuilder:rbac:groups=batch,resources=cronjobs,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=batch,resources=cronjobs,verbs=get;list;watch;create;update;patch;delete
```
Uburro, line 22: `create;update;patch;delete`

why is it necessary ?

ChessProfessor: I don't know, I just removed space in the beginning of the line to make it consistent with others :)

ChessProfessor: Do you think we can remove that?

Uburro: remove it please. I do not see where we modificate the jobs

api/v1alpha1/activecheck_types.go
```diff
@@ -107,9 +108,22 @@ type K8sJobSpec struct {
...
+
+	LastK8sJobCompletionTime *metav1.Time                   `json:"lastK8sJobEndTime"`
+	LastK8sJobName           string                         `json:"lastK8SJobName"`
+	LastK8sJobStatus         consts.ActiveCheckK8sJobStatus `json:"lastK8sJobStatus"`
+}
+
 // ActiveCheckStatus defines the observed state of ActiveCheck.
 type ActiveCheckStatus struct {
 	StatusMetadata `json:",inline"`
+	K8sJobsStatus  ActiveCheckK8sJobsStatus `json:"k8sJobsStatus,omitempty"`
```
theyoprst, line 125: Maybe name it `LastK8sJobStatus` instead of listing each field inside as "Last"?

api/v1alpha1/activecheck_types.go
```diff
@@ -107,9 +108,22 @@ type K8sJobSpec struct {
...
 
+// ActiveCheckK8sJobsStatus defines the observed state of ActiveCheck k8s jobs.
+type ActiveCheckK8sJobsStatus struct {
+	LastTransitionTime metav1.Time `json:"lastTransitionTime"`
+
+	LastK8sJobScheduleTime   *metav1.Time `json:"lastK8sJobScheduleTime"`
+	LastK8sJobSuccessfulTime *metav1.Time `json:"lastK8sJobSuccessfulTime"`
+
+	LastK8sJobCompletionTime *metav1.Time                   `json:"lastK8sJobEndTime"`
+	LastK8sJobName           string                         `json:"lastK8SJobName"`
```
theyoprst: `lastK8SJobName -> lastK8sJobName`

api/v1alpha1/activecheck_types.go
ChessProfessor: Idea was that it may contain not only information about the last job. For example LastK8sJobSuccessfulTime uses time from last successful job

ChessProfessor: But I agree, maybe we have to think about namings around here, it's pretty messy

rdjjke (approved):

PR #730, opened 2025-04-22
Turn on unparam linter

theyoprst:


internal/controller/clustercontroller/accounting.go
```diff
@@ -460,31 +456,31 @@ func (r SlurmClusterReconciler) ValidateAccounting(
...
+		return r.updateAccountingAvailabilityStatus(ctx, cluster, metav1.ConditionFalse, "NotAvailable", message, 10*time.Second)
 	case existingDeployment.Status.AvailableReplicas == 0:
 		message := "Slurm accounting is not available yet"
-		return r.updateStatus(ctx, cluster, slurmv1.ConditionClusterAccountingAvailable, metav1.ConditionFalse, "NotAvailable", message, 10*time.Second)
+		return r.updateAccountingAvailabilityStatus(ctx, cluster, metav1.ConditionFalse, "NotAvailable", message, 10*time.Second)
 	case existingDeployment.Status.AvailableReplicas != targetReplicas:
 		message := fmt.Sprintf("Slurm accounting available replicas: %d, but target replicas: %d", existingDeployment.Status.AvailableReplicas, targetReplicas)
-		return r.updateStatus(ctx, cluster, slurmv1.ConditionClusterAccountingAvailable, metav1.ConditionFalse, "NotAvailable", message, 10*time.Second)
+		return r.updateAccountingAvailabilityStatus(ctx, cluster, metav1.ConditionFalse, "NotAvailable", message, 10*time.Second)
 	}
```
dstaroff, line 481: Maybe it would be nicer to just set the message in switch case, and then
```go
if message != "" {
return r.updateAccountingAvailabilityStatus(ctx, cluster, metav1.ConditionFalse, "NotAvailable", message, 10*time.Second)
}
```

dstaroff (approved):
LGTM

internal/controller/clustercontroller/accounting.go
theyoprst: Thanks for the suggestion, it looks reasonable, although I would prefer to leave it as is, because:
- the other returns in this function outside of this switch are done the same way, and it's not obvious how to generalize all of them
- not sure whether some of the parameters should be the same in this switch (e.g. in the future)
- not sure it would be more readable

If refactor, I would go in a bit another direction: move get rid of some params like reason and requeueAfter, and in this case message could be moved to the function call instead of a separate variable

PR #733, opened 2025-04-22
add manual job trigger for runAfterCreation

itechdima:


internal/render/soperatorchecks/job.go
```diff
@@ -0,0 +1,41 @@
...
+			},
+			Labels:      labels,
+			Annotations: cronJob.Spec.JobTemplate.Annotations,
+		},
+		Spec: *cronJob.Spec.JobTemplate.Spec.DeepCopy(),
+	}
+}
+
+func RenderK8sJobName(check *slurmv1alpha1.ActiveCheck) string {
+	return fmt.Sprintf("%s-manual-run", check.Spec.Name)
```
rdjjke: The run is not "manual", can we name it differently?

itechdima: `%s-triggered`, wdyt?

rdjjke: What about `-initial-run`? To show the relation to the `runAfterCreation` param that controls it

rdjjke (dismissed):

asteny (dismissed):

asteny (approved):

PR #812, opened 2025-05-01
Add k8s job base image for SlurmJob check

ChessProfessor:


images/k8s_job/k8s_job.dockerfile
```diff
@@ -0,0 +1,101 @@
...
+
+# Update linker cache
+RUN ldconfig
+
+# Delete users & home because they will be linked from jail
+RUN rm /etc/passwd* /etc/group* /etc/shadow* /etc/gshadow*
+RUN rm -rf /home
+
+# Make sbatch script executable
+RUN chmod +x /opt/bin/sbatch.sh
```
asteny: I am not sure that we need it

images/k8s_job/k8s_job_entrypoint.sh
```diff
@@ -0,0 +1,48 @@
...
+
+if [[ -z "$SLURM_OUTPUT" ]]; then
+    echo "Failed to submit Slurm job"
+    exit 1
+fi
+
+SLURM_JOB_ID="$SLURM_OUTPUT"
+echo "Slurm Job ID: $SLURM_JOB_ID"
+
+POD_NAME=$(hostname)
```
asteny: U already have env var `K8S_POD_NAME=worker-0`

images/k8s_job/k8s_job_entrypoint.sh
```diff
@@ -0,0 +1,45 @@
...
+if [[ -z "$SLURM_OUTPUT" ]]; then
+    echo "Failed to submit Slurm job"
+    exit 1
+fi
+
+SLURM_JOB_ID="$SLURM_OUTPUT"
+echo "Slurm Job ID: $SLURM_JOB_ID"
+
+echo "Looking up owning Job for pod: $K8S_POD_NAME in namespace: $K8S_POD_NAMESPACE"
+K8S_JOB_NAME=$(kubectl get pod "$K8S_POD_NAME" -n "$K8S_POD_NAMESPACE" -o jsonpath='{.metadata.ownerReferences[?(@.kind=="Job")].name}')
```
Uburro: What do you think about curl request without kubectl? This way, updating the kubectl version based on the Kubernetes version will no longer be required.

internal/controller/soperatorchecks/serviceaccount_controller.go
```diff
@@ -0,0 +1,315 @@
...
+		For(&slurmv1alpha1.ActiveCheck{}, builder.WithPredicates(
+			predicate.Funcs{
+				CreateFunc: func(e event.CreateEvent) bool {
+					return true
+				},
+				DeleteFunc: func(e event.DeleteEvent) bool {
+					return false
+				},
+				UpdateFunc: func(e event.UpdateEvent) bool {
+					if ac, ok := e.ObjectNew.(*slurmv1alpha1.ActiveCheck); ok {
```
Uburro, line 84: don't we support updates of the SC via Active Check?

internal/controller/soperatorchecks/serviceaccount_controller.go
```diff
@@ -0,0 +1,315 @@
...
+
+	cluster := &slurmv1.SlurmCluster{}
+	clusterNN := types.NamespacedName{
+		Namespace: req.Namespace,
+		Name:      check.Spec.SlurmClusterRefName,
+	}
+	err = r.Get(ctx, clusterNN, cluster)
+	if err != nil {
+		if apierrors.IsNotFound(err) {
+			logger.V(1).Info("SlurmCluster resource not found")
```
Uburro: return ctrl.Result{}, errors.Wrap(err, "SlurmCluster resource not found") ?

internal/controller/soperatorchecks/serviceaccount_controller.go
```diff
@@ -0,0 +1,315 @@
...
+					stepLogger.V(1).Info("Reconciling")
+
+					existing := &corev1.ServiceAccount{}
+					err := r.Get(ctx, types.NamespacedName{
+						Name:      naming.BuildServiceAccountActiveCheckName(clusterNN.Name),
+						Namespace: clusterNN.Namespace,
+					}, existing)
+
+					if err == nil {
+						stepLogger.V(1).Info("ServiceAccount already exists, skipping")
```
Uburro: Why are we skipping it? If it already exists, we should let it reconcile. We might add some labels later for service discovery of this SA.

internal/controller/soperatorchecks/serviceaccount_controller.go
```diff
@@ -0,0 +1,315 @@
...
+					stepLogger.V(1).Info("Reconciling")
+
+					existing := &rbacv1.Role{}
+					err := r.Get(ctx, types.NamespacedName{
+						Name:      naming.BuildRoleActiveCheckName(clusterNN.Name),
+						Namespace: clusterNN.Namespace,
+					}, existing)
+
+					if err == nil {
+						stepLogger.V(1).Info("Role already exists, skipping")
```
Uburro: same

internal/controller/soperatorchecks/serviceaccount_controller.go
```diff
@@ -0,0 +1,315 @@
...
+					stepLogger := log.FromContext(stepCtx)
+					stepLogger.V(1).Info("Reconciling")
+
+					existing := &rbacv1.RoleBinding{}
+					err := r.Get(ctx, types.NamespacedName{
+						Name:      naming.BuildRoleBindingActiveCheckName(clusterNN.Name),
+						Namespace: clusterNN.Namespace,
+					}, existing)
+
+					if err == nil {
```
Uburro: same

internal/controller/soperatorchecks/serviceaccount_controller.go
ChessProfessor: It's needed because deletion of ActiveCheck comes with update event with finalizers. So in order to properly handle them reconciler should trigger on update

internal/controller/soperatorchecks/serviceaccount_controller.go
ChessProfessor: Makes sense

images/k8s_job/k8s_job_entrypoint.sh
ChessProfessor: Had a discussion in slack, decided to stick to kubectl for now and use plain curl in case there will be problems with versions incompatibility

internal/controller/soperatorchecks/activecheck_controller.go
```diff
@@ -175,6 +179,11 @@ func (r *ActiveCheckReconciler) Reconcile(
 		}
 	}
 
+	if !check.Status.ServiceAccountReady {
+		logger.Info("Waiting for service account to be ready")
+		return ctrl.Result{}, nil
```
Uburro: `return ctrl.Result{Requeue: true, RequeueAfter: 5 * time.Second}, nil` ?

PR #832, opened 2025-05-07
NCCL Debug SPANK plugin

dstaroff:
This PR implements the plugin itself.

Its consequent:
- Building process and deployment process - #795
- And additional features:
  - #1021 
  - #1022

Will be made separately.


asteny:
Just for history: I think, that adding everything slurm related into one repo is a mess. This is a way to the monorepository with all its disadvantages.

images/common/spank-nccl-debug/Makefile
```diff
@@ -0,0 +1,102 @@
...
+ifeq ($(ARCH),arm64)
+ARCH_DOCKER = arm64v8
+else
+ARCH_DOCKER = amd64
+endif
+
+# Or `debug`
+TARGET_COMPILATION_MODE ?= release
+
+SLURM_VERSION ?= 24.05.5.1
```
theyoprst: We just have updated to 24.11.5, looks like you should update something as well.
https://github.com/nebius/soperator/pull/1014

dstaroff: Ok, but this is only applied to external builds. The plan for Soperator is to build the plugin along with the `chroot` plugin, which doesn't require docker at all, as it could be compiled with just `gcc` using Slurm headers and library present on the system.

rdjjke (approved):

images/common/spank-nccl-debug/src/snccld_args.h
```diff
@@ -0,0 +1,505 @@
...
+#define SNCCLD_ARG_OUT_STDOUT_USAGE                                                  \
+    "whether to additionally redirect `" SNCCLD_NCCL_ENV_DEBUG "` outputs to stdout. " \
+    "Possible values are case-insensitive. "                                         \
+    SNCCLD_ARG_OUT_STDOUT_ENV " env var is also supported."
+// clang-format on
+#pragma endregion
+
+/// Parse argument `__arg` with `__parse_fn` if it starts with `__arg_name`.
+#define SNCCLD_PARSE_ARG(__arg, __arg_name, __parse_fn)                        \
+    if (strncmp(__arg, __arg_name, strlen(__arg_name)) == 0) {                 \
```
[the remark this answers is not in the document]
dstaroff: Irrelevant as:
- `SNCCLD_PARSE_ARG` is supposed to parse plugin arguments passed in `plugstack.conf` without double dash
- Double dash is used for options provided from `srun` call, which is handled by `SNCCLD_ARG_OPTION`

rdjjke (approved):

PR #837, opened 2025-05-08
link soperatorchecks home directory

itechdima:


joao-krauchenco-mottu (approved):

images/login/sshd_entrypoint.sh
```diff
@@ -15,6 +15,9 @@ ln -s /mnt/jail/etc/update-motd.d /etc/update-motd.d
 echo "Link home from jail to use SSH keys from there"
 ln -s /mnt/jail/home /home
 
+echo "Link soperatorchecks home from jail to use SSH keys from there"
+ln -s /mnt/jail/opt/soperatorchecks /opt/soperatorchecks
```
asteny, line 19: Why soperatorchecks ssh keys in `/opt/soperatorchecks`. I s this a home dir? If yes, why not /home?

itechdima: Because to use `/home` directory we need to mount jail + nfs (if specified) everywhere, since nfs overwrites `/home`
With using `/opt` for the home directory for system users we could mount only jail in checks

Why we need to mount home directory - to access ssh key (in that approach soperatorchecks will hame ssh private and public key stored in `~/.ssh/`)

asteny: Where is a code which creates `/mnt/jail/opt/soperatorchecks` in jail before login starts?

itechdima: it'll be created in checks

Uburro (approved):

Issue #791, opened 2025-04-30
[soperatorchecks] Use precreated pod template for slurmJob type


asteny:

ChessProfessor:
Eventually decided to create helm chart only for ActiveCheck and prefill it with image and envs required for pod. The pod itself is constructed in soperatorchecks controller

PR #856, opened 2025-05-16
#791 Helm for ActiveCheck

ChessProfessor:


internal/controller/soperatorchecks/activecheck_controller.go
```diff
@@ -126,6 +126,11 @@ func (r *ActiveCheckReconciler) Reconcile(
 	}
 
 	if !controllerutil.ContainsFinalizer(check, consts.ActiveCheckFinalizer) {
+		if !controllerutil.ContainsFinalizer(check, consts.ActiveCheckServiceAccountFinalizer) {
```
ChessProfessor, line 129: This is needed because both reconcilers triggers on ActiveCheck creation and they both want to add their Finalizer at the same time. Actually it might be a good idea to unite them if it's not a problem to reconcile service accout and roles on every ActiveCheck spec update. Then we won't need these finalizer syncs

helm/soperator-activechecks/templates/activecheck.yaml
```diff
@@ -0,0 +1,38 @@
+{{ if .Values.activeCheck.enabled }}
+{{- if .Values.activeCheck.enabled }}
```
asteny, line 1: Why 2 times?

ChessProfessor: Have no idea

internal/controller/soperatorchecks/activecheck_controller.go
ChessProfessor: https://github.com/nebius/soperator/issues/874

PR #885, opened 2025-05-19
Basic Helm chart for DCGM exporter with HPC job mapping

ali-sattari:
This is a bare basic Helm chart generated from existing resources to get a DCGM exporter with HPC job mapping (volume mount and env var)

### DIFF
--- a/helm/soperator-dcgm-exporter/templates/daemonset.yaml
+++ b/helm/soperator-dcgm-exporter/templates/daemonset.yaml
@@ -1,0 +1,137 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: {{ include "soperator-dcgm-exporter.fullname" . }}-ds
+  labels:
+    app: nvidia-dcgm-exporter
+  {{- include "soperator-dcgm-exporter.labels" . | nindent 4 }}
+spec:
+  selector:
+    matchLabels:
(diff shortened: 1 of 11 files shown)


helm/soperator-dcgm-exporter/templates/servicemonitor.yaml
```diff
@@ -0,0 +1,26 @@
+apiVersion: monitoring.coreos.com/v1
+kind: ServiceMonitor
+metadata:
+  name: {{ include "soperator-dcgm-exporter.fullname" . }}-service-monitor
+  labels:
+    app: nvidia-dcgm-exporter
+  {{- include "soperator-dcgm-exporter.labels" . | nindent 4 }}
+spec:
+  endpoints:
+  - interval: 15s
```
Uburro: Maybe we should make this variable configurable?

helm/soperator-dcgm-exporter/templates/daemonset.yaml
```diff
@@ -0,0 +1,137 @@
+apiVersion: apps/v1
+kind: DaemonSet
+metadata:
+  name: {{ include "soperator-dcgm-exporter.fullname" . }}-ds
```
Uburro: "soperator-dcgm-exporter.name" ?
That way, it’ll be shorter name

PR #922, opened 2025-05-20
Delete existing kind cluster if exists

Uburro:


.github/workflows/delete-kind-cluster.yml
```diff
@@ -0,0 +1,23 @@
+name: Delete Kind Cluster
+
+on:
+  workflow_dispatch:
+    inputs:
+      cluster-name:
```
theyoprst: not used?

Issue #668, opened 2025-04-08
Design topology-aware job scheduling


rdjjke:
Slurms supports `topology/tree` and `topology/block` plugins. They allows to inform Slurm about the network topology in which nodes are interconnected. Slurm controller uses this knowledge to make better scheduling choices.

The goal is to automatically fetch network topology paths from K8s nodes (i.e. labels) and update it in Slurm.

The task can be implemented in two stages:
Stage 1. Make the topology config and TopologyPlugin setting configurable by the end user without the need to restart all worker pods
Stage 2. Automatically fetch topology paths from K8s node labels and update the topology config

Before proceeding to Stage 2, we should research the new `topology/block` plugin -- how it differs from `topology/tree` and whether the same algorithm can be used for both plugins.

Uburro:
![Image](https://github.com/user-attachments/assets/18398a5d-c6f8-43c5-9530-3f87be427814)

Uburro:
Design Overview for Topology-Aware Slurm Configuration in Kubernetes
The design is based on the assumption that the cluster may contain hundreds of Slurm worker nodes and Kubernetes nodes. Each Kubernetes node equipped with a GPU is labeled with one or more topologyconf.slurm.nebius.ai/tier-n labels, where n indicates the tier number of the switch closest to the node. The lower the tier number, the closer the switch is to the node. Tiers start from 1 and may increase depending on the network topology.

These labels are used to construct a virtual topology for Slurm (see [Slurm Topology Guide](https://slurm.schedmd.com/topology.html)).

Given previous experience operating clusters with up to 600 nodes, the following architectural decisions were made:

1. nodeTopologyController
This controller watches Kubernetes nodes via the API server. Upon detecting a node with a topologyconf.slurm.nebius.ai/tier-1 label, it records the node’s tier information in the node-topoly-labels ConfigMap in a nodeName: [tier-x: switchName, ...] format.

Example:
Given the following nodes:

yaml
Copy
Edit
apiVersion: v1
kind: Node
metadata:
  labels:
    topologyconf.slurm.nebius.ai/tier-1: leaf00
    topologyconf.slurm.nebius.ai/tier-2: spine00
  name: nodeA
---
apiVersion: v1
kind: Node
metadata:
  labels:
    topologyconf.slurm.nebius.ai/tier-1: leaf00
    topologyconf.slurm.nebius.ai/tier-2: spine00
  name: nodeB
---
apiVersion: v1
kind: Node
metadata:
  labels:
[...]

PR #940, opened 2025-05-22
[topology-aware] add topologyconfcontroller #427

Uburro:
#668
[#427](https://github.com/nebius/soperator/issues/427)

### DIFF
--- a/internal/consts/configmap.go
+++ b/internal/consts/configmap.go
@@ -31,4 +31,6 @@ const (
 	ConfigMapKeySysctl                  = sysctlConfFile
 	ConfigMapKeySupervisord             = supervisordConfFile
 	ConfigMapKeySoperatorcheckSbatch    = "sbatch.sh"
+
+	CongigMapNameNodesTopology = "nodes-topology"
 )
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -50,6 +50,7 @@ import (
 	"nebius.ai/slurm-operator/internal/controller/clustercontroller"
 	"nebius.ai/slurm-operator/internal/controller/nodeconfigurator"
 	"nebius.ai/slurm-operator/internal/controller/nodesetcontroller"
+	"nebius.ai/slurm-operator/internal/controller/topologyconfcontroller"
 	webhookcorev1 "nebius.ai/slurm-operator/internal/webhook/v1"
 	//+kubebuilder:scaffold:imports
 )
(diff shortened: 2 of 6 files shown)


internal/consts/configmap.go
```diff
@@ -31,4 +31,6 @@ const (
 	ConfigMapKeySysctl                  = sysctlConfFile
 	ConfigMapKeySupervisord             = supervisordConfFile
 	ConfigMapKeySoperatorcheckSbatch    = "sbatch.sh"
+	TopologyConfigMapName               = "topologyconf"
```
rdjjke: Probably it should be a part of the `slurm-configs` configmap, if we want it to be updated by SConfigController

internal/controller/topologyconfcontroller/topologyconf_controller.go
```diff
@@ -0,0 +1,352 @@
...
+	nodeList := &corev1.NodeList{}
+	nodeLabelSelector := client.HasLabels{consts.TierOnePrefix}
+
+	if err := r.Client.List(ctx, nodeList, nodeLabelSelector); err != nil {
+		logger.Error(err, "Failed to list nodes with topology labels")
+		return ctrl.Result{}, err
+	}
+
+	// Build node topology map: nodeName -> map[tier]value
+	nodeTiers := make(map[string]map[int]string)
```
rdjjke: I suggest abstracting the algorithm and moving it to a separate function that can be easily covered with unit tests. Because otherwise it looks too complicated IMO.

The function can accept a list of structures that look so:
```
type node struct {
    name string // worker name
    networkPath []string // here are the values from tier-1, tier-2, ... labels -- actually, switch names
}
```
And return the list of structures that look so:
```
type link struct {
	fromSwitch string
	toSwitches []string
	toNodes    []string
}
```

And then another function, that accepts `[]link` and returns the topology.conf file content. Such a function can also be easily covered by unit tests.

internal/consts/configmap.go
Uburro: it's not necessary. SConfigController could store to same path

internal/controller/topologyconfcontroller/topologyconf_controller.go
Uburro: yes, i wil split it. It's just check of theory

theyoprst (approved):

PR #949, opened 2025-05-26
make sync-version-from-scratch actions

Uburro:


.github/workflows/helm_chart.yml
```diff
@@ -51,7 +51,11 @@ jobs:
 
       - name: Checkout repository
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
-     
+
+      - name: Run make sync-version-from-scratch # Every time we should have only stable version in our code. So UNSTABLE=false here
+        run: |
+          make sync-version-from-scratch UNSTABLE=false
+
```
asteny, line 58: Add plz step which will check if there are no changes after that https://github.com/nebius/soperator/blob/dev/.github/workflows/one_job.yml#L168

Issue #874, opened 2025-05-16
Refactor ServiceAccountReconciler


ChessProfessor:
It does make sense to unite two ActiveCheck reconcilers: https://github.com/nebius/soperator/pull/856#discussion_r2093012556

ChessProfessor:
Hey! Yes, we'd be glad any help :) 

So the idea is that both ActiveCheckReconciler and ServiceAccountReconciler are watching ActiveCheck entity updates and because of that we need complex logic with syncing them (using ServiceAccountReady from ActiveCheck status and finalizers). If we unite these reconcilers we may get rid of this logic, make it more straightforward and mistake-proof. So I suggest to get rid of ServiceAccountReconciler and put its logic into ActiveCheckReconciler

This is pretty important task for us, I was going to do it in the nearest weeks. Please drop a message when you start it so that we will not do it twice :)

iamrajiv:
Thanks, @ChessProfessor, for the insights. Got it, I am working on this.

ChessProfessor:
We don't like approach with one unite reconciler either because of unnecessary reconciliation of SA, Role and RoleBinding on every update of ActiveCheck and also because of this [good practice](https://book.kubebuilder.io/reference/good-practices#why-should-one-avoid-a-system-design-where-a-single-controller-is-responsible-for-managing-multiple-crds-custom-resource-definitionsfor-example-an-install_all_controllergo)

So the new idea is to get rid of all the logic for deleting SA, since it's the same for every ActiveCheck we can just leave it. And remove all the sync logic between ActiveCheckController and ServiceAccountController. CRDs they make should be created without a problem even with no sync and there shouldn't be any problem even if ActiveCheck CronJob will be created before SA is ready.

PR #950, opened 2025-05-26
feat: Decouple `ServiceAccountReconciler` from `ActiveCheckReconciler` sync

iamrajiv:
Fixes: #874


internal/controller/soperatorchecks/activecheck_controller.go
```diff
@@ -285,12 +354,55 @@ func (r *ActiveCheckReconciler) reconcileDelete(ctx context.Context, check *slur
 		}
 	}
 
+	var checks slurmv1alpha1.ActiveCheckList
+	if err := r.List(ctx, &checks, client.InNamespace(check.Namespace)); err != nil {
```
Uburro: Events reach the reconciler when the ActiveCheck field of the custom resource is updated. In this case, performing a full List operation inside the reconciler is unnecessary and expensive.

iamrajiv: Ah, okay, thanks @Uburro. Yes, I only need to determine whether the ServiceAccount resources should be deleted, not retrieve all ActiveChecks. I’ll just check if any other ActiveCheck exists for the same cluster. I’ve pushed the new changes.

ChessProfessor:
Hello! I'm sorry for such hustle, we discussed with the core team possible approaches here and the concept has changed a bit
We don't like current approach either because of unnecessary reconciliation of SA, Role and RoleBinding on every update of ActiveCheck and also because of this [good practice](https://book.kubebuilder.io/reference/good-practices#why-should-one-avoid-a-system-design-where-a-single-controller-is-responsible-for-managing-multiple-crds-custom-resource-definitionsfor-example-an-install_all_controllergo)

So we came up with new idea. Let's get back to two controllers for ActiveCheck and ServiceAccount as it was before. Let's get rid of all the logic for deleting SA, since it's the same for every ActiveCheck we can just leave it. And let's remove all the sync logic between ActiveCheckController and ServiceAccountController. CRDs they make should be created without a problem even with no sync and there shouldn't be any problem even if ActiveCheck CronJob will be created before SA is ready.

If you still encouraged to help us with that we'd be thankful but tell me if it's too much for you :)

iamrajiv:
Hey hi @ChessProfessor! Yeah sure, I will refactor this. Thanks for the insights, great learning to be honest.
Should I make a fresh PR or continue with the existing one?

cc: @Uburro

ChessProfessor:
I'd say we can continue here :)

iamrajiv:
@ChessProfessor I made the changes and refactored, can you check once?

internal/controller/soperatorchecks/serviceaccount_controller.go
```diff
@@ -81,10 +75,8 @@ func (r *ServiceAccountReconciler) SetupWithManager(
 					return false
 				},
 				UpdateFunc: func(e event.UpdateEvent) bool {
```
ChessProfessor, line 77: Here we don't need to trigger on update because we don't reconcile finalizer removal anymore, just `return false` is good.

internal/controller/soperatorchecks/serviceaccount_controller.go
```diff
@@ -125,25 +115,9 @@ func (r *ServiceAccountReconciler) Reconcile(
 	}
 
 	if !check.DeletionTimestamp.IsZero() {
```
ChessProfessor: And we also don't need this condition because we don't care about ActiveCheck updates, only creation

config/crd/bases/slurm.nebius.ai_activechecks.yaml
```diff
@@ -7192,9 +7192,6 @@ spec:
               reason:
                 description: Reason defines human readable error reason
                 type: string
-              serviceAccountReady:
```
ChessProfessor, line 7195: You also have to update these files similarly: helm/soperator-crds/templates/slurmcluster-crd.yaml, helm/soperator/crds/slurmcluster-crd.yaml
It can be done by make sync-version-from-scratch from root.

ChessProfessor:
It looks good now, I started CI for this PR, if it will be green I'll have to test that pipeline still works on my testing cluster and then we are done here :) 
I'll let you know when I finish with this

ChessProfessor (approved):
I've tested these changes on my cluster and looks like everything works as we expected. Thank you for contributing!

Issue #793, opened 2025-04-30
[EPIC] Develop Soperator metrics exporter


rdjjke:
## Existing Solutions

[https://github.com/SlinkyProject/slurm-exporter/blob/main/internal/exporter/exporter.go](https://github.com/SlinkyProject/slurm-exporter/blob/main/internal/exporter/exporter.go)  

Pros:
- modern supported solution

Cons:
- labels breakdown doesn't meet our needs

[https://github.com/vpenso/prometheus-slurm-exporter](https://github.com/vpenso/prometheus-slurm-exporter)

Pros:
- has a ready-to-use dashboard

Cons:
- is not supported
- based on SLURM CLI and has its ugly artifacts (like state with asterisk)
## Why do we need our own solution
- We need another breaking-down of the metrics by labels
- We might need to to log something
## Why not fork existing solutions
They are not big, less than 1K LOC.
## Which metrics do we need




| **Metric Name**                              | **Labels**                                                                                                                                                                                                                            | **Description**                                                                                                   |
| -------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------- |
| `slurm_job_info`<br>gauge<br>always =1       | `job_id=12345`<br>`job_state=COMPLETED\|FAILED\|NODE_FAIL...`<br>`job_state_reason=NonZeroExitCode\|None\|...`<br>`slurm_partition=none`<br>`job_name=...`<br>`user_name=...`<br>`standard_error=...`<br>`standard_output=...`        | Slurm job indicator, use cases:<br>- jobs state timeline dashboard and/or table<br>- (?) join with slurm_node_job |
| `slurm_node_info`<br>gauge<br>always =1      | `node_name=worker-111`<br>`compute_instance_id=...`<br>`address=...`<br>`state_base=ALLOCATED\|DOWN\|ERROR\|IDLE\|MIXED`<br>`state_is_drain=true\|false`<br>`state_is_maintenance=true\|false`<br>`state_is_reserved=true\|false`<br> | Slurm node indicator, use cases:<br>- Nodes in drain graph<br>- Down nodes graph<br>- Total nodes graph           |
| `slurm_node_job`<br>gauge<br>always =1       | `job_id`<br>`node_name`                                                                                                                                                                                                               | "Job on node" indicator<br>Use cases:<br>- join slurm and nodes<br>                                               |
[...]

PR #959, opened 2025-06-02
Add Soperator exporter app.

theyoprst:
Initial version of the soperator exporter app. For now, it does two things:
- exposes a single metric
- does an API request which is not used by any metric so far

Infra for the app (the reconciler) will be provided separately.

Issue: #793

### DIFF
--- a/cmd/exporter/main.go
+++ b/cmd/exporter/main.go
@@ -19,6 +19,7 @@ package main
 import (
 	"context"
 	"flag"
+	"fmt"
 	"os"
 	"os/signal"
 	"syscall"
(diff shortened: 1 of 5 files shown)


internal/exporter/collector.go
```diff
@@ -0,0 +1,54 @@
...
+	jobNode     *prometheus.Desc
+}
+
+// NewMetricsCollector creates a new MetricsCollector
+func NewMetricsCollector(slurmAPIClient slurmapi.Client, soperatorVersion string) *MetricsCollector {
+	clusterInfoConstLabels := prometheus.Labels{"soperator_version": soperatorVersion}
+	return &MetricsCollector{
+		slurmAPIClient: slurmAPIClient,
+
+		clusterInfo: prometheus.NewDesc("slurm_cluster_info", "Slurm cluster information", []string{}, clusterInfoConstLabels),
```
asteny: This is a soperator version, not slurm. Why you want to put it in labels? Better I think is a separate metic with value

theyoprst: Discussed: decided to rename to `soperator_cluster_info`.

cmd/exporter/main.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+func parseFlags() Flags {
+	var flags Flags
+	flag.StringVar(&flags.logFormat, "log-format", "json", "Log format: plain or json")
+	flag.StringVar(&flags.logLevel, "log-level", "debug", "Log level: debug, info, warn, error, dpanic, panic, fatal")
+	flag.StringVar(&flags.metricsAddr, "metrics-bind-address", ":8080", "The address the metric endpoint binds to.")
+	flag.StringVar(&flags.slurmAPIServer, "slurm-api-server", "http://localhost:6820", "The address of the Slurm REST API server.")
+	flag.StringVar(&flags.clusterNamespace, "cluster-namespace", "soperator", "The namespace of the Slurm cluster")
+	flag.StringVar(&flags.clusterName, "cluster-name", "soperator", "The name of the Slurm cluster")
```
Uburro: In my opinion, an empty argument indicates an issue, so the code should fail if the argument is empty.

theyoprst: Agree, done

PR #971, opened 2025-06-05
Soperator Exporter: add node metrics.

theyoprst:


internal/slurmapi/node.go
```diff
@@ -81,3 +92,34 @@ func (n *Node) IsIdleDrained() bool {
...
+}
+
+var baseStates = []slurmapispec.V0041NodeState{
+	slurmapispec.V0041NodeStateUNKNOWN,
+	slurmapispec.V0041NodeStateDOWN,
+	slurmapispec.V0041NodeStateIDLE,
+	slurmapispec.V0041NodeStateALLOCATED,
+	slurmapispec.V0041NodeStateERROR,
+	slurmapispec.V0041NodeStateMIXED,
+}
```
itechdima, line 113: drain?

itechdima (approved):

internal/slurmapi/node.go
theyoprst: Discussed: drain can be used with any of these states, there is a separate metric label `is_drain` for that.

PR #981, opened 2025-06-11
#1000 Support eachWorkerJobArray in ActiveCheck spec

ChessProfessor:


images/slurm_check_job/slurm_check_job_entrypoint.sh
```diff
@@ -21,6 +21,20 @@ rm -rf /etc/slurm && ln -s /mnt/jail/slurm /etc/slurm
 echo "Bind-mount /opt/bin/sbatch.sh script"
 mount --bind /opt/bin/sbatch.sh opt/bin/sbatch.sh
 
+echo "Setting Extra field to all nodes..."
+for i in $(seq 0 $((NODE_NUM - 1))); do
+    node="worker-${i}"
```
theyoprst: Please let's not hardcode current naming scheme for the workers, it might change in the future.

theyoprst (approved):

images/slurm_check_job/slurm_check_job_entrypoint.sh
```diff
@@ -21,15 +21,19 @@ rm -rf /etc/slurm && ln -s /mnt/jail/slurm /etc/slurm
...
 
-echo "Submitting Slurm job..."
-SLURM_OUTPUT=$(/usr/bin/sbatch --parsable /opt/bin/sbatch.sh)
-
-if [[ -z "$SLURM_OUTPUT" ]]; then
-    echo "Failed to submit Slurm job"
-    exit 1
+if [[ "$EACH_WORKER_JOB_ARRAY" == "true" ]]; then
+    echo "Submitting job using slurm_submit_array_job.sh..."
+    SLURM_JOB_ID=$(/opt/bin/slurm/slurm_submit_array_job.sh | tail -n 1)
```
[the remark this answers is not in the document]
ChessProfessor: I guess we don't need it because `slurm_submit_array_job.sh` checks for the correct output and returns exit code 1 in case of error. Also we have `set -e` in this script, so this error will be handled

internal/controller/sconfigcontroller/file_store.go
```diff
@@ -51,3 +51,19 @@ func (s *FileStore) Add(name, content, subPath string) error {
 
 	return nil
 }
+
+func (s *FileStore) SetExecutable(name, subPath string) error {
```
theyoprst, line 55: How about adding a unit test for this function? We already have file_store_test.go.
But that's up to you

theyoprst (dismissed):

internal/controller/sconfigcontroller/file_store.go
ChessProfessor: Let me do that in a separate PR

edit: fixed it in this PR

internal/controller/soperatorchecks/activecheck_prolog_controller.go
```diff
@@ -0,0 +1,138 @@
...
+	ctx context.Context,
+	req ctrl.Request,
+) (ctrl.Result, error) {
+	logger := log.FromContext(ctx).WithName(SlurmActiveCheckPrologControllerName)
+	logger.Info(
+		"Starting reconciliation", "SlurmCluster", req.Name, "Namespace", req.Namespace,
+	)
+
+	if err := r.updatePrologConfigMap(ctx, req.Namespace, r.getPrologScript()); err != nil {
+		logger.Error(err, "Failed to update ConfigMap with topology config")
```
itechdima:
```suggestion
		logger.Error(err, "Failed to update ConfigMap with active check prolog config")
```

PR #998, opened 2025-06-12
#710 Add reactions to active checks

ChessProfessor:


internal/slurmapi/job.go
```diff
@@ -126,6 +122,31 @@ func (j Job) GetRequiredNodeList() ([]string, error) {
...
+		string(slurmapispec.V0041JobInfoJobStateOUTOFMEMORY),
+		string(slurmapispec.V0041JobInfoJobStateBOOTFAIL),
+		string(slurmapispec.V0041JobInfoJobStateDEADLINE),
+		string(slurmapispec.V0041JobInfoJobStateLAUNCHFAILED),
+		string(slurmapispec.V0041JobInfoJobStateNODEFAIL),
+		string(slurmapispec.V0041JobInfoJobStatePREEMPTED),
+		string(slurmapispec.V0041JobInfoJobStateRECONFIGFAIL),
+		string(slurmapispec.V0041JobInfoJobStateREVOKED),
+		string(slurmapispec.V0041JobInfoJobStateSPECIALEXIT),
+		"":
```
theyoprst: why empty state is terminal?

internal/slurmapi/interface.go
```diff
@@ -10,6 +10,6 @@ type Client interface {
 	api.ClientWithResponsesInterface
 	ListNodes(ctx context.Context) ([]Node, error)
 	GetNode(ctx context.Context, nodeName string) (Node, error)
-	GetJobStatus(ctx context.Context, jobID string) (JobStatus, error)
+	GetJob(ctx context.Context, jobID string) ([]Job, error)
```
theyoprst: What do you think of renaming it to `GetJobsByID()` to underline the fact that many jobs can be returned?

internal/slurmapi/client.go
```diff
@@ -116,41 +116,29 @@ func (c *client) GetNode(ctx context.Context, nodeName string) (Node, error) {
...
-		return JobStatus{}, fmt.Errorf("get job %s: %w", jobID, err)
+		return []Job{}, fmt.Errorf("get job %s: %w", jobID, err)
 	}
 	if getJobResp.JSON200 == nil {
-		return JobStatus{}, fmt.Errorf("json200 field is nil for job ID %s", jobID)
+		return []Job{}, fmt.Errorf("json200 field is nil for job ID %s", jobID)
 	}
 	if getJobResp.JSON200.Errors != nil && len(*getJobResp.JSON200.Errors) != 0 {
-		return JobStatus{}, fmt.Errorf("get job %s responded with errors: %v", jobID, *getJobResp.JSON200.Errors)
+		return []Job{}, fmt.Errorf("get job %s responded with errors: %v", jobID, *getJobResp.JSON200.Errors)
```
theyoprst: `return nil, ... ` ?

internal/controller/soperatorchecks/activecheck_jobs_controller.go
```diff
@@ -167,38 +169,80 @@ func (r *ActiveCheckJobReconciler) Reconcile(
...
-			LastJobStartTime:   slurmJobState.StartTime,
-			LastJobEndTime:     slurmJobState.EndTime,
-		}
+		var failReasons []string
+		var jobName string
+		var submitTime *metav1.Time
+		for _, slurmJob := range slurmJobs {
+			if slurmJob.IsFailedState() {
+				if slurmJob.StateReason != "" {
+					failReasons = append(failReasons, slurmJob.StateReason)
```
theyoprst, line 189: As I can see, node name is not stored here, how can we understand which node failed?

internal/slurmapi/job.go
ChessProfessor: Maybe it shouldn't actually, I was thinking it shouldn't be possible that's why put it here
Do you know if there are any scenarios when state may be empty?

theyoprst: I think it is never empty: I never saw an empty one)

internal/controller/soperatorchecks/activecheck_jobs_controller.go
ChessProfessor: Should we understand that based on ActiveCheck status? Initially we discussed that there will be no information about failed node and only reasons of failure, but I guess we can add it too

internal/slurmapi/job.go
ChessProfessor: Well, I'm not sure what to do with that then :) 
Let's maybe remove until we encounter it

internal/controller/soperatorchecks/activecheck_jobs_controller.go
theyoprst: We should understand it somehow, if there will be drain reason in the node description, that's perfect, I'm closing this conversation then 👍

internal/slurmapi/job.go
theyoprst: Tried to understand job state, it is a bit mask, and as far as I understand it always matches with some state (states), even 0 means "pending".

https://github.com/SchedMD/slurm/blob/1cb50f245f05d851f2383e326db2f20a01820a88/slurm/slurm.h#L249

As an option, we can return an error from `JobFromAPI` if state is absent / empty / any unknown value

internal/controller/soperatorchecks/slurm_nodes_controller.go
```diff
@@ -169,6 +169,7 @@ func (c *SlurmNodesController) processDegradedNode(
 		return c.processKillTaskFailed(ctx, k8sNode, slurmClusterName, node)
 	case consts.SlurmNodeReasonNodeReplacement:
 		return c.processSlurmNodeMaintenance(ctx, k8sNode, slurmClusterName, node.Name)
+	// TODO: support SlurmNodeReasonActiveCheckFailed here
```
theyoprst, line 172: Next PR?

ChessProfessor: That's different part and we have separate ticket to support that
For now just added comment to make sure we don't forget about it

Issue #1008, opened 2025-06-17
[soperatorchecks] Add helm release for maintenance ActiveCheck that deletes old enroot containers


ChessProfessor:

ChessProfessor:
https://github.com/nebius/soperator/pull/1012

PR #1012, opened 2025-06-18
#1008 Add clear_enroot_check ActiveCheck

ChessProfessor:


helm/soperator-activechecks/overrides/clear_enroot_check.yaml
```diff
@@ -0,0 +1,12 @@
+activeCheck:
+  enabled: true
+  checkType: "slurmJob"
+  schedule: "0 */3 * * *"    # every 3 hours
+  slurmJobSpec:
+    eachWorkerJobArray: true
+    sbatchScript: |
+      #!/bin/bash
```
ChessProfessor, line 10: Add `#SBATCH exclusive` here

helm/soperator-activechecks/overrides/clear_enroot_check.yaml
```diff
@@ -0,0 +1,12 @@
+activeCheck:
+  enabled: true
+  checkType: "slurmJob"
+  schedule: "0 */3 * * *"    # every 3 hours
```
rdjjke: Let's make it every 24 hours

helm/soperator-activechecks/overrides/clear_enroot_check.yaml
```diff
@@ -0,0 +1,12 @@
...
+  slurmJobSpec:
+    eachWorkerJobArray: true
+    sbatchScript: |
+      #!/bin/bash
+
+      echo "Cleaning up Enroot containers on node: $(hostname)"
+      enroot list --quiet | xargs -r -n1 enroot remove --force
```
ChessProfessor: Delete only containers with name `pyxis_*`

helm/soperator-activechecks/overrides/clear_enroot_check.yaml
ChessProfessor: Do it every day instead, (at what time?)

helm/soperator-activechecks/overrides/clear_enroot_check.yaml
```diff
@@ -0,0 +1,12 @@
+activeCheck:
+  enabled: true
```
ChessProfessor, line 2: We can also set `suspend: false` so that check will be active right from installation

helm/soperator-activechecks/overrides/clear_enroot_check.yaml
ChessProfessor: Does it matter what time this check starts? We can set 15.00 CET for example?

helm/soperator-activechecks/overrides/clear_enroot_check.yaml
rdjjke: Let's only remove containers named like `pyxis_*`

helm/soperator-activechecks/overrides/clear_enroot_check.yaml
rdjjke, line 10: Let's make jobs exclusive

helm/soperator-activechecks/overrides/clear_enroot_check.yaml
rdjjke: No, it doesn't matter

helm/soperator-activechecks/values.yaml
```diff
@@ -1,6 +1,7 @@
 activeCheck:
   enabled: false
   namespace: "soperator"
+  name: "activecheck"
```
ChessProfessor, line 4: I added this field instead of `{{ .Release.Name }}-activecheck` to be able to control active check name in helm release overrides, otherwise it would be set to any string that user specifies as release name during `helm upgrade`

PR #1013, opened 2025-06-18
Enhance node metrics with state labels and remove job GPU metric

theyoprst:
## Labels renamed
- Rename state labels to use consistent state_ prefix:
  - base_state -> state_base
  - is_drain -> state_is_drain
  - is_maintenance -> state_is_maintenance
  - is_reserved -> state_is_reserved
- Rename compute_instance_id -> instance_id to be consistent with Slurm terms

## GPU-time and failure-related metrics were redesigned:
- Add state labels to slurm_node_fails_total
- Rename slurm_active_node_gpu_seconds_total -> slurm_node_gpu_seconds_total and add the state labels
- Remove slurm_job_alloc_gpu_seconds_total metric entirely (will slurm_node_gpu_seconds_total with proper states instead)

### DIFF
--- a/internal/exporter/collector.go
+++ b/internal/exporter/collector.go
@@ -36,17 +36,17 @@ func NewMetricsCollector(slurmAPIClient slurmapi.Client, soperatorVersion string
 		slurmAPIClient: slurmAPIClient,
 
 		sopClusterInfo: prometheus.NewDesc("soperator_cluster_info", "Soperator cluster information", []string{}, sopClusterInfoConstLabels),
-		nodeInfo:       prometheus.NewDesc("slurm_node_info", "Slurm node info", []string{"node_name", "compute_instance_id", "base_state", "is_drain", "is_maintenance", "is_reserved", "address"}, nil),
+		nodeInfo:       prometheus.NewDesc("slurm_node_info", "Slurm node info", []string{"node_name", "instance_id", "state_base", "state_is_drain", "state_is_maintenance", "state_is_reserved", "address"}, nil),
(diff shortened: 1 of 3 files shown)


internal/exporter/collector.go
```diff
@@ -38,24 +36,19 @@ func NewMetricsCollector(slurmAPIClient slurmapi.Client, soperatorVersion string
 		slurmAPIClient: slurmAPIClient,
 
 		sopClusterInfo: prometheus.NewDesc("soperator_cluster_info", "Soperator cluster information", []string{}, sopClusterInfoConstLabels),
-		nodeInfo:       prometheus.NewDesc("slurm_node_info", "Slurm node info", []string{"node_name", "compute_instance_id", "base_state", "is_drain", "address"}, nil),
+		nodeInfo:       prometheus.NewDesc("slurm_node_info", "Slurm node info", []string{"node_name", "compute_instance_id", "base_state", "is_drain", "is_maintenance", "is_reserved", "address"}, nil),
```
rdjjke: What about renaming "instance_id" to "compute instance id"?
As the first is a Slurm term, while the second is a Nebius one

rdjjke: Did you check how it works when one worker is added to many partitions? Won't we have duplicates?

theyoprst: > Did you check how it works when one worker is added to many partitions? Won't we have duplicates?

I didn't, it doesn't make sense. API returns list of all nodes, we don't ask them per partition. In the response, for each node we have list of partitions, and they are ignored just not to duplicate.

rdjjke: Can we also add node comment field?

theyoprst: Done

theyoprst: It's better not to add the comment to the metric just not to increase cardinality uncontrollably.

PR #1036, opened 2025-06-20
NCCL debug plugin deployment

dstaroff:


images/common/scripts/complement_jail.sh
```diff
@@ -124,12 +124,19 @@ pushd "${jaildir}"
...
-    mount --bind "/usr/lib/${ALT_ARCH}-linux-gnu/slurm/spank_pyxis.so" "${SLURM_LIB_PATH}/spank_pyxis.so"
+    mount --bind "/${SLURM_LIB_PATH}/spank_pyxis.so" "${SLURM_LIB_PATH}/spank_pyxis.so"
 
     echo "Bind-mount slurm configs"
     mkdir -p etc/slurm
     mount --bind /mnt/jail/slurm etc/slurm
 
+    echo 'Creating Soperator output directory'
+    ( \
+      umask 000 ; \
```
rdjjke, line 135: You can put it simpler:
```
( 
  mkdir -p opt/soperator-outputs
  chmod 777 opt/soperator-outputs
)
```

(no need for `;` and `\`, and no need to do `umask` if you already do `chmod`)

rdjjke: Hm, maybe only if we change the path in the future. Ok we can leave umask then

rdjjke (approved):

images/common/scripts/complement_jail.sh
dstaroff: Got it. Will simplify it in the next PR

PR #1047, opened 2025-06-23
#1044 Add all_reduce_perf nccl check

ChessProfessor:


helm/soperator-activechecks/overrides/all_reduce_perf_nccl.yaml
```diff
@@ -0,0 +1,31 @@
...
+  enabled: true
+  suspend: false
+  name: "all-reduce-perf-nccl"
+  checkType: "slurmJob"
+  schedule: "0 0 * * *"    # every 24 hours
+  slurmJobSpec:
+    eachWorkerJobArray: true
+    sbatchScript: |
+      #!/bin/bash
+
```
ChessProfessor, line 19: Not sure if it's better to put `#SBATCH --exclusive` here

rdjjke: No, it's not neccessary

helm/soperator-activechecks/overrides/all_reduce_perf_nccl.yaml
```diff
@@ -0,0 +1,31 @@
...
+      echo "Checking for running GPU processes..."
+      if [[ -n "$(nvidia-smi --query-compute-apps=pid --format=csv,noheader | grep -v '^ *$')" ]]; then
+        echo "Another GPU process is currently running. Exiting."
+        exit 0
+      fi
+
+      echo "Running all_reduce_without_ib check on $(hostname)..."
+      HC_OUTPUT=$(srun bash -c "health-checker run -e soperator -p 8xH100 -n all_reduce_without_ib --json-log")
```
rdjjke: Can we print the HC output?

helm/soperator-activechecks/overrides/all_reduce_perf_nccl.yaml
```diff
@@ -0,0 +1,31 @@
+activeCheck:
+  enabled: true
+  suspend: false
+  name: "all-reduce-perf-nccl"
+  checkType: "slurmJob"
+  schedule: "0 0 * * *"    # every 24 hours
+  slurmJobSpec:
```
rdjjke, line 7: What is the working directory from which these Slurm jobs are executed? Slurm job outputs will be stored there

ChessProfessor: It will be just root directory
What specific directory would be better for that? We can create some separate directory for check outputs and put all outputs of all future checks there, wdyt?

helm/soperator-activechecks/overrides/all_reduce_perf_nccl.yaml
ChessProfessor: Also I hardcode `-p 8xH100` here. Do we want to support different options and if so how do we want to distinguish clusters with different gpu types?

helm/soperator-activechecks/overrides/all_reduce_perf_nccl.yaml
```diff
@@ -0,0 +1,31 @@
+activeCheck:
+  enabled: true
+  suspend: false
+  name: "all-reduce-perf-nccl"
+  checkType: "slurmJob"
+  schedule: "0 0 * * *"    # every 24 hours
```
rdjjke: Let's run it every 8 hours

helm/soperator-activechecks/overrides/all_reduce_perf_nccl.yaml
rdjjke: Is it launched by soperatorchecks user? If so, let's write outputs to its home directory (or a subdirectory) there.

Job's working directory is the one you submit the job from. But it can also be controlled via sbatch `--chdir` argument. Maybe we can hardcode it to thу job submission API call

helm/soperator-activechecks/overrides/all_reduce_perf_nccl.yaml
```diff
@@ -0,0 +1,31 @@
...
+  slurmJobSpec:
+    eachWorkerJobArray: true
+    sbatchScript: |
+      #!/bin/bash
```
rdjjke, line 10:
```
#SBATCH --deadline="now+6hours"
#SBATCH --time=10:00 # 10 minutes
#SBATCH --gpus-per-node=8
#SBATCH --cpus-per-task=16
#SBATCH --cpu-bind="verbose,cores"
#SBATCH --mem=32G
#SBATCH --no-requeue
#SBATCH --uid=soperatorchecks
#SBATCH --chdir=/home/soperatorchecks # But better to hard-code for all active checks
#SBATCH --output=/home/soperatorchecks/%x_%A_%j_%n.out` # <job_name>_<job_array_id>_<job_id>_<hostname>.out
#SBATCH --error=/home/soperatorchecks/%x_%A_%j_%n.out`
#SBATCH -vvv # But better to hard-code for all active checks (there should be a field named "verbose" and we should set it to "3")
```

helm/soperator-activechecks/overrides/all_reduce_perf_nccl.yaml
rdjjke: It also supports at least H200 or L40s. So let's drop it.
It just should be turnable on and off for now.

ChessProfessor: Unfortunately we have to specify platform when using health-checker lib, it cannot be left empty. So if we want to support different gpu types here we have to figure out it on the go and pass this info to the script

rdjjke: Then let's grep it from nvidia-smi.

In the bash script:
```
platform=""
gpus_on_node=$(nvidia-smi --query-gpu=name --format=csv,noheader | sort | uniq -c)
if [[ "${gpus_on_node}" == *"8 NVIDIA H100"* ]]; then
    platform = "8xH100"
elif [[ "${gpus_on_node}" == *"8 NVIDIA H200"* ]]; then
    platform = "8xH200"
else
    echo "Unsupported platform"
    exit 0
fi
# ...
# srun call
```

internal/consts/activecheck.go
```diff
@@ -3,6 +3,9 @@ package consts
 const (
 	ActiveCheckFinalizer                    = "slurm.nebius.ai/activecheck-finalizer"
 	SlurmNodeReasonActiveCheckFailedUnknown = "Soperator activecheck unknown: node drain process"
+	SlurmNodeReasonAllReducePerfNcclFailed  = "[HC] All reduce perf nccl failed"
+
+	AllReducePerfNcclActiveCheckName = "all-reduce-perf-nccl"
```
rdjjke: Can we instead take the name from the ActiveCheck CR?

internal/consts/activecheck.go
```diff
@@ -3,6 +3,9 @@ package consts
 const (
 	ActiveCheckFinalizer                    = "slurm.nebius.ai/activecheck-finalizer"
 	SlurmNodeReasonActiveCheckFailedUnknown = "Soperator activecheck unknown: node drain process"
+	SlurmNodeReasonAllReducePerfNcclFailed  = "[HC] All reduce perf nccl failed"
```
rdjjke: Can we generalize it to `[HC] failed <active_check_name> [slurm_job]`?

internal/consts/activecheck.go
ChessProfessor: Job name and ActiveCheck name are the same and we compare this constant with ActiveCheck name to determine the drain reason

internal/consts/activecheck.go
ChessProfessor: Slurm job id? Yes, we can do that

internal/consts/activecheck.go
ChessProfessor: If we want to generalize reason to `[HC] failed <active_check_name> [slurm_job_id]` we can just remove this constant and put active check name in the reason directly

helm/soperator-activechecks/overrides/all_reduce_perf_nccl.yaml
```diff
@@ -0,0 +1,57 @@
...
+      #SBATCH --deadline="now+6hours"
+      #SBATCH --time=10:00
+      #SBATCH --gpus=8
+      #SBATCH --mem-per-gpu=16G
+      #SBATCH --gpus-per-node=8
+      #SBATCH --cpus-per-task=16
+      #SBATCH --cpu-bind="verbose,cores"
+      #SBATCH --mem=128G
+      #SBATCH --no-requeue
+      #SBATCH --uid=soperatorchecks
```
ChessProfessor: This one should be used while calling sbatch in the image

helm/soperator-activechecks/overrides/all_reduce_perf_nccl.yaml
```diff
@@ -0,0 +1,57 @@
...
+      #SBATCH --chdir=/home/soperatorchecks
```
ChessProfessor: And this one

helm/soperator-activechecks/overrides/all_reduce_perf_nccl.yaml
```diff
@@ -0,0 +1,57 @@
...
+      #SBATCH --output=/home/soperatorchecks/%x_%A_%j_%n.out
+      #SBATCH --error=/home/soperatorchecks/%x_%A_%j_%n.out
+      #SBATCH -vvv
```
ChessProfessor: This one should be configured via ActiveCheck spec and also passed in sbatch command

images/slurm_check_job/slurm_check_job_entrypoint.sh
```diff
@@ -26,7 +26,12 @@ if [[ "$EACH_WORKER_JOB_ARRAY" == "true" ]]; then
     SLURM_JOB_ID=$(/opt/bin/slurm/slurm_submit_array_job.sh | tail -n 1)
 else
     echo "Submitting regular Slurm job..."
-    SLURM_OUTPUT=$(/usr/bin/sbatch --parsable /opt/bin/sbatch.sh)
+    SLURM_OUTPUT=$(/usr/bin/sbatch --parsable \
+      --job-name="$ACTIVE_CHECK_NAME" \
+      --chdir=/home/soperatorchecks \
+      --uid=soperatorchecks \
+      -vvv \
```
ChessProfessor: Created a ticket to move it to ActiveCheck spec https://github.com/nebius/soperator/issues/1057

helm/soperator-activechecks/overrides/all_reduce_perf_nccl.yaml
```diff
@@ -0,0 +1,55 @@
...
+  name: "all-reduce-perf-nccl"
+  checkType: "slurmJob"
+  schedule: "0 */8 * * *"   # every 8 hours
+  slurmJobSpec:
+    eachWorkerJobArray: true
+    sbatchScript: |
+      #!/bin/bash
+      #SBATCH --deadline="now+6hours"
+      #SBATCH --time=10:00
+      #SBATCH --gpus=8
```
rdjjke: Let's use `--gpus-per-node=8`, `--gpus` can be confusing in general because it's the total number of GPUs for a job (from any node).
In this specific case it will work, but I'm afraid someone can copy this to a context where it will be wrong.

helm/soperator-activechecks/overrides/all_reduce_perf_nccl.yaml
```diff
@@ -0,0 +1,55 @@
...
+      #SBATCH --mem-per-gpu=16G
```
rdjjke: Let's use `--mem` (this is "per node") because 32G of memory in total is more than enough

helm/soperator-activechecks/overrides/all_reduce_perf_nccl.yaml
```diff
@@ -0,0 +1,55 @@
...
+      elif [[ "${gpus_on_node}" == *"8 NVIDIA H200"* ]]; then
+        platform="8xH200"
+      else
+        echo "Unsupported platform"
+        exit 0
+      fi
+      echo "Platform found: $platform"
+
+      echo "Running all_reduce_without_ib check on $(hostname)..."
+      HC_OUTPUT=$(srun bash -c "health-checker run -e soperator -p $platform -n all_reduce_without_ib --json-log")
```
rdjjke: Let's either add `--cpu-bind=verbose,cores` here or `export SLURM_CPU_BIND="verbose,cores"` before the `srun` call

helm/soperator-activechecks/overrides/all_reduce_perf_nccl.yaml
```diff
@@ -0,0 +1,55 @@
...
+      echo "Health checker output: $HC_OUTPUT"
```
rdjjke: You capture the HC exit code (`HC_EXIT_CODE=$?`) after this `echo` so the exit code is actually of the echo command, not srun.

helm/soperator-activechecks/overrides/all_reduce_perf_nccl.yaml
```diff
@@ -0,0 +1,55 @@
...
+      HC_EXIT_CODE=$?
+      echo "Health checker exit code: $HC_EXIT_CODE"
```
rdjjke: Let's put it as "Health checker job step exit code: $HC_EXIT_CODE"

Because exit code 1 should be returned by srun unchanged, but it's not true for any possible exit code. For example you can get an exit code like "235" which will show some Slurm error, not health checker one.

rdjjke (approved):

PR #1078, opened 2025-06-26
Retry if slurm-job-id is not yet set

ChessProfessor:


ChessProfessor:
Nevermind, it's not necessary

PR #1103, opened 2025-06-30
[FIX] Replace `kebabcase` with custom function

dstaroff:
Otherwise, kebabcase turns something like `slurm-k9` to `slurm-k-9`


dstaroff:
Maybe we should use it in other Helm charts as well.

PR #1108, opened 2025-07-01
Reorganize log directory structure by worker node

theyoprst:
This change implements a worker-separated directory structure for logs
to reduce filesystem contention on the shared jail storage.

## Changes:
- NCCL logs: `/opt/soperator-outputs/%h/nccl_logs/` (using Slurm `%h` substitution)
- Remove worker name prefix from filenames (redundant with directory structure)
- Update OpenTelemetry collector to read from worker-specific directories
- Update documentation to reflect new layout
- Add hostname substitution support to NCCL plugin for `%h` replacement
- Update CRD documentation for new filename format

## Benefits:
- Eliminates cross-worker filesystem contention
- Enables parallel log collection
- Simplifies filename patterns
- Improves shared filesystem performance

### DIFF
--- a/api/v1/slurmcluster_types.go
+++ b/api/v1/slurmcluster_types.go
@@ -271,7 +271,7 @@ type PluginConfigNcclDebug struct {
 
 	// OutputToFile defines whether to additionally redirect `NCCL_DEBUG` outputs to the output file.
 	// Output filename will have the following format:
-	//  <WORKER_NAME>.<JOB_ID>.<STEP_ID>.out
+	//  <JOB_ID>.<STEP_ID>.out
 	//
 	// +kubebuilder:validation:Optional
 	// +kubebuilder:default=true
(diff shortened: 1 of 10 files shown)


images/slurm_check_job/slurm_check_job_entrypoint.sh
```diff
@@ -22,7 +22,8 @@ echo "Bind-mount /opt/bin/sbatch.sh script"
 mount --bind /opt/bin/sbatch.sh opt/bin/sbatch.sh
 
 echo "Create directory for slurm job outputs"
-(umask 000; mkdir -p /mnt/jail/opt/soperator-outputs/slurm_jobs)
+WORKER_NAME=$(hostname)
+(umask 000; mkdir -p /mnt/jail/opt/soperator-outputs/${WORKER_NAME}/{nccl_logs,slurm_jobs,slurm_scripts})
```
itechdima: I'm not sure `$(hostname)` is the right thing here
it's k8s job that could be scheduled on any node

I think it should be done somewhere where we can determine which worker would be used
maybe it's `slurm_submit_array_job.sh` and `sbatch.sh`?

theyoprst: Thanks, rolled back this change, it's fixed here: https://github.com/nebius/soperator/pull/1111

asteny (approved):

PR #1159, opened 2025-07-04
Make jail log collector great again

theyoprst:
## Summary

- Reverts worker-separated directory structure introduced in PRs #1106, #1107, #1108, and #1111.
- Migrates jail log collector from DaemonSet to Deployment running on system nodes
- Re-enables jail logs collection by default

## Changes

1. Simplified logging structure: Removed /worker-X/ subdirectories, keeping only type-based directories (/nccl_logs/, /slurm_jobs/, /slurm_scripts/)
2. Node info in filenames: Node information now included in filenames using %N pattern instead of directory paths
3. Deployment optimization: Jail log collector now runs as a single Deployment on system nodes instead of DaemonSet on all nodes
4. Default configuration: Jail logs collection enabled by default in values.yaml

## Rationale

Performance optimization: A single collector scanning 3 directories (/nccl_logs/, /slurm_jobs/, /slurm_scripts/) is much more cost-efficient than scanning O(count of workers) directories. This significantly reduces I/O operations and improves collector performance at scale.

Issue: #1119

### DIFF
--- a/helm/soperator-fluxcd/templates/opentelemetry-collector-jail-logs.yaml
+++ b/helm/soperator-fluxcd/templates/opentelemetry-collector-jail-logs.yaml
@@ -1,6 +1,5 @@
 {{- if and .Values.observability.enabled
             .Values.observability.opentelemetry.enabled
-            .Values.observability.opentelemetry.logs.values.hcOutputEnabled
             .Values.observability.opentelemetry.logs.values.jailLogs.enabled }}
 apiVersion: helm.toolkit.fluxcd.io/v2
 kind: HelmRelease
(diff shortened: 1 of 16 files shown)


helm/soperator-fluxcd/templates/opentelemetry-collector-jail-logs.yaml
```diff
@@ -49,45 +48,102 @@ spec:
...
         batch:
           send_batch_max_size: 700
           send_batch_size: 250
         k8sattributes:
+          pod_association:
+          - sources:
+            - from: resource_attribute
+              name: k8s.pod.name
+            - from: resource_attribute
+              name: k8s.namespace.name
```
Uburro, line 144: how otel collector collect id or node name?

theyoprst: It's done in the k8sattributes processor: this processor collects all the pods information automatically.
And then it matches "k8s.namespace.name" + "k8s.pod.name" attributes of pods with exactly the same attributes of the log, the latter are added by the filelog collector, see the other configuration in this file:

```yaml
            - type: move
              from: attributes.slurm_node_name
              to: resource["k8s.pod.name"]
            - type: add
              field: resource["k8s.namespace.name"]
              value: {{ .Values.slurmCluster.namespace | default "soperator" | quote }}
```

See also https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/k8sattributesprocessor, is has some configuration examples

asteny (approved):

PR #1190, opened 2025-07-08
[fix] remove node topology

Uburro:


internal/controller/topologyconfcontroller/nodetopology_controller.go
```diff
@@ -103,8 +104,14 @@ func (r *NodeTopologyReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	logger := log.FromContext(ctx).WithName(NodeTopologyReconcilerName)
 	logger.Info("Starting reconciliation", "node", req.Name)
 
-	node, err := r.getNode(ctx, req.Name, logger)
+	node, err := r.getNode(ctx, req.Name)
 	if err != nil {
+		if errors.IsNotFound(err) {
```
theyoprst, line 109: getNode never returns not found, it returns `nil, nil`:

```go
func (r *NodeTopologyReconciler) getNode(ctx context.Context, nodeName string, logger logr.Logger) (*corev1.Node, error) {
	node := &corev1.Node{}
	nodeKey := client.ObjectKey{Name: nodeName}

	if err := r.Client.Get(ctx, nodeKey, node); err != nil {
		if client.IgnoreNotFound(err) == nil {
			// Node not found - it was deleted
			return nil, nil
		}
```

theyoprst: Sorry, my bad, that's not true :/

theyoprst: Indeed, it never returns it `return nil, client.IgnoreNotFound(err)`

internal/controller/topologyconfcontroller/nodetopology_controller.go
```diff
@@ -171,17 +177,36 @@ func ExtractTierLabels(k8sNodeLabels map[string]string, topologyLabelPrefix stri
...
+		return nil // Nothing to remove if Data is nil
+	}
+	if _, exists := configMap.Data[nodeName]; exists {
+		delete(configMap.Data, nodeName)
+		if err := r.Client.Update(ctx, configMap); err != nil {
+			return fmt.Errorf("update ConfigMap %s/%s: %w", r.Namespace, configMap.ObjectMeta.Name, err)
+		}
+		return nil
+	}
+	return fmt.Errorf("node %s not found in ConfigMap %s/%s", nodeName, r.Namespace, configMap.ObjectMeta.Name)
```
theyoprst: I think if node not found, we should not return error: what if the event came twice for some reason? Maybe lust log the message and return nil?

Issue #1083, opened 2025-06-26
[soperatorchecks] Fix ActiveCheck status update for manual runs of CronJob


ChessProfessor:
Currently annotation `slurm.nebius.ai/activecheck` with the name of the ActiveCheck is not passed to k8s Job created manually from CronJob

PR #1207, opened 2025-07-10
#1083 Fix active checks for manual runs

ChessProfessor:


internal/controller/soperatorchecks/activecheck_jobs_controller.go
```diff
@@ -122,7 +123,11 @@ func (r *ActiveCheckJobReconciler) Reconcile(
 		return ctrl.Result{}, err
 	}
 
-	activeCheckName := k8sJob.Annotations[consts.AnnotationActiveCheckKey]
+	activeCheckName, err := r.getActiveCheckNameFromJob(ctx, k8sJob)
+	if err != nil {
+		logger.Error(err, "Failed to get ActiveCheckName")
```
theyoprst: Maybe return wrapped error instead?

```go
return ctrl.Result{}, fmt.Errorf("get active check name: %w", err)
```

Otherwise, it will be logged here, plus on higher level.

theyoprst (dismissed):

PR #1232, opened 2025-07-14
undrain slurm nodes in case of false positive gres/gpu

itechdima:


internal/controller/soperatorchecks/slurm_nodes_controller.go
```diff
@@ -509,6 +521,43 @@ func (c *SlurmNodesController) processMaintenance(
...
+	if err != nil {
+		return fmt.Errorf("get node: %w", err)
+	}
+
+	if _, ok := slurmNode.States[api.V0041NodeStateINVALIDREG]; ok {
+		if err := c.reconfigure(ctx, slurmClusterName); err != nil {
+			return err
+		}
+
+		// sleep because reconfigure is async (by default MessageTimeout=60)
```
theyoprst, line 550: Maybe log a message for better observability?

PR #1243, opened 2025-07-16
fix rebooter incorrect limit cpu

Uburro:


internal/render/nodeconfigurator/container.go
```diff
@@ -58,9 +58,9 @@ func renderContainerRebooter(rebooter slurmv1alpha1.Rebooter) corev1.Container {
 			"-log-format=" + rebooter.LogFormat,
 		},
 		Resources: createResourceRequirements(
-			*rebooter.Resources.Limits.Cpu(),
 			*rebooter.Resources.Requests.Cpu(),
 			*rebooter.Resources.Requests.Memory(),
+			*rebooter.Resources.Limits.Memory(),
```
ChessProfessor: Is that correct? I see it expects (limitsMemory, requestsCPU, requestsMemory), but we pass here (requestsCPU, requestsMemory, limitsMemory)

PR #1244, opened 2025-07-16
Fix empty user_name labels in job_info metric

### LINKED ISSUE #1238: [Soperator exporter] Metric job_info: user_name label is empty for all users except root
It is requored to mount jail and use users infromation from there.

theyoprst:
## Summary
- Add user_id label to job_info metric in soperator exporter for fallback when user_name is empty
- Enable REST container to access jail passwd files for username resolution

Closes https://github.com/nebius/soperator/issues/1238

### DIFF
--- a/images/restd/slurmrestd_entrypoint.sh
+++ b/images/restd/slurmrestd_entrypoint.sh
@@ -2,6 +2,14 @@
 
 set -e # Exit immediately if any command returns a non-zero error code
 
+echo "Link users from jail"
+for file in passwd group shadow gshadow; do
+  rm -f /etc/$file
+  ln -s /mnt/jail/etc/$file /etc/$file
+done
+
+chown -h 0:42 /etc/{shadow,gshadow}
+
 echo "Symlink slurm configs from K8S config map"
 rm -rf /etc/slurm && ln -s /mnt/slurm-configs /etc/slurm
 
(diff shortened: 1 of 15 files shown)


images/restd/slurmrestd_entrypoint.sh
```diff
@@ -2,6 +2,14 @@
 
 set -e # Exit immediately if any command returns a non-zero error code
 
+echo "Link users from jail"
+for file in passwd group shadow gshadow; do
+  rm -f /etc/$file
```
asteny, line 7: In all other images we removed users in Docker file and link them in entrypoint

theyoprst: Why is it so important?

theyoprst: > In all other images we removed users in Docker file and link them in entrypoint

I don't think it's good idea to split /etc/passwd file creation by different files, that's why I did it in one place. Any reasons I should split it?

rdjjke (approved):

images/restd/slurmrestd_entrypoint.sh
asteny: My point is not about splitting... I think we need to do that in one way in all places)

PR #1253, opened 2025-07-17
Add macOS support in FileStore

mcheshkov:
Also setup simple job in CI with basic checks to catch errors like this early

### DIFF
--- a/.github/workflows/one_job.yml
+++ b/.github/workflows/one_job.yml
@@ -86,6 +86,50 @@ jobs:
           echo "Version: $VERSION"
           echo "Unstable: $UNSTABLE"
 
+  lint:
+    runs-on: self-hosted
+
+    steps:
+      - name: Harden Runner
+        uses: step-security/harden-runner@6c439dc8bdf85cadbbce9ed30d1c7b959517bc49 # v2.12.2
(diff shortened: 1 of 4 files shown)


.github/workflows/one_job.yml
```diff
@@ -86,6 +86,35 @@ jobs:
...
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: Install GO
+        uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5 # v5.5.0
+        with:
+          go-version-file: 'go.mod'
+          cache: false
+
+      - name: Run make vet
+        run: GOARCH=${{ matrix.arch }} GOOS=${{ matrix.os }} make vet
```
theyoprst: I'd propose using `make lint` instead: at least it will be cached. And it also will "catch" build bug. Plus `govet` linter is already turned on.

Also:
- call it "lint" instead of "hygiene".
- remove `make lint` call from a separate step
- remove `make vet` command at all

theyoprst (approved):

Uburro (approved):

Issue #1268, opened 2025-07-18
Fix Slurm controller failover


rdjjke:
- Add a K8s service for controllers and don't use the headless service
- Perhaps, remove the readiness probe

PR #1273, opened 2025-07-21
Fix Slurm controller failover #1268

Uburro:
Here’s a summary of the changes in this pull request:

**internal/controller/clustercontroller/controller.go**
- Refactors the reconciliation logic for controller services to create a separate Kubernetes Service for each controller pod (per-pod services) based on the number of replicas.
- Adds logic to clean up excess services that are no longer needed.
- Removes repetitive error logging statements (error logging is now mostly handled by returning errors).
- Updates usage of the Service rendering function to support per-pod labels and naming.

**internal/render/common/configmap.go**
- Changes how SlurmctldHost entries are generated in config files:
  - Uses the pattern `<statefulset-name>-<index>` for both host and FQDN, instead of relying on the previous `naming.BuildServiceHostFQDN` function.

**internal/render/controller/container.go**
- Removes the readiness probe from the Slurmctld container definition.

**internal/render/controller/service.go**
- Updates the `RenderService` function to allow custom labels and selectors for per-pod services.
- Service names now use the provided name parameter (for per-pod differentiation).
- Selectors and labels can include additional data for targeting specific pods.

**In summary:**  
This PR refactors the controller Service logic to create individual Kubernetes Services for each controller pod, adjusts config rendering to match the new naming scheme, removes redundant error logs, and enables more flexible service labeling and selection. The readiness probe for controller containers is also removed.

#1268

### DIFF
--- a/internal/controller/clustercontroller/controller.go
+++ b/internal/controller/clustercontroller/controller.go
@@ -1,6 +1,7 @@
 package clustercontroller
 
 import (
+	"cmp"
 	"context"
 	"fmt"
 	"strings"
@@ -63,11 +64,7 @@ func (r SlurmClusterReconciler) ReconcileControllers(
 					stepLogger := log.FromContext(stepCtx)
 					stepLogger.V(1).Info("Reconciling")
 
-					replicas := int32(1)
-					if clusterValues.NodeController.StatefulSet.Replicas > 0 {
-						replicas = clusterValues.NodeController.StatefulSet.Replicas
-					}
-
+					replicas := cmp.Or(clusterValues.NodeController.StatefulSet.Replicas, 1)
 					createdServices := map[string]bool{}
 
 					for i := int32(0); i < replicas; i++ {
--- a/internal/controller/clustercontroller/controller.go
+++ b/internal/controller/clustercontroller/controller.go
@@ -3,6 +3,7 @@ package clustercontroller
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	kruisev1b1 "github.com/openkruise/kruise-api/apps/v1beta1"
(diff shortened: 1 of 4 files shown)


internal/render/common/configmap.go
```diff
@@ -58,13 +58,8 @@ func generateSlurmConfig(cluster *values.SlurmCluster) renderutils.ConfigFile {
 	res.AddComment("")
 	// example: SlurmctldHost=controller-0(controller-0.controller.slurm-poc.svc.cluster.local)
```
theyoprst: Looks like this example now should be updated to something like this?

```
SlurmctlHost=controller-0(controller-0)
```

internal/controller/clustercontroller/controller.go
```diff
@@ -62,17 +63,49 @@ func (r SlurmClusterReconciler) ReconcileControllers(
 					stepLogger := log.FromContext(stepCtx)
 					stepLogger.V(1).Info("Reconciling")
 
-					desired := controller.RenderService(clusterValues.Namespace, clusterValues.Name, &clusterValues.NodeController)
-					stepLogger = stepLogger.WithValues(logfield.ResourceKV(&desired)...)
-					stepLogger.V(1).Info("Rendered")
+					replicas := int32(1)
+					if clusterValues.NodeController.StatefulSet.Replicas > 0 {
+						replicas = clusterValues.NodeController.StatefulSet.Replicas
```
theyoprst: (nit) You can use [cmp.Or](https://pkg.go.dev/cmp#Or) for such things, e.g.:

```go
replicas := cmp.Or(clusterValues.NodeController.StatefulSet.Replicas, 1)
```

PR #1274, opened 2025-07-22
#1270 Drain slurm nodes right after Active Check failure

ChessProfessor:


internal/controller/soperatorchecks/activecheck_jobs_controller.go
```diff
@@ -253,6 +270,10 @@ func (r *ActiveCheckJobReconciler) Reconcile(
 			logger.Error(err, "Failed to reconcile ActiveCheckJob")
 			return ctrl.Result{}, fmt.Errorf("reconciling ActiveCheckJob: %w", err)
 		}
+
+		if requeue {
```
Uburro, line 274: you get 2 reconcile.
first  when condition changed and second form requeue

ChessProfessor: This is not activecheck reconciler, this one reconciles jobs created by activecheck cronjob.
I guess it's triggered only when slurm-job-id field is set to k8s job annotations

Uburro: suppose to [here](https://github.com/nebius/soperator/blob/651d137567eb8a41e6bd168052ce95529b83bb68/internal/controller/soperatorchecks/activecheck_jobs_controller.go#L61C45-L61C62) reconcile get events for every changes. requeue is not necessary this case

ChessProfessor: Yes, every changes to k8s Job, but I don't get where the job gets updated in this reconciler. I see only Active Check status update, which won't trigger the same reconciler

Uburro (approved):

PR #1287, opened 2025-07-23
[WIP] Jailed config CR

mcheshkov:


internal/controller/sconfigcontroller/fs.go
```diff
@@ -0,0 +1,171 @@
...
+
+var _ Fs = &PrefixFs{}
+
+func (pfs *PrefixFs) addPrefix(path string) (string, error) {
+	if !filepath.IsAbs(path) {
+		return "", fmt.Errorf("path %q is not an absolute path", path)
+	}
+	// Clean for absolute paths should remove all of `.` and `..` sections, should be safe to join w.r.t path traversing
+	path = filepath.Clean(path)
+	return filepath.Join(pfs.Prefix, path), nil
```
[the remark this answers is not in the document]
mcheshkov: Yes but no, exactly because clean is called before joining, and path is absolute
But a check should not hurt

internal/controller/sconfigcontroller/fs.go
```diff
@@ -0,0 +1,171 @@
...
+	delayBetweenWriteAndReconfigure = direntCacheTTL + 1*time.Second
+)
+
+type PrefixFs struct {
+	Prefix string
+}
+
+var _ Fs = &PrefixFs{}
+
+func (pfs *PrefixFs) addPrefix(path string) (string, error) {
```
mcheshkov: This should not only check for absolute paths, but resolve symlinks as well, both from functionality and security PoV.
* From jail user perspective, symlink should point inside jail, not outside, so when JailedConfig would try to write to path like `/dir/symlink_to_dir/config.txt` symlink should be resolved as-if inside jail
* From sconfigcontroller perspective, it is not safe to just write into arbitrary directory, because user that has access to jail can create symlink pointing to, say, `/etc`, and sconfigcontroller should not allow to write to `/etc` in it's rootfs. Same goes for symlink like `../../..`: inside jail this should be capped to jail root, but outside of jail there would be no syntactic path traversal, but semantically it is still there.

There's no simple way to resolve symlinks relative to jail root.
One way would be to copy-paste `filepath.EvalSymlinks`, make it prepend prefix on every absolute symlink, and clean and cap every relative symlink.
Another would be to `chroot`/`pivot_root` into jail before working with FS. This would simplify most of issues: prefix management, path traversal, symlink resolving.

For now I'll just use simple split on path separator, and deny any symlinks or path traversal, I think it should be safe enough for Linux.

internal/controller/sconfigcontroller/jailedconfig_controller.go
```diff
@@ -0,0 +1,551 @@
...
+		}
+
+		if len(nodeToStartBefore) == 0 {
+			break
+		}
+
+		select {
+		case <-ctx.Done():
+			return context.DeadlineExceeded
+		case <-r.clock.After(1 * time.Second):
```
itechdima: maybe 5 sec?
I'm not sure what could happen on a big cluster

or even better - make this duration configurable

itechdima (dismissed):

Uburro (approved):

PR #1323, opened 2025-07-28
#1312 Add ib-write-bw/lat cpu checks

ChessProfessor:


itechdima (approved):

helm/soperator-activechecks/scripts/ib-write-bw-cpu.sh
```diff
@@ -0,0 +1,37 @@
...
+  --container-mounts=$(which health-checker):/usr/local/bin/health-checker --cpu-bind=verbose,cores \
+  bash -c "health-checker run -e soperator -p 8xB200 -n ib_write_bw --json-log")
+HC_EXIT_CODE=$?
+
+echo "Health checker output: $HC_OUTPUT"
+echo "Health checker job step exit code: $HC_EXIT_CODE"
+HC_STATUS=$(echo "$HC_OUTPUT" | awk '/^\s*{/,/^\s*}/' | jq -r '.status')
+
+echo "Health checker status: $HC_STATUS"
+if [[ "$HC_STATUS" == "ERROR" && $HC_EXIT_CODE -eq 1 ]]; then
```
rdjjke: Status "ERROR" means that the health-checker itself broke, it doesn't mean that the health check failed

rdjjke: In case of the ERROR status, we should log this but exit with code 0

PR #1344, opened 2025-07-30
[ha controller] add placeholder and single replicas

### LINKED ISSUE #1330: [ha controller] add placeholder and single replicas
- Creates the Slurm cluster with a single replica
- Creates placeholder pods if the size is greater than one. These pods don't perform any actions and have lower priority than the main container. Their purpose is to speed up image loading when the controller pod is relocated, in order to avoid downtime (e.g., in case the registry becomes unavailable or the network fails)
- The main pod should have a liveness probe with scontrol ping in case the Health Check Controller becomes unavailable for any reason

Uburro:
Fixes #1330

### DIFF
--- a/internal/render/controller/daemonset.go
+++ b/internal/render/controller/daemonset.go
@@ -15,7 +15,7 @@ import (
 )
 
 // RenderDaemonSet renders new [appsv1.DaemonSet] containing additional Slurm controller pods
-func RenderDaemonSet(
+func RenderPlaceholderDaemonSet(
 	namespace,
 	clusterName string,
 	nodeFilters []slurmv1.K8sNodeFilter,
@@ -58,7 +58,7 @@ func RenderDaemonSet(
 					NodeSelector: nodeFilter.NodeSelector,
 					Tolerations:  nodeFilter.Tolerations,
 					InitContainers: []corev1.Container{
-						common.RenderContainerMungeSleep(&controller.ContainerMunge),
+						common.RenderPlaceholderContainerMunge(&controller.ContainerMunge),
 					},
 					Containers: append(
 						[]corev1.Container{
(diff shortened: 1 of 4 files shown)


internal/render/common/configmap.go
```diff
@@ -56,12 +56,8 @@ func generateSlurmConfig(cluster *values.SlurmCluster) renderutils.ConfigFile {
 
 	res.AddProperty("ClusterName", cluster.Name)
 	res.AddComment("")
-	// example: SlurmctldHost=controller-0(controller-0)
-	// beause in kubernetes, dns uses the dot notation to separate the service name and the namespace
-	for i := int32(0); i < cluster.NodeController.Size; i++ {
-		svcName := fmt.Sprintf("%s-%d", cluster.NodeController.StatefulSet.Name, i)
-		res.AddProperty("SlurmctldHost", fmt.Sprintf("%s(%s)", svcName, svcName))
-	}
+	svcName := cluster.NodeController.Service.Name
```
rdjjke, line 59: As multi-controller setups aren't supported anymore, can we get rid of the size option for controller nodes?

rdjjke (dismissed):

internal/render/controller/daemonset.go
```diff
@@ -0,0 +1,98 @@
...
+
+	slurmv1 "nebius.ai/slurm-operator/api/v1"
+	"nebius.ai/slurm-operator/internal/consts"
+	"nebius.ai/slurm-operator/internal/render/common"
+	"nebius.ai/slurm-operator/internal/utils"
+	"nebius.ai/slurm-operator/internal/values"
+)
+
+// RenderDaemonSet renders new [appsv1.DaemonSet] containing additional Slurm controller pods
+func RenderDaemonSet(
```
theyoprst: Naming suggestion: `RenderPlaceholderDaemonSet` to give the idea that it is for a placeholder only. 
Also I'd rename `RenderContainerMungeSleep` -> `RenderPlaceholderContainerMunge` for the same reason

(rename test names as well)

Uburro: done

PR #1347, opened 2025-07-30
Bump container toolkit and do not install it into the jail

asteny:


rdjjke (approved):

images/common/scripts/install_container_toolkit.sh
```diff
@@ -4,7 +4,7 @@ set -e # Exit immediately if any command returns a non-zero error code
 
 # 1.17.4 latest working version
```
rdjjke: Let's drop this comment as outdated

PR #1356, opened 2025-07-31
remove size  from controller spec

Uburro:
#1330


helm/slurm-cluster/values.yaml
```diff
@@ -301,7 +301,13 @@ slurmNodes:
...
     volumes:
       spool:
-        volumeSourceName: "controller-spool"
+        volumeClaimTemplateSpec:
+          accessModes:
+            - ReadWriteOnce
+          resources:
+            requests:
+              storage: 50Gi
+          storageClassName: "compute-csi-network-ssd-ext4"
```
dstaroff, line 310: Doesn't look like general approach. It should be parameterised with a variable, at least. I don't like it to be Nebius-specific

dstaroff (changes requested):

helm/slurm-cluster/values.yaml
Uburro: https://github.com/nebius/soperator/blob/main/helm/slurm-cluster/values.yaml#L54
https://github.com/nebius/soperator/blob/main/helm/slurm-cluster/values.yaml#L344
We already have a lot of such places.
What's the difference?

api/v1/slurmcluster_types.go
```diff
@@ -740,7 +740,16 @@ type AccountingSlurmConf struct {
 
 // SlurmNodeController defines the configuration for the Slurm controller node
 type SlurmNodeController struct {
-	SlurmNode `json:",inline"`
```
theyoprst, line 743: Some existing installations have `size:` in their SlurmCluster custom resources. Does that mean that their CR will be invalid now? Shouldn't we add an optional Size / return SlurmNode back for backward compatibility?

Uburro: There won’t be any issues — the CRD will be updated and this field will disappear from the CR.

theyoprst (approved):

helm/slurm-cluster/values.yaml
dstaroff: Ah, it's `values.yaml`. I thought it's template. Ok then

PR #1377, opened 2025-08-05
slurm reservation for suspicious nodes 2

MeenaAlfons:
## Feature:
When health check fails, create a slurm reservation for the suspicious node to prevent workload from being scheduled on it and to run extensive health check on it. After the extensive health check, the node could be released from the reservation if it's healthy again or get decommissioned if it's definitely not healthy.

## Summary of the logic

When an active health check fails, a reservation is created on node:
- If a job was queued while the active check is running, it will manage to be scheduled on the node before the controller applies the reactions and create the reservation.
- The reservation will still be created while letting the job continue running without interruption. 
- After this job ends, no more jobs will be scheduled on this node because it’s reserved.

For old cluster, when an active health check fails, the slurm node is drained.
- The controller picks up this drained node, creates a reservation, and undrains it. - 
- This doesn’t interrupt any running job on the node.

Every 10 minutes, extensive check will be scheduled on reserved nodes.
- If an extensive check is still in progress on a node, the current extensive check will keep running and no new one will be scheduled  for this node.

When the extensive check ends:
- If it succeeds, the reservation is removed.
- If it fails, nothing will happen for now. The node will stay reserved. The extensive check will be rerun on it again.

Currently we provide an extensive check that always fails.

## Changes

Depends on: https://github.com/nebius/soperator/pull/1424

**Breaking changes**
- Change all slurm active checks to use `reactions.AddReservation` instead of `drainSlurmNode` and `setCondition`.
- For backward compatibility, for drained slurm nodes with reasons `[HC]` (health check), create a reservation and undrain them.

**Non-breaking changes**
- Adds `successReactions` and `failureReactions` to `ActiveCheck` specs.
  - Support `AddReservation` with `prefix` in failure reactions.
  - Support `RemoveReservation` with `prefix` in success reactions
- **[TESTED]** Add `CronJob` named `run-extensive-check-on-reservations` which runs `extensive-check` on each reservation with prefix `soperatorchecks.suspicious`
- **[TESTED]** Add `extensive-check` which is run by `cronjob/run-extensive-check-on-reservations` on each reservation.
  - Right now it doesn't result in any reactions.
  - It should have `successReactions.RemoveReservation` and `failureReactions.setHardwareIssuesSuspected`

## Notes

- Active checks are deployed from terraform in [nebius/nebius-solutions-library](https://github.com/nebius/nebius-solutions-library/tree/main/soperator) not from the helm chart in this repo. However, we make the updates in both places.
- The other PR in [nebius/nebius-solutions-library](https://github.com/nebius/nebius-solutions-library/tree/main/soperator) includes:
  - The new version of the controller
  - The new version of soperator helm chart which contains the CRDs
  - The changes to active checks.


go.mod
```diff
@@ -3,7 +3,7 @@ module nebius.ai/slurm-operator
 go 1.24.2
 
 require (
-	github.com/SlinkyProject/slurm-client v0.1.0-20241108145209-7bfdd4923f75
+	github.com/SlinkyProject/slurm-client v0.3.1-20250801102150-043094ce1e4e
```
asteny, line 6: If we want to bump it, we should test a lot of thing around this client: slurm exporter, controllers which use it

helm/soperator-activechecks/scripts/run-check-on-reservation.sh
```diff
@@ -0,0 +1,14 @@
+set -e
+
+apt update && apt install -y retry
```
asteny: May be better to have it inside k8sjob image? But I didnt find how are you using retry...

internal/slurmapi/client.go
```diff
@@ -23,8 +25,14 @@ func DefaultHTTPClient() *http.Client {
 }
 
 type client struct {
+	/**
+	 * Refactor: hide the APIs of a specific SLURM REST API version
```
asteny, line 36: Is this a TODO?

internal/slurmapi/job_test.go
```diff
@@ -272,6 +273,6 @@ func TestJob_GetIDString(t *testing.T) {
 	}
 }
 
-func ptr[T any](v T) *T {
-	return &v
-}
+// func ptr.To[T any](v T) *T {
```
asteny: ?

go.mod
MeenaAlfons: This bump is needed to be able to create reservations which is only available beginning from `v0.0.43` for slurm rest api.

The code is still using `v0.0.41` for all current operations. Only this new operation is using `v0.0.43`.

MeenaAlfons: We still need some regression testing just in case the upgraded version of the client has any issues.

helm/soperator-activechecks/scripts/run-check-on-reservation.sh
MeenaAlfons: Copy&Paste. Removed.

internal/slurmapi/client.go
MeenaAlfons: It's a recommendation. It's not a TODO for this PR. Happy to mark it TODO for later.

helm/soperator-activechecks/scripts/run-check-on-reservation.sh
```diff
@@ -0,0 +1,12 @@
+set -e
+
+echo "Submitting k8s jobs for active check $TARGET_ACTIVE_CHECK_NAME for reserved nodes with prefix $RESERVATION_PREFIX ..."
+for reservationName in $(scontrol show reservation --json | jq ".reservations.[] | select(.name | startswith(\"$RESERVATION_PREFIX\")) | .name"); do
+  jobName="$TARGET_ACTIVE_CHECK_NAME-$reservationName"
+
+  # We need to pass the reservationName to this job
+  kubectl create job --from=cronjob/$TARGET_ACTIVE_CHECK_NAME $jobName
```
ChessProfessor: That's old code, right?

helm/soperator-activechecks/scripts/run-check-on-reservation.sh
```diff
@@ -0,0 +1,12 @@
...
+  kubectl create job --from=cronjob/$TARGET_ACTIVE_CHECK_NAME $jobName --dry-run -o "json" \
+  | jq ".spec.template.spec.containers[0].env += [{ \"name\": \"RESERVATION_NAME\", value:\"$reservationName\" }]" \
```
ChessProfessor: Maybe there is a better solution to pass RESERVATION_NAME to extensive active check, need to discuss this with the team
Maybe it's also fine

helm/soperator-activechecks/templates/extensive-check.yaml
```diff
@@ -0,0 +1,30 @@
+apiVersion: slurm.nebius.ai/v1alpha1
+kind: ActiveCheck
+metadata:
+  name: "extensive-check"
+spec:
+  checkType: "slurmJob"
+  name: "extensive-check"
+  slurmClusterRefName: {{ .Values.slurmClusterRefName | quote }}
+  schedule: "0 10 * * *" # Once a day at 10:00 UTC
```
ChessProfessor: Maybe let's do once a year or something like that, if this check is suspended anyway

helm/soperator-activechecks/templates/run-check-on-reservation.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: slurm.nebius.ai/v1alpha1
+kind: ActiveCheck
+metadata:
+  name: "run-check-on-reservation"
+spec:
+  checkType: "k8sJob"
+  name: "run-check-on-reservation"
+  slurmClusterRefName: {{ .Values.slurmClusterRefName | quote }}
+  suspend: true
```
ChessProfessor: Should it be suspended? I was thinking this check should run periodically to trigger extensive checks
We also need cron schedule for it

internal/controller/soperatorchecks/activecheck_jobs_controller.go
```diff
@@ -352,3 +336,96 @@ func getK8sJobStatus(k8sJob *batchv1.Job) consts.ActiveCheckK8sJobStatus {
...
+
+	err := processRemoveReservation(ctx, successReactions.RemoveReservation, slurmJob, slurmAPIClient)
+	if err != nil {
+		return fmt.Errorf("removing reservation: %w", err)
+	}
+	return nil
+}
+
+func processAddReservation(ctx context.Context, addReservation *slurmv1alpha1.ReservationSpec, slurmJob slurmapi.Job, slurmAPIClient slurmapi.Client) error {
+	if addReservation != nil && addReservation.Prefix != "" {
```
ChessProfessor:
```suggestion
	if addReservation == nil || addReservation.Prefix == "" {
	    return nil
	}
```

internal/controller/soperatorchecks/activecheck_jobs_controller.go
```diff
@@ -352,3 +336,96 @@ func getK8sJobStatus(k8sJob *batchv1.Job) consts.ActiveCheckK8sJobStatus {
...
+			if err != nil {
+				return fmt.Errorf("post reservation: %w", err)
+			}
+		}
+	}
+	return nil
+}
+
+func processRemoveReservation(ctx context.Context, removeReservation *slurmv1alpha1.ReservationSpec, slurmJob slurmapi.Job, slurmAPIClient slurmapi.Client) error {
+	if removeReservation != nil && removeReservation.Prefix != "" {
```
ChessProfessor: The same here

internal/controller/soperatorchecks/activecheck_jobs_controller.go
```diff
@@ -352,3 +336,96 @@ func getK8sJobStatus(k8sJob *batchv1.Job) consts.ActiveCheckK8sJobStatus {
...
+			err := slurmAPIClient.StopReservation(ctx, reservationName)
+			if err != nil {
+				return fmt.Errorf("stop reservation: %w", err)
+			}
+		}
+	}
+	return nil
+}
+
+func reservationNameForNode(reservationPrefix string, nodeName string) string {
```
ChessProfessor: Let's move it to naming.go?

internal/controller/soperatorchecks/slurm_nodes_controller.go
```diff
@@ -204,49 +204,71 @@ func (c *SlurmNodesController) processHealthCheckFailed(
...
 	}
 
-	reason, message, err := parseHealthCheckReason(nodeReason.OriginalReason)
-	if err != nil {
-		return fmt.Errorf("parse health check reason: %w", err)
-	}
+	/**
+	Health checks have success and failure reactions.
+	When a health check fails, we can already create a reservation using failureReaction.addReservation
+	There is no need to reactions.drainSlurmNode then execute logic here to handle DRAINED clurm nodes with [HC] reason
```
ChessProfessor: clurm -> slurm

internal/slurmapi/client.go
```diff
@@ -6,7 +6,9 @@ import (
 	"net/http"
 
 	api "github.com/SlinkyProject/slurm-client/api/v0041"
```
ChessProfessor, line 10: Let's rename it to api0041 to make it clear?

helm/soperator-activechecks/templates/all-reduce-perf-nccl-ib.yaml
```diff
@@ -24,5 +24,5 @@ spec:
     mungeContainer:
       image: {{ .Values.images.munge | quote }}
   reactions:
-    setCondition: true
-    drainSlurmNode: true
+    addReservation:
+      prefix: {{ .Values.reservationPrefix | quote }}
```
ChessProfessor, line 28: There are more active checks coming right now (mem-bw/lat & gpu-fryer + there will be ib_write_* checks soon). Adding comment not to forget about them too

helm/soperator-activechecks/scripts/run-check-on-reservation.sh
MeenaAlfons: The reservation name is different per node. It's not constant to pass it in the ActiveCheck specs. Hence I'm passing it as an environment variable to the job. Each job will have a different reservation name.

helm/soperator-activechecks/templates/all-reduce-perf-nccl-ib.yaml
MeenaAlfons: Great! it's also good that we have the backward compatibility code in case some active checks are defined with `drainSlurmNode`.

internal/slurmapi/client.go
MeenaAlfons: Let's leave this refactor out of this PR to keep the PR focused.

helm/soperator-activechecks/templates/run-check-on-reservation.yaml
MeenaAlfons: I set the schedule for it to run every 10 mins.

helm/soperator-activechecks/templates/run-extensive-check-on-reservations.yaml
```diff
@@ -0,0 +1,125 @@
+apiVersion: batch/v1
+kind: ConJob
```
ChessProfessor: CronJob?

internal/controller/soperatorchecks/activecheck_jobs_controller.go
```diff
@@ -198,21 +204,48 @@ func (r *ActiveCheckJobReconciler) Reconcile(
 			}
 
 			// Job has already been seen in one of the previous reconciler runs
-			if activeCheck.Status.SlurmJobsStatus.LastJobEndTime != nil && !slurmJob.EndTime.After(activeCheck.Status.SlurmJobsStatus.LastJobEndTime.Time) {
+			if k8sJob.Annotations[K8sAnnotationSoperatorChecksFinalStateTime] != "" {
 				continue
 			}
 
 			if lastEndTime == nil || slurmJob.EndTime.After(lastEndTime.Time) {
 				lastEndTime = slurmJob.EndTime
```
ChessProfessor: I guess we don't need lastEndTime in the ActiveCheck status anymore, it was used only to check if the job was already reconciled, now we have K8sAnnotationSoperatorChecksFinalStateTime

internal/controller/soperatorchecks/activecheck_jobs_controller.go
```diff
@@ -198,21 +204,48 @@ func (r *ActiveCheckJobReconciler) Reconcile(
...
+					return ctrl.Result{}, fmt.Errorf("executing failure reactions: %w", err)
+				}
+			case slurmJob.IsCompletedState():
+				err = executeSuccessReactions(ctx, slurmJob, activeCheck, slurmAPIClient)
+				if err != nil {
+					return ctrl.Result{}, fmt.Errorf("executing success reactions: %w", err)
 				}
+			default:
+				// Do nothing. The job could have been cancelled or interrupted. The job will run again.
+				logger.Info(fmt.Sprintf("unhandled state. The job is probably cancelled or interrrupted and it will run again. Current state: %s ", slurmJob.State))
```
ChessProfessor: Interrupted

internal/controller/soperatorchecks/slurm_nodes_controller.go
```diff
@@ -204,49 +204,72 @@ func (c *SlurmNodesController) processHealthCheckFailed(
...
-				message,
-			),
-		); err != nil {
-			return fmt.Errorf("drain slurm nodes: %w", err)
-		}
-		return nil
-	}
+	For backward compatability, we add some logic here to handle already drained slurm nodes with [HC] reason and create a reservation for them then undrain them.
+	*/
+	_ = nodeReason
```
ChessProfessor: Why do we need that?

helm/soperator-activechecks/templates/run-extensive-check-on-reservations.yaml
```diff
@@ -36,8 +36,7 @@ spec:
         spec:
           containers:
           - name: run-extensive-check-on-reservations
-            # TODO replace with stable image
-            image: cr.eu-north1.nebius.cloud/soperator-unstable/slurm_check_job:1.21.10-noble-slurm24.11.6-2e1138d0
+            image: cr.eu-north1.nebius.cloud/soperator/slurm_check_job:1.21.10-noble-slurm25.05.2
```
ChessProfessor: We can use `{{ .Values.images.slurmJob | quote }}`?

internal/controller/soperatorchecks/activecheck_jobs_controller.go
MeenaAlfons: Yes that's the idea of my change.

Also the reason this became more relevant here is that an ActiveCheck used to have only one job. Now, for extensive check, it will have multiple jobs. So if we want to store some state on whether the job has been processed or not we should store it on the job itself not on the active check.

There is still some logic that stores some status on the ActiveCheck itself. I left this code intact event though it's not totally compatible with having multiple jobs for an ActiveCheck. With testing it, it doesn't cause any harm.

MeenaAlfons: I guess your comment was about suggesting the removal of the field `LastJobEndTime` on `struct SlurmJobsStatus`. OK I will do that.

helm/soperator-activechecks/scripts/extensive-check.sh
```diff
@@ -0,0 +1,78 @@
...
+elif [[ "${gpus_on_node}" == *"8 NVIDIA B200"* ]]; then
+  platform="8xB200"
+else
+  echo "Unsupported platform"
+  exit 0
+fi
+echo "Platform found: $platform"
+
+
+# (1) I don't get why we use srun here. We are already on a worker node!
```
ChessProfessor: It's for the eachWorkerJobArray, we discussed it in slack

helm/soperator-activechecks/scripts/extensive-check.sh
```diff
@@ -0,0 +1,78 @@
...
+# srun --cpu-bind=verbose,cores --gpus-per-node=8 bash -c "health-checker run -e soperator -p $platform -n all_reduce_without_ib --report-format json-pretty"
+
+# (2) Do we need --cpu-bind=verbose,cores for the following tests? and why?
```
ChessProfessor: TBH I'm not sure, better ask Misha when he's back, he told me to add it

helm/soperator-activechecks/scripts/extensive-check.sh
```diff
@@ -0,0 +1,78 @@
...
+#SBATCH --gpus-per-node=8
+
+apt update
+apt install --only-upgrade nc-health-checker
+apt-cache policy nc-health-checker
+
+echo "Checking for running GPU processes..."
+if [[ -n "$(nvidia-smi --query-compute-apps=pid --format=csv,noheader | grep -v '^ *$')" ]]; then
+  echo "Another GPU process is currently running. Exiting."
+  exit 0
```
ChessProfessor, line 17: Isn't it success of the extensive check? Maybe we should fail in such case, otherwise node will go available after this

helm/soperator-activechecks/scripts/extensive-check.sh
```diff
@@ -0,0 +1,78 @@
...
+gpus_on_node=$(nvidia-smi --query-gpu=name --format=csv,noheader | sort | uniq -c)
+if [[ "${gpus_on_node}" == *"8 NVIDIA H100"* ]]; then
+  platform="8xH100"
+elif [[ "${gpus_on_node}" == *"8 NVIDIA H200"* ]]; then
+  platform="8xH200"
+elif [[ "${gpus_on_node}" == *"8 NVIDIA B200"* ]]; then
+  platform="8xB200"
+else
+  echo "Unsupported platform"
+  exit 0
```
ChessProfessor, line 30: And maybe here

helm/soperator-activechecks/scripts/extensive-check.sh
```diff
@@ -0,0 +1,105 @@
...
+  srun --container-image="$CONTAINER_IMAGE" \
+  --container-mounts=$(which health-checker):/usr/local/bin/health-checker \
+  --cpu-bind=verbose \
+  bash -c "HC_GPU_FRYER_DURATION=300 health-checker run -e soperator -p $platform -n gpu_fryer --report-format json-pretty"
+}
+
+funcs_to_test=(
+  all_reduce_without_ib
+  all_reduce_with_ib
+  # mem_bw: gives an error on B200: Error: unable to bind thread to core 159 with hwid 159
```
ChessProfessor: You fixed it with --exclusive, right?

helm/soperator-activechecks/scripts/extensive-check.sh
ChessProfessor: Maybe we need some pipeline for rescheduling the extensive check in case of failures like this?

helm/soperator-activechecks/scripts/extensive-check.sh
```diff
@@ -0,0 +1,105 @@
...
+else
+  echo "Unsupported platform"
+  exit 0
+fi
+echo "Platform found: $platform"
+
+echo "Listing available health checks for platform $platform"
+health-checker list -e soperator -p $platform
+
+CONTAINER_IMAGE="cr.eu-north1.nebius.cloud#soperator/active_checks:12.9.0-ubuntu24.04-nccl_tests2.16.4-b8189f7"
```
ChessProfessor: Let's use the values from values.yaml? Like {{ .Values.activeCheckImage }}?

helm/soperator-activechecks/scripts/extensive-check.sh
```diff
@@ -1,26 +1,105 @@
 #!/bin/bash
-#SBATCH --deadline="now+6hours"
-#SBATCH --time=10:00
-#SBATCH --mem=32G
-#SBATCH --gpus-per-node=1
-#SBATCH --cpus-per-task=2
+#SBATCH --time=01:00:00
```
ChessProfessor, line 3: Maybe we can have deadline like 24 hours?

helm/soperator-activechecks/scripts/extensive-check.sh
MeenaAlfons: This helm chart is not actually being deployed anywhere. I'm just copying the code here for some consistency.

helm/soperator-activechecks/scripts/extensive-check.sh
MeenaAlfons: In this unsupported platform case, it's not clear to me whether we should:
- consider it a failure
- consider it a success
- retry later (I'm not sure what would change later)

helm/soperator-activechecks/scripts/extensive-check.sh
MeenaAlfons: When running this job with `--exclusive`, it's unlikely that another GPU process is still running. It's quite an edge case. If that happens to be the case, it's not clear to me what could be a _good_ solution. We can already introduce some retries here, but eventually when it fails, will we consider it a failure or a success?

Notice that a success makes the node available for scheduled jobs which seems to be safer than draining the node and marking it with a hardware issue.

So I'm kinda leaning towards returning exit code 0 here after some retries.

helm/soperator-activechecks/scripts/extensive-check.sh
MeenaAlfons: Failure will drain the node and mark it with hardware issue. Is this what we want in this case?
Success makes the node available for scheduled jobs.

MeenaAlfons: I think this case will happen if `soperator` is installed on a new cluster with a new set of GPUs that are not in this list of platforms. In that case, we would want to signal that something is not right without marking those nodes with hardware issues.

Active checks are designed to produce either success or failures. There is no more states in the result. Hence, I'd rather return exit code 0 here and let the cluster keep running. 🤷 

Again, this is another edge case.

MeenaAlfons: In some sense, we should only fail if we are sure something is totally wrong with this node because failure will be attached to "hardware issues".

So any other failure like:
- Not being able to run the test
- Not being able to pull an image
- Unsupported platform
should result in exit code 0.

helm/soperator-activechecks/scripts/extensive-check.sh
rdjjke: I agree that this is an edge case, and maybe we don't need to think about it much.
I'm fine with returning exit code 0 (it's better than returning exit code 1).
But one more solution is to simply kill these GPU processes.

helm/soperator-activechecks/scripts/extensive-check.sh
rdjjke: I agree, let's return exit code 0 for now.
But in general, we can think about a "do nothing" resolution (the node remains reserved, and the extensive check continues to run there indefinitely)

PR #1385, opened 2025-08-06
del readeness probe from worker

Uburro:
We don’t do anything and have no automation in place for when this probe fails. But we’re getting a lot of spam in the events. That’s why the useless probe is being removed.

### DIFF
--- a/internal/render/worker/container.go
+++ b/internal/render/worker/container.go
@@ -98,21 +98,6 @@ func renderContainerSlurmd(
 			Protocol:      corev1.ProtocolTCP,
 		}},
 		VolumeMounts: volumeMounts,
-		ReadinessProbe: &corev1.Probe{
-			ProbeHandler: corev1.ProbeHandler{
-				Exec: &corev1.ExecAction{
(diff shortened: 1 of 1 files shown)


internal/render/worker/container.go
```diff
@@ -98,21 +98,6 @@ func renderContainerSlurmd(
...
-						"show",
-						"slurmd",
-					},
-				},
-			},
-			PeriodSeconds:    1,
-			TimeoutSeconds:   common.DefaultProbeTimeoutSeconds,
-			SuccessThreshold: common.DefaultProbeSuccessThreshold,
-			FailureThreshold: common.DefaultProbeFailureThreshold,
-		},
```
itechdima, line 115: let's maybe not remove it, but move it to the startup probe, otherwise we don't have any indication, that slurmd worker pod successfully registered in the controller

Uburro: command terminated with exit code 137
When I set StartupProbe with same values. It’s not obvious what value to set, as it depends on many factors.

PR #1392, opened 2025-08-07
Undrain nodes after user problems + rewrite passive checks on Python

rdjjke:
Now, passive checks (Slurm `Prolog`, `Epilog`, and `HealthCheckProgram` scripts) also run on drained nodes.
Nodes, drained by `alloc_gpus_busy` and `boot_disk_full` checks, are automatically undrained if the user fixes the problem.

The logic for launching passive check commands is moved to a new Python script, `check_runner.py`, which accepts a JSON configuration file containing the settings for all checks.

Other changes:
- Moved utility scripts (`map_job_dcgm.sh`, `unmap_job_dcgm.sh`, `cleanup_enroot.sh`) to the same check runner.
- Nodes are drained with either `[node_problem]` or `[user_problem]` prefix instead the previous `[HC]`.
- Some checks run out of jail, where possible.
- Added the possibility for commenting nodes instead of draining.
- Long-running calls (`chroot`, `nvidia-smi`, `sinfo`) are executed only once instead of several times.
- Added recommendations for fixing `[user_problem]` issues to node drain reasons.
- Fixed rare failures of `enroot_cleanup.sh` script.
- Fixed `map_job_dcgm.sh` script so it maps only allocated GPUs.
- Added utility script `/opt/soperator-utils/fs_usage.sh` that prints usage for shared, local, or in-memory volumes.
- Enabled Slurm `RebootProgram` and `scontrol reboot` command.
- Fixed `10-system-info.sh` SSH welcome message by printing FS usage correctly.
- Added SSH welcome message to worker nodes (the same as on login nodes).
- Changed `eachWorkerJobArray: true` active checks so that they don't schedule jobs for reserved nodes.
- Increased `SlurmctldTimeout` from 30 to 180 sec.
- Added healthcheck for draining Slurm nodes that have available memory < allocated memory.


helm/slurm-cluster/slurm_scripts/check_runner.py
```diff
@@ -0,0 +1,303 @@
+import os
+import subprocess
+import sys
+import json
+import string
+import datetime
+import typing
+
+SLURMD_NODENAME = ""
```
asteny: `os.environ.get("SLURMD_NODENAME")`

helm/slurm-cluster/slurm_scripts/check_runner.py
```diff
@@ -0,0 +1,303 @@
...
+  contexts: list[str] = ["any"]
+
+class NodeInfo(typing.NamedTuple):
+  state_flags: list[str] = []
+  reason: str = ""
+  comment: str = ""
+
+def main():
+  # Environment
+  global SLURMD_NODENAME, SLURM_JOB_GPUS, CHECKS_OUTPUTS_BASE_DIR, CHECKS_CONTEXT, CHECKS_CONFIG
```
asteny: after `os.environ.get` you do not need this part

helm/slurm-cluster/slurm_scripts/check_runner.py
```diff
@@ -0,0 +1,303 @@
...
+  try:
+    SLURMD_NODENAME = os.environ["SLURMD_NODENAME"]
+    SLURM_JOB_GPUS = os.environ.get("SLURM_JOB_GPUS", "")
+    CHECKS_OUTPUTS_BASE_DIR = os.environ["CHECKS_OUTPUTS_BASE_DIR"]
+    CHECKS_CONTEXT = os.environ["CHECKS_CONTEXT"]
+    CHECKS_CONFIG = os.environ["CHECKS_CONFIG"]
+  except Exception as e:
+    print(f"Failed to get environment variable '{e.args[0]}', exiting: {e}", file=sys.stderr)
```
asteny: https://docs.python.org/3/howto/logging.html#a-simple-example
or better https://github.com/mosquito/prettylog

helm/slurm-cluster/slurm_scripts/check_runner.py
```diff
@@ -0,0 +1,303 @@
...
+  if host_checks:
+    print("Running checks on the host filesystem:")
+    print(json.dumps(host_checks, indent=2))
+    chdir_into_checks_dir()
+    for check in host_checks:
+      run_check(check, in_jail=False)
+
+  # Run checks on the jail filesystem
+  jail_checks = [c for c in applicable_checks if c.run_in_jail]
+  if jail_checks:
```
asteny, line 173: Hadouken
```
if not jail_checks:
    sys.exit(0)

?

File: helm/slurm-cluster/slurm_scripts/check_runner.py [hunk 5]
```diff
@@ -0,0 +1,303 @@
...
+    chroot_into_jail()
+    chdir_into_checks_dir()
+    for check in jail_checks:
+      run_check(check, in_jail=True)
+
+  sys.exit(0)
+
+# Filter checks for the current environment
+def filter_applicable_checks(checks: list[Check]) -> list[Check]:
+  slurm_job_cpu_only = SLURM_JOB_GPUS == "" and CHECKS_CONTEXT in ["prolog", "epilog"]
```
asteny:
```
is_slurm_job_cpu_only: bool = ...

File: helm/slurm-cluster/slurm_scripts/check_runner.py [hunk 6]
```diff
@@ -0,0 +1,303 @@
...
+  log_rel_path = string.Template(check.log).safe_substitute(
+    worker=SLURMD_NODENAME, context=CHECKS_CONTEXT, name=check.name
+  )
+  log_abs_path = os.path.join(CHECKS_OUTPUTS_BASE_DIR, log_rel_path)
+  if not in_jail:
+    log_abs_path = "/mnt/jail" + log_abs_path
+  print(f"[{timestamp}] <{rootfs}> Running check {check.name} ({check.command}), logging to {log_abs_path}")
+
+  # Create parent dirs with full permissions
+  os.makedirs(os.path.dirname(log_abs_path), mode=0o777, exist_ok=True)
```
asteny, line 244: all such things can be done by using logging

helm/slurm-cluster/slurm_scripts/check_runner.py
```diff
@@ -0,0 +1,303 @@
...
+  reason = reason_base
+  details = result.stdout.strip()
+  if check.append_details and details:
+    reason += f": {details}"
+  reason += f" [{CHECKS_CONTEXT}]"
+
+  # Get info about the Slurm node
+  node_info = get_node_info()
+  # React to the check result
+  if result.returncode == 0:
```
asteny: if not can be without Hadouken

helm/slurm-cluster/slurm_scripts/check_runner.py
```diff
@@ -0,0 +1,303 @@
...
+# Chroot into jail rootfs directory
+def chroot_into_jail():
+  try:
+    os.chroot("/mnt/jail")
+  except Exception as e:
+    print(f"Failed to chroot into /mnt/jail, exiting: {e}", file=sys.stderr)
+    sys.exit(0)
+
+# Get info about the Slurm node from sinfo
+_cached_node_info: typing.Optional[NodeInfo] = None
```
asteny: https://docs.python.org/3/library/functools.html#functools.lru_cache ?

helm/slurm-cluster/slurm_scripts/check_runner.py
rdjjke: Now, host checks and jail checks are processed in the same way. I can do what you suggest, but then their processing will look different.
Considering this, do you think it is worth it?

helm/slurm-cluster/slurm_scripts/check_runner.py
rdjjke: What do you mean by logging?

helm/slurm-cluster/slurm_scripts/check_runner.py
rdjjke: Sorry, I didn't get it. What do you suggest changing here?

helm/slurm-cluster/slurm_scripts/check_runner.py
rdjjke: prettylog is not an option for now, while we don't use virtual envs and can't rely on the python installation in jail.
Regarding the standard logging - what for? Do we really need to configure different log levels?

helm/slurm-cluster/slurm_scripts/check_runner.py
rdjjke: Got it. My opinion is the same as here - https://github.com/nebius/soperator/pull/1392/files#r2260307810

helm/slurm-cluster/slurm_scripts/check_runner.py
asteny: I suggest change `if` to `if not`

Hadouken explanation https://www.reddit.com/r/ProgrammerHumor/comments/27yykv/indent_hadouken/

helm/slurm-cluster/slurm_scripts/check_runner.py
asteny: This is a same as in Go. You do not use pritnf for logging in Go) Why?

rdjjke: I actually use printf in Go. OK, I'll give it a try

helm/slurm-cluster/slurm_scripts/check_runner.py
rdjjke: OK, so here are two versions:
Hadouken:
```
  # React to the check result
  if result.returncode == 0:
    print(f"Check {check.name}: PASS\n")

    # Undrain / uncomment the Slurm node, if it was marked with the same reason
    if check.on_ok == "undrain" and "DRAIN" in node_info().state_flags:
      if node_info().reason and node_info().reason.startswith(reason_base):
        undrain_node()
    elif check.on_ok == "uncomment":
      if node_info().comment and node_info().comment.startswith(reason_base):
        uncomment_node()
  else:
    print(f"Check {check.name}: FAIL ({details})\n")

    # Drain / comment the Slurm node
    if check.on_fail == "drain" and "DRAIN" not in node_info().state_flags:
      drain_node(reason)
    elif check.on_fail == "comment":
      comment_node(reason)

    # Exit, if the check failed
    # In prolog context, restart the job and print a message to the job output
    if CHECKS_CONTEXT == "prolog":
      print(f"print Slurm healthcheck failed on node {SLURMD_NODENAME}, trying to automatically requeue")
      sys.exit(1)
    else:
      sys.exit(0)
```
Non-hadouken:
```
  # React to the check result
  if result.returncode != 0:
    print(f"Check {check.name}: FAIL ({details})\n")

    # Drain / comment the Slurm node
    if check.on_fail == "drain" and "DRAIN" not in node_info().state_flags:
      drain_node(reason)
    elif check.on_fail == "comment":
      comment_node(reason)

      # Exit, if the check failed
      # In prolog context, restart the job and print a message to the job output
    if CHECKS_CONTEXT == "prolog":
      print(f"print Slurm healthcheck failed on node {SLURMD_NODENAME}, trying to automatically requeue")
      sys.exit(1)

    sys.exit(0)

  print(f"Check {check.name}: PASS\n")

  # Undrain / uncomment the Slurm node, if it was marked with the same reason
  if check.on_ok == "undrain" and "DRAIN" in node_info().state_flags:
    if node_info().reason and node_info().reason.startswith(reason_base):
      undrain_node()
  elif check.on_ok == "uncomment":
    if node_info().comment and node_info().comment.startswith(reason_base):
      uncomment_node()
```
For me, the hadouken version looks more clear

rdjjke: I can change it, the difference isn't big. But do you think that the non-hadouken version is more readable in this specific case?

helm/slurm-cluster/slurm_scripts/check_runner.py
rdjjke: Actually, using logging package doesn't help with it, as:
1. It doesn't create parent directories for log files
2. These mkdirs are created not for the check_runner.py's logs, but for saving check command's stderr and stdout

helm/slurm-cluster/slurm_scripts/check_runner.py
rdjjke: Made it use stdlib logging

helm/slurm-cluster/slurm_scripts/check_runner.py
rdjjke: Moved dir creation logic to logging setting up code

helm/slurm-cluster/slurm_scripts/check_runner.py
dstaroff: JFYI, a general term for anti-Hadouken is a **guard clause**

helm/slurm-cluster/slurm_scripts/check_runner.py
rdjjke: Also, the behavior will be different. Now, the final logging is done regardless there are jail_checks or not.

helm/slurm-cluster/slurm_scripts/check_runner.py
rdjjke: OK, I did is as you suggested

helm/slurm-cluster/slurm_scripts/check_runner.py
```diff
@@ -0,0 +1,500 @@
...
+import json
+import logging
+import os
+import string
+import subprocess
+import sys
+import time
+import typing
+
+# Ensure Python has sufficient version
```
asteny: I dont think that we need such check:
Ubuntu 22 - python 3.9
Ubuntu 24 - python 3.12

asteny (dismissed):

helm/slurm-cluster/slurm_scripts/check_runner.py
rdjjke: But it's currently used from jail, and the user can theoretically reinstall Python. Or they won't be able to install an older Python on Ubuntu 22?

rdjjke: OK, removed this check

asteny (approved):

PR #1423, opened 2025-08-13
Pre release fixes 1.22/0

rdjjke:
- Make SSH welcome messages work when connecting to workers
- Disable prompts for saving host SSH keys when connecting to workers
- Disable GECOS (Full name, room number, phone, etc.) prompts in `createuser` by default
- Bump health-checker version
- Fix slurm-scripts configmap
- Add new slurm scripts that cleans up /mnt/memory (not used by default)


images/jail/scripts/createuser.py
```diff
@@ -0,0 +1,250 @@
...
+import grp
+import shlex
+import shutil
+import subprocess
+import sys
+
+def ensure_root():
+  if hasattr(os, "geteuid"):
+    if os.geteuid() != 0:
+      print("This script must be run as root.", file=sys.stderr)
```
asteny, line 28: why not logging?

images/jail/scripts/createuser.py
```diff
@@ -0,0 +1,250 @@
...
+                 help="Do not generate an internal SSH keypair (ECDSA)")
+  p.add_argument("--without-external-ssh", action="store_true",
+                 help="Do not prompt for an external SSH public key")
+
+  # Collect remaining args intended for adduser
+  args, rest = p.parse_known_args()
+  return args, rest
+
+
+def build_adduser_cmd(args, rest):
```
asteny, line 114: Why you didnt use default for args?

ali-sattari (approved):

images/jail/scripts/createuser.py
rdjjke: Because this is a manual tool, not service or something

images/jail/scripts/createuser.py
rdjjke: What default do you mean?

images/jail/scripts/createuser.py
rdjjke: And the output isn't stored in files, it's just printed for the user.

asteny (approved):

PR #1459, opened 2025-08-20
feat: Add standalone mode support for Slurm Operator Exporter with automatic token rotation

mateusclira-nv:
This PR adds secret rotation strategy for Standalone Slurm Exporter, enabling it to run independently of Kubernetes while maintaining secure token-based authentication with Slurm REST APIs.

Feature:
- Configurable token refresh intervals (default to 30min)
- It uses scontrol which is Slurm native JWT token generation
- Log shows whether token has successfully rotated or not 

image


cmd/exporter/main.go
```diff
@@ -115,6 +117,8 @@ func parseFlags() Flags {
 
 	// static token (optional); can also come from SLURM_EXPORTER_TOKEN
 	flag.StringVar(&flags.staticToken, "static-token", "", "Static JWT to send in X-SLURM-USER-TOKEN (use with rest_auth/jwt)")
+	flag.StringVar(&flags.scontrolPath, "scontrol-path", "scontrol", "Path to scontrol command for standalone mode")
+	flag.StringVar(&flags.keyRotationInterval, "key-rotation-interval", "30m", "Key rotation interval for standalone mode (e.g., 30m, 1h)")
```
ali-sattari: These can move to `configs`: 
https://github.com/nebius/soperator/blob/fbbc27f1a12101d93762e7bde826f60b12758a2e/cmd/exporter/main.go#L99

And make use of fallback to env var here:
https://github.com/nebius/soperator/blob/fbbc27f1a12101d93762e7bde826f60b12758a2e/cmd/exporter/main.go#L123-L127

Checking if the flag is not passed is better than checking against the default value in line 139 below.

cmd/exporter/main.go
```diff
@@ -132,6 +136,16 @@ func parseFlags() Flags {
...
+	if flags.scontrolPath == "scontrol" {
+		if v := os.Getenv("SLURM_EXPORTER_SCONTROL_PATH"); v != "" {
+			flags.scontrolPath = v
+		}
+	}
+	if flags.keyRotationInterval == "30m" {
+		if v := os.Getenv("SLURM_EXPORTER_KEY_ROTATION_INTERVAL"); v != "" {
+			flags.keyRotationInterval = v
+		}
+	}
```
ali-sattari: Read above comment, no need for it when `configs` is used.

cmd/exporter/main.go
```diff
@@ -185,9 +199,22 @@ func main() {
...
+		// Parse and set rotation interval
+		if rotationInterval, err := time.ParseDuration(flags.keyRotationInterval); err == nil {
+			standaloneIssuer.WithRotationInterval(rotationInterval)
+		} else {
+			log.Error(err, "Failed to parse key rotation interval, using default", "interval", flags.keyRotationInterval)
+		}
+		issuer = standaloneIssuer
 	case flags.staticToken != "":
 		issuer = staticIssuer{tok: flags.staticToken}
 	default:
```
ali-sattari: This part about selecting issuer is growing in size, please move this and the interface out as a function.
(I have comments further on about usage of `standaloneIssuer` and the background loop)

internal/jwt/standalone_issuer.go
```diff
@@ -0,0 +1,167 @@
...
+	rotationInterval  time.Duration
+}
+
+// NewStandaloneTokenIssuer creates a new standalone token issuer
+func NewStandaloneTokenIssuer(clusterID types.NamespacedName, username string) *StandaloneTokenIssuer {
+	return &StandaloneTokenIssuer{
+		clusterID:        clusterID,
+		username:         username,
+		scontrolPath:     "scontrol", // default path, can be overridden
+		rotationInterval: 30 * time.Minute, // rotate every 30 minutes
```
ali-sattari: Please define `const`s for these values

cmd/exporter/main.go
```diff
@@ -185,9 +199,22 @@ func main() {
...
 		Issue(ctx context.Context) (string, error)
 	}
+	var standaloneIssuer *jwt.StandaloneTokenIssuer
+
 	switch {
 	case ctrlClient != nil:
 		issuer = jwt.NewToken(ctrlClient).For(slurmClusterID, "root").WithRegistry(jwt.NewTokenRegistry().Build())
+	case flags.standalone:
+		standaloneIssuer = jwt.NewStandaloneTokenIssuer(slurmClusterID, "root")
+		standaloneIssuer.WithScontrolPath(flags.scontrolPath)
```
ali-sattari: Why not use it since you made it possible?

```suggestion
		standaloneIssuer = jwt.NewStandaloneTokenIssuer(slurmClusterID, "root").WithScontrolPath(flags.scontrolPath)
```

internal/jwt/standalone_issuer.go
```diff
@@ -0,0 +1,167 @@
...
+	outputStr := strings.TrimSpace(string(output))
+	if len(outputStr) == 0 {
+		return "", fmt.Errorf("scontrol returned empty output")
+	}
+
+	// Parse the SLURM_JWT=value format
+	parts := strings.Split(outputStr, "=")
+	if len(parts) != 2 || parts[0] != "SLURM_JWT" {
+		return "", fmt.Errorf("unexpected scontrol output format: %s", outputStr)
+	}
```
ali-sattari: Output parsing logic can be moved out of `getSlurmToken`, to make both functions more specific and testable.

internal/jwt/standalone_issuer.go
```diff
@@ -0,0 +1,167 @@
...
+func (s *StandaloneTokenIssuer) Stop() {
+	if s.rotationTicker != nil {
+		s.rotationTicker.Stop()
+	}
+	if s.stopCh != nil {
+		close(s.stopCh)
+	}
+}
+
+// rotationLoop runs the token rotation in the background
```
ali-sattari: Why not use the existing flow?
* get tokens on demand
* cache them until their expiry

This simplifies the implementation by removing `rotationLoop`, `Start`, and `Stop` methods, along with `stopCh` and probably the mutex.

ali-sattari (changes requested):
The new `standalone_issuer.go` doesn't really belong to `internal/jwt`. It doesn't use any other parts of that package. Also the existing package is used in other components of Soperator (hence it being K8S specific). So the standalone one should be a separate package, maybe under `internal/jwt-standalone` or similar.

Other than that please:
* Check the comments on changes
* Add some unit tests for the newly introduced package and functions
* Fix the linter issues

cmd/exporter/main.go
```diff
@@ -146,6 +152,37 @@ type staticIssuer struct{ tok string }
 
 func (s staticIssuer) Issue(_ context.Context) (string, error) { return s.tok, nil }
 
+func selectTokenIssuer(flags Flags, ctrlClient client.Client, slurmClusterID types.NamespacedName, log logr.Logger) (interface {
```
ali-sattari, line 155: Since the `issuer` interface is being used in function return, moving it outside and reusing it would make this cleaner.

```suggestion
type issuer interface {
	Issue(ctx context.Context) (string, error)
}

func selectTokenIssuer(flags Flags, ctrlClient client.Client, slurmClusterID types.NamespacedName, log logr.Logger) issuer {
	var issuer issuer
	var standaloneIssuer *tokenstandalone.StandaloneTokenIssuer


```

cmd/exporter/main.go
```diff
@@ -146,6 +152,37 @@ type staticIssuer struct{ tok string }
...
+	Issue(ctx context.Context) (string, error)
+}, *tokenstandalone.StandaloneTokenIssuer) {
+	var issuer interface {
+		Issue(ctx context.Context) (string, error)
+	}
+	var standaloneIssuer *tokenstandalone.StandaloneTokenIssuer
```
ali-sattari: - Since the `issuer` interface is being used in function return, moving it outside and reusing it would make this cleaner.
- Returning `*tokenstandalone.StandaloneTokenIssuer` just to log some parameters later in `main()` is excessive, we already have all that we need there.

Something like:
```golang
type issuer interface {
	Issue(ctx context.Context) (string, error)
}

func selectTokenIssuer(flags Flags, ctrlClient client.Client, slurmClusterID types.NamespacedName, log logr.Logger) issuer {
	var issuer issuer
```

cmd/exporter/main.go
```diff
@@ -218,6 +245,10 @@ func main() {
 	ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
 	defer stop()
 
+	if standaloneIssuer != nil {
```
ali-sattari: We have `flags.standalone` for this, and other logged parameters.

ali-sattari (approved):

PR #1466, opened 2025-08-22
NFS server Helm chart

ali-sattari:


helm/nfs-server/values.yaml
```diff
@@ -0,0 +1,160 @@
...
+  tag: ""
+
+imagePullSecrets: []
+
+# NFS Server Configuration
+nfs:
+  # Required: Directory path inside container to export
+  sharedDirectory: "/exports"
+  # Allowed client IPs/networks (e.g., "10.0.0.0/8,192.168.0.0/16" or "*")
+  permitted: "*"
```
Uburro: maybe we could add rfc-1918 here?

helm/nfs-server/templates/statefulset.yaml
```diff
@@ -0,0 +1,153 @@
...
+            - containerPort: 20048
+              name: mountd
+          livenessProbe:
+            tcpSocket:
+              port: nfs
+            initialDelaySeconds: 30
+            periodSeconds: 10
+            timeoutSeconds: 5
+            failureThreshold: 3
+          readinessProbe:
```
Uburro: But why is there a readiness probe here? After all, it will only remove the pod from the endpoints, but that makes no sense in a single-instance pod.

helm/nfs-server/values.yaml
```diff
@@ -0,0 +1,160 @@
...
+
+securityContext:
+  # NFS server requires privileged containers
+  privileged: true
+  runAsUser: 0
+  runAsGroup: 0
+
+resources:
+  limits:
+    cpu: 500m
```
Uburro: let's remove limit of cpu?

Uburro (approved):

PR #1487, opened 2025-08-27
Add user_mail to slurm exporter

nevgeny:


helm/soperator-notifier/templates/_helpers.04_rule.tpl
```diff
@@ -52,5 +53,5 @@ job_state=~"{{ include "son.config.jobStatus.completed" . }}"
 
 {{/* MetricsQL selector for system jobs. */}}
 {{- define "son.rule.jobSelector.system" -}}
-user_name!~"^(nebius|soperatorchecks)$"
+user_name!~"^(nebius|soperatorchecks)$", user_mail!=""
```
dstaroff: I think that filtering by `user_mail` presence must be configurable by dedicated lever

helm/soperator-notifier/tests/02_config_test.yaml
```diff
@@ -76,39 +76,40 @@ tests:
             {{- $state := .Labels.job_state }}
             {{- $reason := .Labels.job_state_reason }}
             {{- $user := .Labels.job_user }}
+            {{- $user_mail := .Labels.job_user_mail }}
             
             {{- if eq $state "FAILED" }}
-            - Job *{{ $job }}* (ID `{{ $id }}`){{ if $user }}, submitted by *{{ $user }}*,{{ end }} has *FAILED* (reason: `{{ $reason }}`).
+            - Job *{{ $job }}* (ID `{{ $id }}`){{ if $user_mail }}, submitted by *{{ $user_mail }}*,{{ end }} has *FAILED* (reason: `{{ $reason }}`).
```
dstaroff: Here and further: Linux user should be left. It would be better to put `user_mail` to the end of the message like `Cc: <user_mail>`
However, if the Slack handle is used as the email address like `@bob`, there may be the following problems:
- Idk if Slurm has validation for email. No problem if it treats it like a raw string
- Slack may not accept `@bob` in the message as a user mention. It may need Slack user ID to be passed instead of the handle

dstaroff (changes requested):

dstaroff:
Would the lever for disabling messaging on successful jobs be in a scope of this PR?

helm/soperator-notifier/tests/02_config_test.yaml
nevgeny: Slurm doesn’t perform any validation of this string
Slack supports only @member_id mentions in such cases

helm/soperator-notifier/templates/_helpers.04_rule.tpl
```diff
@@ -52,5 +53,5 @@ job_state=~"{{ include "son.config.jobStatus.completed" . }}"
 
 {{/* MetricsQL selector for system jobs. */}}
 {{- define "son.rule.jobSelector.system" -}}
-user_name!~"^(nebius|soperatorchecks)$"
+user_name!~"^(nebius|soperatorchecks)$", {{- if .Values.slack.forceUserMail}} user_mail!=""{{ end -}}
```
dstaroff: I think it's not the best place to put this filter. Much better place would be in `ignoreFilters` [section](https://github.com/nebius/soperator/blob/main/helm/soperator-notifier/templates/04_rule.yml#L1-L4).
It just seems odd to filter by user email inside system job selector.

helm/soperator-notifier/values.yaml
```diff
@@ -8,6 +8,9 @@ slack:
   # Whether to disable sending messages on changes of system job (health checks, etc.) statuses.
   ignoreSystemJobs: true
 
+  # Set true to ignore jobs without `user_mail` label`
+  forceUserMail: true
```
dstaroff: Maybe `sendWithUserMailOnly` or something like that (`ignoreJobsWithoutUserMail`)? `force` looks very crutch to me

dstaroff (changes requested):

dstaroff:
> including the content of the notification message.

Why so? It doesn't seem to be difficult, but preserves the original user submitted a job.

helm/soperator-notifier/values.yaml
nevgeny: Let's use `requireUserMail`

helm/soperator-notifier/templates/_helpers.04_rule.tpl
nevgeny: you're right, thanks

helm/soperator-notifier/templates/04_rule.yml
```diff
@@ -1,7 +1,10 @@
 {{- $ignoreFilter := "" }}
 {{- if .Values.slack.ignoreSystemJobs }}
-{{- $ignoreFilter = printf ", %s" (include "son.rule.jobSelector.system" .) }}
-{{- end -}}
+{{- $ignoreFilter = printf ", %s" (include "son.rule.jobSelector.system" .)  }}
+{{ end -}}
+{{- if .Values.slack.requireUserMail}}
+{{- $ignoreFilter = printf "%s, %s" $ignoreFilter (include "son.rule.jobSelector.require_user_mail" .)  }}
+{{ end -}}
```
dstaroff:
```suggestion
{{- if .Values.slack.requireUserMail }}
{{- if $ignoreFilter }}
{{- $ignoreFilter = printf "%s, %s" $ignoreFilter (include "son.rule.jobSelector.require_user_mail" .)  }}
{{- else }}
{{- $ignoreFilter = (include "son.rule.jobSelector.require_user_mail" .) }}
{{- end }}
{{- end -}}
```

dstaroff (changes requested):
It should be checked that the filter is not empty before updating it with leading comma.
The rest LGTM

helm/soperator-notifier/templates/04_rule.yml
dstaroff: It would be great if there is a unit-test for this case

dstaroff (approved):

PR #1557, opened 2025-09-10
Fill out status for Active Checks

ChessProfessor:


internal/controller/soperatorchecks/activecheck_jobs_controller.go
```diff
@@ -252,21 +262,24 @@ func (r *ActiveCheckJobReconciler) Reconcile(
...
-			state = consts.ActiveCheckSlurmJobStatusComplete
-		case len(failReasons) == totalJobs:
+		case len(failJobs) != 0:
 			state = consts.ActiveCheckSlurmJobStatusFailed
-		default:
+		case len(degradeJobs) != 0:
 			state = consts.ActiveCheckSlurmJobStatusDegraded
+		default:
+			state = consts.ActiveCheckSlurmJobStatusComplete
 		}
```
ChessProfessor: So the status will be "IN_PROGRESS" until all the jobs for the active check run are finished. After that:
At least one job with status "FAILED" -> "FAILED"
At least one job with non "COMPLETE" status -> "DEGRADED"
All the jobs with "COMPLETE" status -> "COMPLETE"

WDYT?

api/v1alpha1/activecheck_types.go
```diff
@@ -198,6 +198,12 @@ type ActiveCheckSlurmJobsStatus struct {
 	// +kubebuilder:validation:Optional
 	LastJobFailReasons []string `json:"lastJobFailReasons"`
 	// +kubebuilder:validation:Optional
+	LastJobDegradeReasons []string `json:"lastJobDegradeReasons"`
```
ChessProfessor: Would be nice to rename LastJob -> LastRun everywhere. It makes more sense with current implementation.
Also maybe it would make sense to unite fail reason and failed job id into one struct. 
wdyt?

itechdima: renaming looks OK, will be unified
but we'll need to adjust waiter script also https://github.com/nebius/soperator/blob/main/helm/soperator-activechecks/templates/wait-for-checks-job.yaml

itechdima (dismissed):

itechdima (approved):

PR #1582, opened 2025-09-16
Fix passive check filtering

rdjjke:
- Fix passive check filtering (check for GPU jobs didn't run)
- Adjust cron schedules for active checks so that many checks don't start at once
- Change Slurm configuration to allow more jobs to be in the queue at the same time


helm/slurm-cluster/slurm_scripts/check_runner.py
```diff
@@ -190,61 +190,86 @@ def main():
 
 # Filter checks for the current environment
 def filter_applicable_checks(checks: typing.List[Check]) -> typing.List[Check]:
```
asteny: checks: list[Check]

helm/slurm-cluster/slurm_scripts/check_runner.py
```diff
@@ -190,61 +190,86 @@ def main():
...
+  checks = filter_by_skip_for_cpu_jobs(checks)
+  # Filter by platform (needs platform tags)
+  checks = filter_by_platform(checks)
+  # Filter by skip_for_partial_gpu_jobs (needs platform tags)
+  checks = filter_by_skip_for_partial_gpu_jobs(checks)
+  # Filter by node state
+  checks = filter_by_node_state(checks)
+  return checks
+
+def filter_by_context(checks: typing.List[Check]) -> typing.List[Check]:
```
asteny: Here we have 2 iterators. May be we can have one? Thats also related to all filter_by functions

helm/slurm-cluster/slurm_scripts/check_runner.py
```diff
@@ -190,61 +190,86 @@ def main():
...
-    ]
+def filter_by_skip_for_partial_gpu_jobs(checks: typing.List[Check]) -> typing.List[Check]:
+  # Skip if all checks don't care
+  if all(not check.skip_for_partial_gpu_jobs for check in checks):
+    return checks
+  # Skip for non-job-related runs
+  if not job_related_run():
+    return checks
+  job_alloc_gpus = get_job_alloc_gpus()
+  node_platform_tags = get_platform_tags()
```
asteny, line 250: By the way get_platform_tags() has a bug for GB200. It will be B200:
```
elif "B200" in first

File: helm/soperator-activechecks/scripts/all-reduce-perf-nccl.sh [hunk 4]
```diff
@@ -1,15 +1,9 @@
...
+#SBATCH --deadline="now+4hours"
+#SBATCH --time=20:00
 #SBATCH --gpus-per-node=8
-#SBATCH --cpus-per-task=16
-
-echo "Checking for running GPU processes..."
-if [[ -n "$(nvidia-smi --query-compute-apps=pid --format=csv,noheader | grep -v '^ *$')" ]]; then
-  echo "Another GPU process is currently running. Exiting."
-  exit 0
-fi
```
ChessProfessor: Why did you remove it?

helm/slurm-cluster/slurm_scripts/check_runner.py
```diff
@@ -442,6 +467,14 @@ def get_job_info() -> JobInfo:
...
     return JobInfo()
 
+def job_related_run() -> bool:
+  return CHECKS_CONTEXT in ("prolog", "epilog")
+
+def get_job_alloc_gpus() -> int:
+  if SLURM_JOB_GPUS == "":
+    return 0
+  return len(SLURM_JOB_GPUS.split(","))
+
```
asteny, line 476: 500 strings of Python code and 0 tests(

ali-sattari (dismissed):

helm/soperator-activechecks/scripts/prepull-container-image.sh
```diff
@@ -1,5 +1,8 @@
 #!/bin/bash
-#SBATCH --deadline="now+6hours"
-#SBATCH --time=15:00
+#SBATCH --deadline="now+8hours"
+#SBATCH --time=30:00
+#SBATCH --exclusive
+#SBATCH --mem=0
+#SBATCH --partition="hidden"
```
ChessProfessor: I'd say better to set partition in the image, for every check at the same time
Why here?

helm/soperator-activechecks/scripts/cuda-samples.sh
```diff
@@ -1,13 +1,9 @@
...
+#SBATCH --time=10:00
 #SBATCH --gpus-per-node=8
-
-echo "Checking for running GPU processes..."
-if [[ -n "$(nvidia-smi --query-compute-apps=pid --format=csv,noheader | grep -v '^ *$')" ]]; then
-  echo "Another GPU process is currently running. Exiting."
-  exit 0
-fi
+#SBATCH --exclusive
+#SBATCH --mem=0
```
ChessProfessor, line 6: Have you tested all these SBATCH changes? In the release branch active check statuses are still not good, we have to check output of every check

helm/slurm-cluster/slurm_scripts/check_runner.py
rdjjke: This syntax is supported in Python 3.9.
Are you sure we can limit it to this version? The current limit is Python 3.6

helm/slurm-cluster/slurm_scripts/check_runner.py
rdjjke: We can, but only in this `filter_by_context` function. We can't in others, because additional actions are performed after the first iterator and before the second, which are either long-running or add load to Slurm controller.
The situation when no checks pass the filter is quite common, so I added these quick exit paths.

For example, if we don't have the first iterator in `filter_by_platform` function, simple jobs like `srun hostname` would have a lag ~10 seconds for prolog to complete.

So, as it's needed in all functions except for `filter_by_context`, I'd leave it in this function as well just to have it uniform.

helm/slurm-cluster/slurm_scripts/check_runner.py
rdjjke: Good catch.
But, BTW, do GB200 have "GB200" in `nvidia-smi` output or it's "B200"? (In either case it won't work correctly now, I understand)

helm/slurm-cluster/slurm_scripts/check_runner.py
rdjjke: All these script files are added to actual clusters. We need to make up something to write tests for scripts like that and integrate them to out CI

rdjjke: I don't mind writing tests, but I didn't plan to prepare the infrastructure for the tests in this PR

helm/soperator-activechecks/scripts/all-reduce-perf-nccl.sh
rdjjke: It's already done in prolog. If there are running processes out of Slurm control, the job will be cancelled

helm/soperator-activechecks/scripts/prepull-container-image.sh
rdjjke: In what image? You mean the entrypoint script?
The problem is that we set partition to "hidden" only for "eachWorkerJob" and "eachWorkerJobArray" active checks. But there's no such limitation for normal job checks, and I though it's intentional.
But I know that this normal-job active check is "system" so moved it to the hidden partition as well.

helm/soperator-activechecks/scripts/cuda-samples.sh
rdjjke: Yes.
In general, what I made here:
- Made all jobs exclusive because such jobs provide less load on the Slurm controller
- Made exclusive checks use all RAM (so that checks don't depend on cluster configuration for the default memory allocation).
- Changed time according to check running time (longer checks have longer time limit, faster checks have shorter). And increased time limits in general, because many checks can run longer in the bad scenario. If they won't fail sooner than the timeout, we'd see the ActiveCheck in "Complete" status instead of "Failed"
- Also changed deadlines according to the new CronJob schedules so that different launches of the same check do not overlap

helm/soperator-activechecks/scripts/prepull-container-image.sh
ChessProfessor: I don't know if it's intentional, I'd say we can set hidden partition for normal job checks too. 
For eachWorkerJobArray we have separate background partition with lower priority (otherwise it has problems with PLANNED state).

P.S. I actually just noticed that in the main branch we set background partition for regular slurm checks (in 1.22 we don't set anything). I don't think that's what we want...
Maybe let's set it to hidden in the image entrypoint?

helm/soperator-activechecks/templates/mem-perf.yaml
```diff
@@ -7,14 +7,13 @@ spec:
...
   dependsOn: [ "create-user-soperatorchecks", "wait-for-topology", "prepull-container-image" ]
   slurmClusterRefName: {{ .Values.slurmClusterRefName | quote }}
-  schedule: "0 16 * * *" # 1 time a day at 16:00 UTC
+  schedule: "20 10-22/12 * * *" # 2 times a day at 10:20 and 22:20 UTC
   suspend: false
   runAfterCreation: true
   slurmJobSpec:
     sbatchScript: |
 {{ tpl (.Files.Get "scripts/mem-perf.sh") . | indent 6 }}
     eachWorkerJobs: true
```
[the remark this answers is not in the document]
rdjjke: It's removed from everywhere except for only one check where it's relevant (single-node all-reduce-perf that uses IB)

helm/soperator-activechecks/scripts/prepull-container-image.sh
rdjjke: I can add this limit for normal job checks as well, if we decide it's needed. IDK to be honest.
I think we should use this partition by default for any checks, but it's not achievable in the current implementation. Default settings added to the entrypoint scripts have higher priority than sbatch comments, so it won't be possible to override them.

rdjjke: > P.S. I actually just noticed that in the main branch we set

Yes, let's make it the same in 1.22 and main

rdjjke: OK, I'll set hidden partition for normal job checks

rdjjke: Done

helm/slurm-cluster/slurm_scripts/check_runner.py
rdjjke: Done

ChessProfessor (dismissed):

helm/slurm-cluster/slurm_scripts/check_runner.py
asteny: Default Ubuntu 24 version of python is Python 3.12.3, for Ubuntu 22 default is python3.10
We are using default python here....

rdjjke: OK, will change it then

rdjjke: Done

asteny (approved):

PR #1645, opened 2025-10-09
Ensure healthy nodes check

itechdima:
## Problem
After cluster is provisioned and all checks run we do not check states of nodes.

## Solution
Add new ActiveCheck that would check nodes after all checks run.

## Testing
Dev cluster

## Release Notes
New `ensure-healthy-nodes` AcitveCheck that would run after all checks during cluster provisioning.

### DIFF
--- a/api/v1alpha1/activecheck_types.go
+++ b/api/v1alpha1/activecheck_types.go
@@ -21,6 +21,8 @@ type ActiveCheckSpec struct {
 	// DependsOn specifies dependency on another active checks that should be completed
 	// before running this one.
 	// Name of another checks in the same namespace.
+	// Note: All checks that the current one depends on must have runAfterCreation: true;
+	// otherwise, the dependency relationship won’t be respected.
 	// +kubebuilder:validation:Optional
 	DependsOn []string `json:"dependsOn"`
 
--- a/helm/soperator-activechecks/templates/ensure-healthy-nodes.yaml
+++ b/helm/soperator-activechecks/templates/ensure-healthy-nodes.yaml
@@ -1,0 +1,32 @@
+apiVersion: slurm.nebius.ai/v1alpha1
+kind: ActiveCheck
+metadata:
+  name: "ensure-healthy-nodes"
+spec:
+  checkType: "slurmJob"
+  name: "ensure-healthy-nodes"
+  dependsOn:
+  - "all-reduce-perf-nccl-with-ib"
+  - "all-reduce-perf-nccl-without-ib"
+  - "cuda-samples"
+  - "dcgmi-diag-r2"
+  - "dcgmi-diag-r3"
+  - "gpu-fryer"
+  - "ib-perf"
+  - "mem-perf"
+  slurmClusterRefName: {{ .Values.slurmClusterRefName | quote }}
+  suspend: {{ .Values.checks.ensureHealthyNodes.suspend }}
+  runAfterCreation: {{ .Values.checks.ensureHealthyNodes.runAfterCreation }}
+  slurmJobSpec:
+    sbatchScript: |
+{{ tpl (.Files.Get "scripts/ensure-healthy-nodes.sh") . | indent 6 }}
+    jobContainer:
+      image: {{ .Values.images.slurmJob | quote }}
+      env:
+{{ toYaml .Values.jobContainer.env | indent 8 }}
+      volumeMounts:
+{{ toYaml .Values.jobContainer.volumeMounts | indent 8 }}
+      volumes:
+{{ toYaml .Values.jobContainer.volumes | indent 8 }}
+    mungeContainer:
+      image: {{ .Values.images.munge | quote }}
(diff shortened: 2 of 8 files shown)


api/v1alpha1/activecheck_types.go
```diff
@@ -21,9 +21,17 @@ type ActiveCheckSpec struct {
...
+	// Note: only works for checks with RunAfterCreation: true
 	// +kubebuilder:validation:Optional
 	DependsOn []string `json:"dependsOn"`
 
+	// RequiredBy specifies dependency of another active checks, so it won't be started
+	// before the current one is completed.
+	// Name of another checks in the same namespace.
+	// Note: only works for checks with RunAfterCreation: true
+	// +kubebuilder:validation:Optional
+	RequiredBy []string `json:"requiredBy"`
```
ChessProfessor: Why do we need it? Why not use just dependsOn?

cmd/soperatorchecks/main.go
```diff
@@ -224,7 +224,17 @@ func main() {
...
 		return []string{pod.Spec.NodeName}
 	}); err != nil {
-		setupLog.Error(err, "unable to setup index field")
+		setupLog.Error(err, "unable to setup index field for Pod .spec.nodeName")
+		os.Exit(1)
+	}
+
+	// Index active checks by `requiredBy` field, so for every check we'd be able to select
+	// only required checks and not filter them in the code.
+	if err := mgr.GetFieldIndexer().IndexField(ctx, &slurmv1alpha1.ActiveCheck{}, "spec.requiredBy", func(rawObj client.Object) []string {
```
ChessProfessor: How will it work for checks added after cluster creation, for example by customer? Should the controller be restarted to take into account new checks?

api/v1alpha1/activecheck_types.go
itechdima: hm
maybe it's a bit overengineering here
agree, it adds unnecessary complexity

cmd/soperatorchecks/main.go
itechdima: not relevant anymore

but this is for controller, not for checks
so every new check created will be indexed in controller by specified field
nothing should've been restarted

helm/soperator-activechecks/templates/ensure-healthy-nodes.yaml
```diff
@@ -0,0 +1,32 @@
...
+  name: "ensure-healthy-nodes"
+  dependsOn:
+  - "all-reduce-perf-nccl-with-ib"
+  - "all-reduce-perf-nccl-without-ib"
+  - "cuda-samples"
+  - "dcgmi-diag-r2"
+  - "dcgmi-diag-r3"
+  - "gpu-fryer"
+  - "ib-perf"
+  - "mem-perf"
```
ChessProfessor, line 16: I feel like we will definitely forget to update it one day after adding some new check, but I'm not sure how to make it better...

ChessProfessor: I was thinking we can create separate acceptance-testing-check (for example) that will include everything we want to run on cluster creation/ new worker allocation or any other acceptance we want to do. It will be a mess of checks but in that case we won't forget what should run after what
But it's just an idea

itechdima: that's why I had `requiredBy` field :D 
so, during copy-pasting new check you also copy-paste reverse dependency on this `ensure-healthy-nodes` check
but it adds extra complexity for dependencies

itechdima: don't fully get an idea about acceptance testing
for cluster creation we'll run all checks
for new workers we'd better adapt already existing `extensive` check if we want to have acceptance of new nodes

ChessProfessor: Yep, I guess we have no simple fixes without much overhead, let's see how often we will forget some dependency :)

api/v1alpha1/activecheck_types.go
```diff
@@ -21,6 +21,7 @@ type ActiveCheckSpec struct {
 	// DependsOn specifies dependency on another active checks that should be completed
 	// before running this one.
 	// Name of another checks in the same namespace.
+	// Note: only works for checks with RunAfterCreation: true
```
ChessProfessor: It's not clear that the checks we depend on should have runAfterCreation: true, not the check with DependsOn itself.
Maybe 
`// Note: All checks that the current one depends on must have runAfterCreation: true; otherwise, the dependency relationship won’t be respected.`

api/v1alpha1/activecheck_types.go
```diff
@@ -21,7 +21,8 @@ type ActiveCheckSpec struct {
 	// DependsOn specifies dependency on another active checks that should be completed
 	// before running this one.
 	// Name of another checks in the same namespace.
-	// +kubebuilder:validation:Optional
+	// Note: All checks that the current one depends on must have runAfterCreation: true;
+	// otherwise, the dependency relationship won’t be respected.	// +kubebuilder:validation:Optional
```
ChessProfessor: Is it based to have kubebuilder line with another comment together?

PR #1651, opened 2025-10-13
SCHED-137: add  --instance-id and --extra both static and dynamic config

Uburro:
## Problem
Dynamic update of instance ID and extra parameters at slurmd startup, with static configuration.

## Solution
- Updated worker entrypoint

## Testing
-  manual testing feature when `SOPERATOR_NODE_SETS_ON=true`: slurmd uses static configuration from `slurm.conf`

## Release Notes
**Feature:** Updated worker entrypoint

### DIFF
--- a/internal/render/common/configmap.go
+++ b/internal/render/common/configmap.go
@@ -10,6 +10,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
+	slurmv1 "nebius.ai/slurm-operator/api/v1"
 	slurmv1alpha1 "nebius.ai/slurm-operator/api/v1alpha1"
 	"nebius.ai/slurm-operator/internal/consts"
 	"nebius.ai/slurm-operator/internal/naming"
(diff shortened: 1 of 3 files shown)


internal/render/common/configmap.go
```diff
@@ -247,15 +257,16 @@ func generateSlurmConfig(cluster *values.SlurmCluster) renderutils.ConfigFile {
...
+	case slurmv1.PartitionConfigTypeCustom:
 		for _, l := range cluster.PartitionConfiguration.RawConfig {
 			line := strings.TrimSpace(l)
 			if strings.HasPrefix(line, "PartitionName") {
 				clearLine := strings.Replace(line, "PartitionName=", "", 1)
 				res.AddProperty("PartitionName", clearLine)
 			}
 		}
-	case "structured":
+		AddNodeSetFeaturesToSlurmConfig(res, cluster)
```
theyoprst, line 268: I think we can... break backward compatibility here and don't add it here: these features (`slurm_worker_features` in terraform) are not used by anybody, so we should probably reinvent how to we set features in nodesets.

theyoprst: It's irrelevant, discussed

theyoprst (approved):

PR #1658, opened 2025-10-15
SCHED-155: automatic block topology

itechdima:
## Problem
Automatic block topology support.

## Solution
Added new logic in topology controller to generate automatic block topology from NVL domain labels on k8s nodes.

## Testing
Manually on dev cluster, with adjusting k8s labels

## Release Notes
Feature: added automatic block topology based on NVL domain labels on k8s nodes.

### DIFF
--- a/helm/soperator/templates/deployment.yaml
+++ b/helm/soperator/templates/deployment.yaml
@@ -41,6 +41,8 @@ spec:
         - name: SLURM_OPERATOR_WATCH_NAMESPACES
           value: {{ quote .Values.controllerManager.manager.env.slurmOperatorWatchNamespaces
             }}
+        - name: TOPOLOGY_LABEL_PREFIX
+          value: {{ quote .Values.controllerManager.manager.env.topologyLabelPrefix }}
         - name: KUBERNETES_CLUSTER_DOMAIN
           value: {{ quote .Values.kubernetesClusterDomain }}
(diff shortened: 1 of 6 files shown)


internal/controller/topologyconfcontroller/nodetopology_controller.go
```diff
@@ -65,6 +68,7 @@ func NewNodeTopologyReconciler(
 // metadata:
 //
 //	labels:
+//	  topologyconf.slurm.nebius.ai/tier-0: nvl0
```
asteny, line 73: Can we support different names for labels? For using them in different clouds

asteny (approved):

helm/soperator/templates/deployment.yaml
```diff
@@ -41,6 +41,8 @@ spec:
         - name: SLURM_OPERATOR_WATCH_NAMESPACES
           value: {{ quote .Values.controllerManager.manager.env.slurmOperatorWatchNamespaces
             }}
+        - name: TOPOLOGY_LABEL_PREFIX
```
Uburro, line 44: create task for that https://github.com/kelseyhightower/envconfig plz. It high a time

itechdima: Created for now here:
https://github.com/nebius/soperator/issues/1660

PR #1659, opened 2025-10-17
SCHED-162: add all_reduce_perf in docker

itechdima:
## Problem
We don't have any active check, that runs `srun docker`.

## Solution
Add `all_reduce_pref` in docker (without any threshold checks), so we can be sure that docker works.

## Testing
On dev cluster.

## Release Notes
Add ActiveCheck for docker runs via `srun`

### DIFF
--- a/helm/soperator-activechecks/scripts/all-reduce-perf-nccl-in-docker.sh
+++ b/helm/soperator-activechecks/scripts/all-reduce-perf-nccl-in-docker.sh
@@ -1,0 +1,20 @@
+#!/bin/bash
+#SBATCH --deadline="now+4hours"
+#SBATCH --time=10:00
+#SBATCH --gpus-per-node=8
+#SBATCH --exclusive
+#SBATCH --mem=0
+
+
+echo "Running all_reduce_perf_nccl_in_docker check on $(hostname)..."
+
+mkdir -p /tmp/soperatorchecks/a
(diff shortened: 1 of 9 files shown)


ChessProfessor (approved):

helm/soperator-activechecks/scripts/all-reduce-perf-nccl-in-docker.sh
```diff
@@ -0,0 +1,20 @@
...
+
+mkdir -p /tmp/soperatorchecks/a
+mkdir -p /tmp/soperatorchecks/b
+
+srun --gpus=8 docker run --rm \
+  --gpus=all --device=/dev/infiniband \
+  -v /tmp/soperatorchecks/a:/a \
+  --mount type=bind,source=/tmp/soperatorchecks/b,target=/b \
+  {{ include "activecheck.image.docker" . }} \
+  bash -c "NCCL_P2P_DISABLE=1 NCCL_SHM_DISABLE=1 NCCL_ALGO=Ring all_reduce_perf -b 512M -e 8G -f 2 -g 8"
```
asteny, line 19: We need to think how parametrize that for different GPU types. For example GBx00 has 4 GPU

itechdima: I'd do this separately, that's not the only issue with GBx00 cards

PR #1663, opened 2025-10-21
add retrigger check

itechdima:
## Problem
There's no option to retrigger all active checks with `runAfterCreation: true` during cluster update

## Solution
Add a separate check that could retrigger all other checks.

## Testing
Dev cluster.

## Release Notes
Add `retrigger-checks` k8s check that will run job for every check that has `runAfterCreation: true`

### DIFF
--- a/helm/soperator-activechecks/scripts/retrigger-checks.py
+++ b/helm/soperator-activechecks/scripts/retrigger-checks.py
@@ -1,5 +1,4 @@
 import os
-import sys
 import subprocess
 import logging
 import time
@@ -8,16 +7,12 @@ import datetime
 
 NS = os.environ["NAMESPACE"]
 
-try:
-  logging.Formatter.converter = time.gmtime
-  logging.basicConfig(
-      format='[%(asctime)s.%(msecs)03d UTC] %(levelname)s: %(message)s',
-      datefmt='%Y-%m-%d %H:%M:%S',
-      level=logging.INFO
-  )
-except Exception as e:
-  print(f"Failed to set up logging, exiting: {e}")
-  sys.exit(0)
+logging.Formatter.converter = time.gmtime
+logging.basicConfig(
+    format='[%(asctime)s.%(msecs)03d UTC] %(levelname)s: %(message)s',
+    datefmt='%Y-%m-%d %H:%M:%S',
+    level=logging.INFO
+)
 
 def run(cmd):
     p = subprocess.run(cmd, capture_output=True, text=True)


helm/soperator-activechecks/scripts/retrigger-checks.py
```diff
@@ -0,0 +1,62 @@
+import os
+import sys
+import subprocess
+import json
+import datetime
+
+NS = os.environ.get("NAMESPACE")
```
asteny: `NS = os.environ["NAMESPACE"]` will produce exception "no env var..." and you do not need `if not NS:`

helm/soperator-activechecks/scripts/retrigger-checks.py
```diff
@@ -0,0 +1,62 @@
...
+
+def fail(msg: str, code: int = 1):
```
asteny: you can use `logging` https://github.com/nebius/soperator/blob/main/helm/slurm-cluster/slurm_scripts/check_runner.py#L144

helm/soperator-activechecks/scripts/retrigger-checks.py
```diff
@@ -0,0 +1,62 @@
...
+
+def get_active_checks():
+    data = json.loads(run([
+        "kubectl", "get", "activechecks.slurm.nebius.ai",
+        "-n", NS, "-o", "json"
+    ]))
+    res = []
+    for it in data.get("items", []):
+        rac = it.get("spec", {}).get("runAfterCreation")
+        if rac is True or str(rac).lower() == "true":
```
asteny: `if rac or ...`

helm/soperator-activechecks/scripts/retrigger-checks.py
```diff
@@ -0,0 +1,62 @@
...
+    if p.returncode != 0:
+        raise RuntimeError(f"cmd failed: {' '.join(cmd)}\n{p.stderr}")
+    return p.stdout.strip()
+
+def get_active_checks():
+    data = json.loads(run([
+        "kubectl", "get", "activechecks.slurm.nebius.ai",
+        "-n", NS, "-o", "json"
+    ]))
+    res = []
```
asteny: why res? active_checks = [] ?

helm/soperator-activechecks/scripts/retrigger-checks.py
```diff
@@ -0,0 +1,62 @@
...
+def main():
+    activechecks = get_active_checks()
+    if not activechecks:
+        print("No CRs with .spec.runAfterCreation=true"); return
+
+    for name in activechecks:
+        if cronjob_exists(name):
+            try:
+                trigger(name)
+            except Exception as e:
```
asteny: You know, that here will be `Exception.RuntimeError`

images/common/scripts/install_kubectl.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
asteny, line 1: We use this code only in one place, lets put it into the docker file

images/common/scripts/install_kubectl.sh
```diff
@@ -0,0 +1,18 @@
...
+chmod 644 /etc/apt/keyrings/kubernetes-apt-keyring.gpg
+
+# Add the repository to Apt sources
+echo 'deb [signed-by=/etc/apt/keyrings/kubernetes-apt-keyring.gpg] https://pkgs.k8s.io/core:/stable:/v1.34/deb/ /' | tee /etc/apt/sources.list.d/kubernetes.list
+chmod 644 /etc/apt/sources.list.d/kubernetes.list
+
+# Install kubectl
+apt update -y
+apt install -y kubectl
+apt clean
```
asteny, line 18: add here rm -rf /var/lib/apt/lists/*

helm/soperator-activechecks/scripts/retrigger-checks.py
itechdima: won't work if `rac` is a string, it'll always pass the first condition

images/common/scripts/install_kubectl.sh
itechdima: but then we need to do the same for `install_docker.sh`, `install_docker_cli.sh`, `install_awscli.sh`, `install_munge.sh` and so on
and before that PR, `install_python.sh` was also with 1 usage

I'd say these scripts easier to have in the separate script file, because in dockerfile they would be uglier

helm/soperator-activechecks/scripts/retrigger-checks.py
itechdima: ok
will try this comment
https://github.com/nebius/soperator/pull/1663#discussion_r2451745533

PR #1665, opened 2025-10-21
Add ib-gpu-perf check

ChessProfessor:
## Problem

No ib gpu checks running on the cluster

## Solution

Adding few

## Testing

Checked on development cluster

## Release Notes

Feature: Introduced ib-gpu-perf Active Check

### DIFF
--- a/helm/soperator-activechecks/values.yaml
+++ b/helm/soperator-activechecks/values.yaml
@@ -23,7 +23,7 @@ images:
   k8sJob: "cr.eu-north1.nebius.cloud/soperator/k8s_check_job:1.22.1-noble-slurm25.05.4"
   munge: "cr.eu-north1.nebius.cloud/soperator/munge:1.22.1-noble-slurm25.05.4"
 # Pyxis syntax with '#' separator for registry
-activeCheckImage: "cr.eu-north1.nebius.cloud#soperator/active_checks:12.9.0-ubuntu24.04-nccl_tests2.16.4-60db3ac"
+activeCheckImage: "cr.eu-north1.nebius.cloud#soperator/active_checks:12.9.0-ubuntu24.04-nccl_tests2.16.4-00ce55c"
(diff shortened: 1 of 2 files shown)


helm/soperator-activechecks/templates/ensure-healthy-nodes.yaml
```diff
@@ -6,12 +6,14 @@ spec:
   checkType: "slurmJob"
   name: "ensure-healthy-nodes"
   dependsOn:
```
ChessProfessor, line 8: @itechdima maybe we can add afterInitialChecks flag, which will create CronJob for the check only after all the initial runs are finished? 
Because we already missing some of the checks

itechdima: yup
we could think about something like that, but outside of this PR :)

helm/soperator-activechecks/templates/ib-gpu-perf.yaml
```diff
@@ -0,0 +1,32 @@
...
+metadata:
+  name: "ib-gpu-perf"
+spec:
+  checkType: "slurmJob"
+  name: "ib-gpu-perf"
+  dependsOn: [ "create-user-soperatorchecks", "wait-for-topology", "prepull-container-image" ]
+  slurmClusterRefName: {{ .Values.slurmClusterRefName | quote }}
+  schedule: "55 1-21/4 * * *" # 6 times a day at 01:55, 05:55, 09:55, 13:55, 17:55 and 21:55 UTC
+  suspend: {{ .Values.checks.ibPerf.suspend }}
+  runAfterCreation: {{ .Values.checks.ibPerf.runAfterCreation }}
```
itechdima: since it's a separate check, let's not reuse `ibPerf` values to parametrise it, but rather add new `ibGpuPerf`, for better granularity

PR #1670, opened 2025-10-22
SCHED-173: Fix the jobs limit increase

rdjjke:
## Problem
We tried to increase the default limit on how many jobs can run on clusters, but that change didn't work correctly because I forgot to set new values in some places.

## Solution
Update `maxJobCount` and `minJobAge` settings in `slurm-cluster` default Helm values.

## Testing
1. Create a new cluster
2. Ensure `MaxJobCount` is `20000` and `MinJobAge` is `28800` in `scontrol show config`

## Release Notes
Increased the default limit on the maximum number of jobs: 10k/1d -> 20k/8h

### DIFF
--- a/helm/slurm-cluster/tests/README.md
+++ b/helm/slurm-cluster/tests/README.md
@@ -39,8 +39,8 @@ These tests verify the following kubebuilder default values:
 - `epilog: ""`
 - `prolog: ""`
 - `taskPluginParam: ""`
-- `maxJobCount: 10000`
-- `minJobAge: 86400`
+- `maxJobCount: 20000`
+- `minJobAge: 28800`
 - `messageTimeout: 60`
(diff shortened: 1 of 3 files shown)


ali-sattari (dismissed):

helm/slurm-cluster/tests/README.md
```diff
@@ -39,8 +39,8 @@ These tests verify the following kubebuilder default values:
 - `epilog: ""`
 - `prolog: ""`
 - `taskPluginParam: ""`
-- `maxJobCount: 10000`
```
Uburro, line 42: Rewrite Helm unit-test values (values used by Helm unit tests)

rdjjke: Done

Uburro (approved):

PR #1723, opened 2025-11-03
SCHED-336: devide building helm and docker images

Uburro:


.github/workflows/one_job.yml
```diff
@@ -280,9 +253,62 @@ jobs:
...
+          echo "unstable=$UNSTABLE" >> "$GITHUB_OUTPUT"
+          echo "Version: $VERSION"
+          echo "Unstable: $UNSTABLE"
+
+      - name: Run make sync-version-from-scratch
+        run: |
+          make kustomize helmify yq
+          make sync-version-from-scratch UNSTABLE=false
+
+      - name: Check for uncommitted changes
```
theyoprst: This check (that versions are consistent) are now run in "build helm", but they don't run in "build images", which mean that we can build and push incorrect versions.

Maybe move these two steps (`Run make sync-version-from-scratch` and `Check for uncommitted changes`) to pre-build?

Uburro: done

theyoprst (approved):

PR #1726, opened 2025-11-04
SCHED-208 SCHED-209 SCHED-210 Refactoring Active Check helm charts and related controller behavior

ChessProfessor:
## Problem

Current state of helm release is not compatible with extensive check feature we are preparing for 1.23

## Solution

Updating helm charts and related controller behavior

## Testing

Tested on dev cluster

## Release Notes

Refactoring Active Check helm charts and related controller behavior

### DIFF
--- a/api/v1alpha1/activecheck_types.go
+++ b/api/v1alpha1/activecheck_types.go
@@ -121,7 +121,7 @@ type Reactions struct {
 }
 
 type DrainSlurmNodeSpec struct {
-	SetHardwareIssueSuspected bool `json:"setHardwareIssueSuspected,omitempty"`
+	DrainReasonPrefix string `json:"drainReasonPrefix,omitempty"`
 }
 
 type ReservationSpec struct {
--- a/helm/soperator-activechecks/scripts/extensive-check.sh
+++ b/helm/soperator-activechecks/scripts/extensive-check.sh
@@ -42,6 +42,16 @@ _run_and_parse_hc() {
   fi
 }
 
+passive_checks() {
+  _run_and_parse_hc srun --cpu-bind=verbose,cores bash -c \
+    "cd /tmp && \
+    HC_DCGMI_DIAG_R1_DEBUGLOGFILE=/dev/null HC_DCGMI_DIAG_R1_DEBUGLEVEL=NONE \
+    health-checker run -e soperator -p $platform \
+    -n module,nvidia_smi,nvidia_smi_nvlink,nvidia_smi_topo,dmesg,ib_link,dcgmi_diag_r1 \
+    -f json-partial --tests-stdout-path /opt/soperator-outputs/health_checker_cmd_stdout \
+    --log-level info"
+}
+
 all_reduce_in_docker() {
   mkdir -p /tmp/soperatorchecks/a
   mkdir -p /tmp/soperatorchecks/b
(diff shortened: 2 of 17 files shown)


helm/soperator-activechecks/scripts/extensive-check.sh
```diff
@@ -22,56 +22,51 @@ echo "Platform found: $platform"
...
+    bash -c "health-checker run -e soperator -p $platform -n ^ib_write_bw_gpu.*$,^ib_send_lat_gpu.*$,^ib_read_lat_gpu.*$ -f json-partial --tests-stdout-path /opt/soperator-outputs/health_checker_cmd_stdout"
+}
+
+mem_perf() {
+  srun --container-image={{ include "activecheck.image.pyxis" . }} \
+    --container-mounts=$(which health-checker):/usr/local/bin/health-checker --cpu-bind=verbose,cores \
+    bash -c "health-checker run -e soperator -p $platform -n mem_bw,mem_lat -f json-partial --tests-stdout-path /opt/soperator-outputs/health_checker_cmd_stdout"
 }
 
+# TODO: add all_reduce_in_docker and passive checks
```
ChessProfessor: @rdjjke May I ask you to verify what else we want to add to extensive check?

rdjjke: We need to run all passive and active checks (except for dcgmi diag r3) that we usually run.
In this context, it's better to overrun that to underrun

helm/soperator-activechecks/templates/extensive-check.yaml
```diff
@@ -18,17 +18,15 @@ spec:
   successReactions:
     removeReservation:
       prefix: {{ .Values.reservationPrefix | quote }}
-  # No failure reactions for now until we implement a meaningful extensive check
-  # failureReactions:
-  #   setHardwareIssuesSuspected: true
+  failureReactions:
+    drainSlurmNode:
```
ChessProfessor: We also have to remove reservation here

api/v1alpha1/activecheck_types.go
```diff
@@ -109,26 +105,25 @@ type ActiveCheckSpec struct {
...
-
-	// DrainSlurmNode enabling slurm node commenting if check failed
+	DrainSlurmNode *DrainSlurmNodeSpec `json:"drainSlurmNode,omitempty"`
+	// CommentSlurmNode enabling slurm node commenting
 	// +kubebuilder:validation:Optional
 	CommentSlurmNode bool `json:"commentSlurmNode,omitempty"`
+
 	// AddReservation adds a slurm reservation with name "<prefix>-<nodeName>"
 	// +kubebuilder:validation:Optional
 	AddReservation *ReservationSpec `json:"addReservation,omitempty"`
```
ChessProfessor, line 117: We actually don't need AddReservation Reaction since we don't use it (reservation is added in the Grishas' controller). I decided to leave it just in case of design changes in future. But we can get rid of it too...

helm/soperator-activechecks/scripts/extensive-check.sh
```diff
@@ -22,73 +22,105 @@ echo "Platform found: $platform"
...
+    bash -c "health-checker run -e soperator -p $platform -n ^ib_write_bw_gpu.*$,^ib_send_lat_gpu.*$,^ib_read_lat_gpu.*$ -f json-partial --tests-stdout-path /opt/soperator-outputs/health_checker_cmd_stdout"
+}
+
+mem_perf() {
+  _run_and_parse_hc srun --container-image={{ include "activecheck.image.pyxis" . }} \
+    --container-mounts=$(which health-checker):/usr/local/bin/health-checker --cpu-bind=verbose,cores \
+    bash -c "health-checker run -e soperator -p $platform -n mem_bw,mem_lat -f json-partial --tests-stdout-path /opt/soperator-outputs/health_checker_cmd_stdout"
+}
+
+passive_checks() {
```
rdjjke: Can we perform passive_checks before any active checks?

helm/soperator-activechecks/templates/all-reduce-perf-nccl-in-docker.yaml
```diff
@@ -23,6 +23,6 @@ spec:
 {{ toYaml .Values.jobContainer.volumes | indent 8 }}
     mungeContainer:
       image: {{ .Values.images.munge | quote }}
-  reactions:
-    addReservation:
-      prefix: {{ .Values.reservationPrefix | quote }}
+  failureReactions:
+    drainSlurmNode:
+      setHardwareIssueSuspected: {{ .Values.checks.allReducePerfNCCLInDocker.setHardwareIssueSuspected }}
```
rdjjke: I'd suggest to make drain reason prefix configurable instead of `setHardwareIssueSuspected`. This will be more transparent, and it will allow us to create active checks with `[user_problem]` prefixes, for example

rdjjke (approved):

PR #1744, opened 2025-11-05
Fix: Check if driver installed in /run/nvidia/driver

andriishestakov:
## Problem
In case when gpu-operator installed with driver.enabled=false
it is supposed that nvidia-smi is installed in image and /run/nvidia/driver dir is empty.

## Solution
Check driver installation mode and use chroot only of required.

### DIFF
--- a/helm/slurm-cluster/slurm_scripts/alloc_gpus_busy.drain.sh
+++ b/helm/slurm-cluster/slurm_scripts/alloc_gpus_busy.drain.sh
@@ -10,8 +10,15 @@ if [[ -z "${SLURM_JOB_GPUS:-}" ]]; then
     exit 0
 fi
 
+# Check if chroot to driver dir required
+chroot_cmd=""
+driver_dir="/run/nvidia/driver"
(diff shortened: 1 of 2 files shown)


helm/slurm-cluster/slurm_scripts/alloc_gpus_busy.drain.sh
```diff
@@ -10,8 +10,15 @@ if [[ -z "${SLURM_JOB_GPUS:-}" ]]; then
     exit 0
 fi
 
+# Check if chroot to driver dir required
+chroot_cmd=""
+driver_dir="/run/nvidia/driver"
+if [[ "$(ls -A $driver_dir)" ]]; then
```
rdjjke, line 16: That's not critical, but there's a simpler way to check if a directory exists in bash:
```
if [ -d "$driver_dir" ]; then
```

rdjjke: Ah, sorry, you check if the directory has any files

PR #1750, opened 2025-11-07
SCHED-472 SCHED-473 Core NodeSet functionality

dstaroff:
## Problem
NodeSet feature is not functional in its purpose due to the incompleteness of the logic and configuration.

## Solution
This PR introduces a bunch of fixes and new functionalities to make NodeSet workers being able to connect to the cluster properly, as well as improvements for reconciliation of NodeSet and SlurmCluster resources.

The changes introduced:
- Added validation of created resources and some conditions are set
- Added dependencies for the STS and made SlurmCluster to reconcile on changes in NodeSets
- Added NodeSets to watch on its resources and SlurmCluster to trigger its reconciliation
- Added `Features` logic
- Reworked `nodesets` Helm chart made fully compatible with the CRD
- Decreased amount of Kube API requests during reconciliation
- Fixed `slurm.conf` rendering
- Fixed dependant resources owner to NodeSet instead of SlurmCluster
- Fix for wrongly applied default AppArmor profile
- Fixed STS matching labels
- Fix for empty SSHD ConfigMap
- Fixed docstrings and default values of CRD fields

## Testing
This PR is tested in parallel with https://github.com/nebius/nebius-solutions-library/pull/625 on simple configuration:
- 2 NodeSets
  - 2 workers each
    - 1xH100 GPU


helm/nodesets/templates/_helpers.tpl
```diff
@@ -51,3 +51,18 @@ Create the name of the service account to use
...
+{{/* Construct container gpu resource from GPU spec ([0]) and GPU resource spec ([1]) */}}
+{{- define "nodesets.resource.gpuFrom" -}}
+  {{- $gpuSpec := (index . 0) | default dict -}}
+  {{- $gpuEnabled := get $gpuSpec "enabled" | default false -}}
+  {{- if not $gpuEnabled -}}
+    {{- "" -}}
+  {{- else -}}
+    {{- $gpuVendorNvidia := hasKey $gpuSpec "nvidia" -}}
+    {{- $gpuVendor := $gpuVendorNvidia | ternary "nvidia.com" "" -}}
+    {{- $gpuResFQID := $gpuVendor | empty | ternary "gpu" (printf "%s/gpu" $gpuVendor)}}
```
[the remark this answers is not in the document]
dstaroff: > The issue is that `$gpuVendor` could be an empty string when `$gpuVendorNvidia` is false, leading to the formatted string being `/gpu` instead of `gpu`.

```
"" | empty | ternary "a" "b" -> true | ternary "a" "b" -> "a"
```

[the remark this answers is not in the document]
dstaroff: No vendor provided will return just `gpu`

Makefile
```diff
@@ -285,6 +285,10 @@ sync-version: yq ## Sync versions from file
 	@$(YQ) -i ".images.munge.tag = \"$(IMAGE_VERSION)\"" "helm/nodesets/values.yaml"
 	@$(YQ) -i ".images.slurmd.repository = \"$(IMAGE_REPO)/worker_slurmd\"" "helm/nodesets/values.yaml"
 	@$(YQ) -i ".images.slurmd.tag = \"$(IMAGE_VERSION)\"" "helm/nodesets/values.yaml"
+	@$(YQ) -i ".nodesets[0].slurmd.image.repository = \"$(IMAGE_REPO)/worker_slurmd\"" "helm/nodesets/values.yaml"
+	@$(YQ) -i ".nodesets[0].slurmd.image.tag = \"$(IMAGE_VERSION)+custom\"" "helm/nodesets/values.yaml"
+	@$(YQ) -i ".nodesets[0].munge.image.repository = \"$(IMAGE_REPO)/munge\"" "helm/nodesets/values.yaml"
+	@$(YQ) -i ".nodesets[0].munge.image.tag = \"$(IMAGE_VERSION)+custom\"" "helm/nodesets/values.yaml"
```
theyoprst, line 291: What's `+custom`? Do we really need it?

helm/nodesets/templates/_helpers.tpl
```diff
@@ -51,3 +51,18 @@ Create the name of the service account to use
...
+    {{- $resources := required ".Values.nodesets[*].slurmd.resources.gpu is required as .Values.nodesets[*].gpu.enabled is true" (index . 1)}}
+{{ $gpuResFQID }}: {{ get $resources "gpu" }}
```
theyoprst, line 66: Why no indentation? Adding 4 spaces and an extra `-` work to me, it also removes extra empty line (which look nice), am I missing something?

```
    {{- $gpuResFQID }}: {{ get $resources "gpu" }}
```

helm/nodesets/templates/priority-class.yaml
```diff
@@ -4,21 +4,26 @@
 apiVersion: scheduling.k8s.io/v1
 kind: PriorityClass
 metadata:
-  name: {{ .name }}
-  {{- with .annotations }}
+  name: {{ required ".Values.priorityClasses[*].name is required." .name | trunc 63 | trimSuffix "-" }}
```
theyoprst, line 7: Why modify priority class? I mean trunc and trim. What are you guarding for?
As an option, we could add a validation, like this (created by LLM):

```
{{- if .Values.priorityClasses }}
{{- range .Values.priorityClasses }}
{{- $name := required ".Values.priorityClasses[*].name is required." .name }}
{{- if gt (len $name) 63 }}
{{- fail (printf ".Values.priorityClasses[*].name '%s' exceeds 63 characters (got %d)" $name (len $name)) }}
{{- end }}
---
```

internal/render/worker/statefulset.go
```diff
@@ -292,8 +293,16 @@ func renderNodeSetAnnotations(nodeSet *values.SlurmNodeSet) map[string]string {
 	mungeAppArmorProfile := nodeSet.ContainerMunge.AppArmorProfile
 	workerAppArmorProfile := nodeSet.ContainerSlurmd.AppArmorProfile
 
-	if nodeSet.AppArmorProfileDefault {
-		workerAppArmorProfile = fmt.Sprintf("%s/%s", "localhost", naming.BuildAppArmorProfileName(nodeSet.ParentalCluster.Name, nodeSet.ParentalCluster.Namespace))
+	// TODO (@dstaroff) make use of defaulting webhook for this
```
theyoprst, line 296: Didn't forget about it?

internal/controller/nodesetcontroller/indexer.go
```diff
@@ -48,6 +49,36 @@ func (r *NodeSetReconciler) setupConfigMapIndexer(mgr ctrl.Manager) error {
...
+	}
+
+	attachedNodeSets := &slurmv1alpha1.NodeSetList{}
+	var requests []reconcile.Request
+
+	listOpts := []client.ListOption{
+		client.InNamespace(slurmCluster.Namespace),
+	}
+	if err := r.Client.List(ctx, attachedNodeSets, listOpts...); err != nil {
+		return requests
```
theyoprst, line 68: Log the `err`? (strange that linters didn't catch it)

Uburro (approved):

Makefile
dstaroff: It's just an example of a custom container that users can use on per-nodeset basis. `nodesets` Helm chart won't work as is, the values are to guide and to make intellisense see these values in templates

helm/nodesets/templates/_helpers.tpl
dstaroff: It's there to distinguish the utilitary variables from the value this definition returns. All of these lines could be either indented or started from the line start. It doesn't matter actually, we generate a lot of empty lines because of the templates here and in TF, so I wouldn't worry about it

helm/nodesets/templates/priority-class.yaml
dstaroff: Helm uses this truncation approach for names when you create a new chart, so I'm kinda stuck to it. Of course, it's a bad practice from Helm and failing would be a better option, but maybe next time if we want to replace this behaviour

internal/render/worker/statefulset.go
dstaroff: Nope. I would use defaulting webhook for different kinds of values including this one. But it's not that critical and would be more appropriate to include it in a bunch of similar changes. So I left it for tomorrow

internal/controller/nodesetcontroller/indexer.go
dstaroff: This function returns a list of reconciliation requests, we don't propagate errors from indexers. It's not our invention BTW, it's a `controller-runtime`'s [logic](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/handler/enqueue_mapped.go#L30-L52).

PR #1757, opened 2025-11-11
SCHED-248 Set-unhealthy on extensive check failure with compute instance id and check run id

ChessProfessor:
## Problem

We want to set HardwareIssueSuspected condition on extensive check failure for node replacement. It should include compute instance id and check run id.

## Solution

Parse and set these fields in the extensive-check to node comment, then use it in controller to set HardwareIssueSuspected condition

## Testing

To test these changes I did:

Good Path:

1. Adjust some Active Check to fail on one of the workers (I used gpu-fryer)
2. Trigger it, wait until it finishes and verify that node is drained with [node_problem] reason
3. Wait until controller undrain it and set reservation on a node
4. Trigger run-extensive-check-on-reservations
5. Wait till it finishes, it should remove reservation and return node to the pull

Bad Path:

1-3: the same
4. Adjust extensive-check Active Check to fail after some test (I used passive checks test). Just in the end of the test function I set custom LAST_RUN_ID and return 1.
5. Trigger run-extensive-check-on-reservations CronJob
6. Wait until it finishes, it should set a json comment with your custom LAST_RUN_ID and compute instance id to a node and drain it with [hardware_problem] reason
7. Some time after that node should be autoreplaced and undrained

### DIFF
--- a/.github/workflows/one_job.yml
+++ b/.github/workflows/one_job.yml
@@ -190,7 +190,13 @@ jobs:
           IMAGE_VERSION="$(make get-image-version UNSTABLE=${UNSTABLE})"
           VERSION=$(make get-version UNSTABLE=${UNSTABLE})
           OPERATOR_IMAGE_TAG=$(make get-operator-tag-version UNSTABLE=${UNSTABLE})
-          DOCKER_BUILD_ARGS="--cache-from=type=local,src=/mnt/shared-fs/docker/ --cache-to=type=local,dest=/mnt/shared-fs/docker/,mode=max"
+          DOCKER_BUILD_ARGS="\
+            --cache-from=type=local,src=/mnt/shared-fs/docker/soperator-build-0 \
+            --cache-from=type=local,src=/mnt/shared-fs/docker/soperator-build-1 \
+            --cache-from=type=local,src=/mnt/shared-fs/docker/soperator-build-2 \
+            --cache-from=type=local,src=/mnt/shared-fs/docker/soperator-build-3 \
+            --cache-to=type=local,dest=/mnt/shared-fs/docker/${{ runner.name }},mode=max \
+          "
           NFS_VERSION=$(make get-nfs-version UNSTABLE=${UNSTABLE})
 
           make sync-version UNSTABLE=${UNSTABLE}
@@ -247,7 +253,7 @@ jobs:
           rm -f images/jail_rootfs*.tar
 
           echo "Building tarball for jail"
-          make docker-build-jail UNSTABLE="${UNSTABLE}" IMAGE_VERSION=${IMAGE_VERSION}
+          make docker-build-jail DOCKER_BUILD_ARGS="${DOCKER_BUILD_ARGS}" UNSTABLE="${UNSTABLE}" IMAGE_VERSION=${IMAGE_VERSION}
 
           echo "Building and pushing image of the populate_jail ${IMAGE_VERSION}"
           make docker-build-and-push DOCKER_BUILD_ARGS="${DOCKER_BUILD_ARGS}" UNSTABLE="${UNSTABLE}" IMAGE_NAME=populate_jail DOCKERFILE=populate_jail/populate_jail.dockerfile IMAGE_VERSION=${IMAGE_VERSION}
--- a/Makefile
+++ b/Makefile
@@ -359,6 +359,7 @@ docker-build-go-base: ## Build go-base multiarch manifest locally
 		--target go-base \
 		-t go-base \
 		-f images/common/go-base.dockerfile \
+		--progress=plain \
 		$(DOCKER_BUILD_ARGS) \
 		.
 
@@ -380,6 +381,7 @@ endif
 		-t "$(IMAGE_REPO)/${IMAGE_NAME}:${IMAGE_VERSION}" \
 		-f images/${DOCKERFILE} \
 		--build-arg SLURM_VERSION="${SLURM_VERSION}" \
+		--progress=plain \
 		--push \
 		$(DOCKER_BUILD_ARGS) \
 		.
@@ -404,6 +406,8 @@ endif
 		-f images/jail/jail.dockerfile \
 		--build-arg SLURM_VERSION="${SLURM_VERSION}" \
 		--output type=tar,dest=images/jail_rootfs_amd64.tar \
+		--progress=plain \
+		$(DOCKER_BUILD_ARGS) \
 		.
 
 	# Build arm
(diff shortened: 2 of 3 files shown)


helm/soperator-activechecks/scripts/extensive-check.sh
```diff
@@ -111,13 +142,19 @@ funcs_to_test=(
...
 )
+
 for test in "${funcs_to_test[@]}"
 do
   echo "Running $test on $(hostname)..."
   $test
   TEST_EXIT_CODE=$?
 
   if [[ $TEST_EXIT_CODE -ne 0 ]]; then
+    COMMENT="Run ID: ${LAST_RUN_ID}, FirstFailedCheck: ${LAST_FAIL_TEST}, Error: ${LAST_FAIL_ERROR}"
```
theyoprst: What do you think of formatting it as a JSON? We can build further automation on that in the future

theyoprst: You should be careful with strings proper escaping though, it could be achieved like this:

```bash
$ export LAST_FAIL_ERROR='Hello, "World"' && echo $(jq -nc \
  --arg runId "$LAST_RUN_ID" \
  --arg firstFailedCheck "$LAST_FAIL_TEST" \
  --arg error "$LAST_FAIL_ERROR" \
  '{runId: $runId, firstFailedCheck: $firstFailedCheck, error: $error}')
{"runId":"","firstFailedCheck":"","error":"Hello, \"World\""}
```

helm/soperator-activechecks/scripts/extensive-check.sh
```diff
@@ -111,13 +115,22 @@ funcs_to_test=(
...
 do
   echo "Running $test on $(hostname)..."
   $test
   TEST_EXIT_CODE=$?
 
   if [[ $TEST_EXIT_CODE -ne 0 ]]; then
+    if [[ -z "$LAST_RUN_ID" ]]; then
+      LAST_RUN_ID="undefined"
+    fi
+    COMMENT="{health_checker_run_id: ${LAST_RUN_ID}}"
```
theyoprst: It is not a valid json, please generate json using `jq` cli, it is the most safe way

helm/soperator-activechecks/scripts/extensive-check.sh
```diff
@@ -111,13 +115,22 @@ funcs_to_test=(
...
 do
   echo "Running $test on $(hostname)..."
+  LAST_RUN_ID=""
   $test
   TEST_EXIT_CODE=$?
 
   if [[ $TEST_EXIT_CODE -ne 0 ]]; then
+    NODE_NAME=$(hostname)
+    COMPUTE_INSTANCE_ID=$(scontrol show node "$NODE_NAME" -o 2>/dev/null | sed -n 's/.*InstanceId=\([^ ]*\).*/\1/p')
+    COMMENT=$(jq -cn --arg run "$LAST_RUN_ID" --arg inst "$COMPUTE_INSTANCE_ID" \
```
ChessProfessor, line 164: Since all_reduce_in_docker test is not a part of health-checker it doesn't have run_id. 
This is the check that doesn't run periodically, it runs only on cluster creation.
Maybe let's get rid of it in the extensive-check? It may fail on cluster acceptance and we'll notice it anyway and will have to do smth about it.
@rdjjke wdyt?

rdjjke: Yes, let's drop this one

internal/controller/soperatorchecks/slurm_nodes_controller.go
```diff
@@ -201,6 +201,51 @@ func (c *SlurmNodesController) processDegradedNode(
...
+func (c *SlurmNodesController) processExtensiveCheckFailed(
+	ctx context.Context,
+	k8sNode *corev1.Node,
+	slurmClusterName types.NamespacedName,
+	slurmNode slurmapi.Node,
+) error {
+	logger := log.FromContext(ctx).WithName("SlurmNodesController.processExtensiveCheckFailed")
+
+	if !c.enabledNodeReplacement {
+		logger.V(1).Info("skipping extensive check failed processing, node replacement is disabled")
```
theyoprst: (nit) Log messages should be capitalized

helm/soperator-activechecks/scripts/extensive-check.sh
rdjjke, line 1: Can we write job step outputs to separate files?
For example, `/opt/soperator-outputs/slurm_jobs/<node_name>.extensive-check:<step_name>.<job_id>.out`
It corresponds to the following `--output` and `--error` arguments in srun: 
```
--output="%N.%x:%s.%j.out"
--error="%N.%x:%s.%j.out"
-J "<name-of-this-step" # "pasive-checks / all-reduce-with-ib / etc
```

Such files will contain only single-line JSON outputs, making it easier for us to build log dashboards with them in the future.

helm/soperator-activechecks/scripts/extensive-check.sh
```diff
@@ -52,18 +56,6 @@ passive_checks() {
     --log-level info"
 }
 
-all_reduce_in_docker() {
```
theyoprst, line 55: Was it intentional? (no mention in PR message / ticket)

ChessProfessor: Yes, sorry, I resolved the comment too quickly - https://github.com/nebius/soperator/pull/1757#discussion_r2517905462

helm/soperator-activechecks/scripts/extensive-check.sh
ChessProfessor: I've done it, but I need to make sure %N works in --output, I remember there where problems before... 
Maybe we have to set it via env vars, not via --output and --error
I'll test it

rdjjke: It should work, I saw you set this from env vars, but I used this before and it works fine

rdjjke, line 1: What happens if the extensive check doesn't complete in 1 hour? Or if it won't be able to complete until the deadline is reached?
The Slurm job will be in TIMEOUT / DEADLINE state, but how the controller will react on that?

ChessProfessor: Controller treat each state except for FAILED and COMPLETED as ERROR, so it will just update the extensive-check  CR status to ERROR and leave node in the reservation. During the next 10 minutes new extensive-check job will be created and the process will run once again.

rdjjke: Got it. Thank you

PR #1762, opened 2025-11-12
SCHED-370: add soperator test to kind

Uburro:


.github/workflows/one_job.yml
```diff
@@ -8,6 +8,24 @@ on:
...
+        default: false
+
+  workflow_dispatch:
+    inputs:
+      unstable:
+        description: "Build unstable version"
+        type: string
+        required: false
+        default: "true"
+      run_e2e_local:
```
theyoprst: Not used variable?

Didn't you forgot adding something like this to the step?

```
if: ${{ inputs.run_e2e_local == true }}
```

.github/workflows/one_job.yml
```diff
@@ -314,3 +332,98 @@ jobs:
...
+          go-version-file: "go.mod"
+          cache: true
+
+      - name: Download version artifact
+        uses: actions/download-artifact@v6
+        with:
+          name: version
+          path: ./version
+
+      - name: Read version and unstable
```
theyoprst, line 339: Version is read before this command, is it OK?

```
make sync-version UNSTABLE=${UNSTABLE}
```

.github/workflows/one_job.yml
```diff
@@ -314,3 +332,98 @@ jobs:
...
+        uses: actions/upload-artifact@v4
+        with:
+          name: e2e-local-test-results
+          path: |
+            test/e2e-local/*.log
+            test/e2e-local/*.xml
+            cluster-logs/
+          retention-days: 1
+
+      - name: Upload test coverage
```
theyoprst: These tests do not generate coverage report, it could be skipped IMO

Makefile
```diff
@@ -622,8 +622,9 @@ yq: $(YQ) ## Download yq locally if necessary.
 $(YQ): $(LOCALBIN)
 	test -s $(LOCALBIN)/yq || GOBIN=$(LOCALBIN) go install github.com/mikefarah/yq/v4@v$(YQ_VERSION)
 
-.PHONY: install-kind
+.PHONY: kind install-kind
 kind: $(KIND) ## Download kind locally if necessary.
+install-kind: kind ## Alias for kind target.
```
theyoprst: Why do we need aliases? I think other targets installing something are called install-*, so we can use only `install-kind`?

test/utils/utils.go
```diff
@@ -0,0 +1,280 @@
...
+		return wd, err
+	}
+	wd = strings.Replace(wd, "/test/e2e-local", "", -1)
+	wd = strings.Replace(wd, "/test/e2e", "", -1)
+	return wd, nil
+}
+
+// UncommentCode searches for target in the file and remove the comment prefix
+// of the target content. The target content may span multiple lines.
+func UncommentCode(filename, target, prefix string) error {
```
theyoprst: `UncommentCode` is not used anywhere, same for `RestartKindCluster`, `UndeployFlux`, do we need them?

.github/workflows/one_job.yml
```diff
@@ -8,6 +8,24 @@ on:
         type: string
         required: false
         default: "true"
+      run_e2e_local:
```
theyoprst: These are now not true e2e (not testing slurm functionality), but even if we had that, we need to distinguish them from current e2e. Maybe rename to integration tests / smoke tests / kind tests / whatever?

test/utils/utils.go
```diff
@@ -0,0 +1,280 @@
+package utils
```
theyoprst: "Utils" is kind of forbidden work for packages :-)

Claude suggest naming it testenv, what do you think of it?

```
  Looking at what's actually in the package, here are better naming options:

  Recommended: testenv

  Rationale:
  - These functions set up and manage the test environment (Kind, FluxCD, kubectl operations)
  - Clear purpose: "test environment management"
  - Common in Go projects (e.g., go/src/cmd/go/internal/testenv)

  Usage:
  import "nebius.ai/slurm-operator/test/testenv"

  testenv.CreateKindCluster(ctx)
  testenv.DeployFlux(ctx, unstable)
  testenv.IsSlurmClusterCRDsInstalled(ctx)

  Alternative Options:

  kindtest

  Pros: Specific to Kind-based testing
  Cons: Some functions are FluxCD-specific, not just Kind

  cluster or testcluster

  Pros: Describes what it manages
  Cons: Might conflict with domain types (SlurmCluster)

  fixture or testfixture

  Pros: Standard testing term for test setup/teardown
  Cons: Usually implies data fixtures, not infrastructure

  harness

  Pros: Testing infrastructure term
  Cons: Less common in Go, might be unfamiliar

  setup

  Pros: Clear purpose
  Cons: Too generic, doesn't indicate it's for testing

  Structure Recommendation:

  If the package grows, consider splitting by responsibility:

  test/
  ├── testenv/           # Environment setup (Kind, FluxCD)
  │   ├── kind.go       # Kind cluster operations
  │   ├── flux.go       # FluxCD operations
  │   └── kubectl.go    # kubectl helpers
  ├── integration/      # Integration tests (renamed from e2e-local)
  │   ├── suite_test.go
  │   └── deployment_test.go
  └── e2e/              # Future true E2E tests
      └── slurm_test.go # Actual job execution tests

  My Recommendation:

  Rename to testenv because:
  1. ✅ Clear purpose: test environment management
  2. ✅ Follows Go stdlib conventions
  3. ✅ Room to grow without becoming a dumping ground
  4. ✅ Reads naturally: testenv.CreateKindCluster()
  5. ✅ Distinguishes from production code packages

  Import path:
  package testenv  // not utils

  // File: test/testenv/kind.go
  // File: test/testenv/flux.go
  // File: test/testenv/crds.go
```

.github/workflows/one_job.yml
```diff
@@ -314,3 +332,98 @@ jobs:
           make sync-version UNSTABLE=${UNSTABLE}
 
           make release-helm UNSTABLE="${UNSTABLE}"
+
+  e2e-local-test:
+    name: Run E2E Local Tests
```
asteny: Why is it local? `runs-on: ubuntu-latest`

PR #1763, opened 2025-11-12
SCHED-248 Set-unhealthy on extensive check failure

ChessProfessor:
## Problem

When extensive check fails we need to mark k8s node with HardwareIssueSuspected condition for replacment

## Solution

Setting this condition

## Testing

Tested on dev cluster

### DIFF
--- a/helm/soperator-activechecks/scripts/extensive-check.sh
+++ b/helm/soperator-activechecks/scripts/extensive-check.sh
@@ -119,6 +119,7 @@ funcs_to_test=(
 for test in "${funcs_to_test[@]}"
 do
   echo "Running $test on $(hostname)..."
+  LAST_RUN_ID=""
   $test
   TEST_EXIT_CODE=$?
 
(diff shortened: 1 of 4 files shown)


helm/soperator-activechecks/scripts/extensive-check.sh
```diff
@@ -119,15 +119,17 @@ funcs_to_test=(
...
   $test
   TEST_EXIT_CODE=$?
 
   if [[ $TEST_EXIT_CODE -ne 0 ]]; then
     if [[ -z "$LAST_RUN_ID" ]]; then
       LAST_RUN_ID="undefined"
     fi
-    COMMENT="{health_checker_run_id: ${LAST_RUN_ID}}"
     NODE_NAME=$(hostname)
+    COMPUTE_INSTANCE_ID=$(scontrol show node "$NODE_NAME" -o 2>/dev/null | sed -n 's/.*InstanceId=\([^ ]*\).*/\1/p')
```
ChessProfessor, line 131: I decided that it's simpler to set full message we need for set-unhealthy in one place (to slurm node comment), but I can also merge COMPUTE_INSTANCE_ID to the json message in the controller itself.

PR #1783, opened 2025-11-17
Fix bug in maintenanceIgnoreNodeLabels 

ali-sattari:
## Problem

There was a bug in the new feature that parses env vsr for a list of `key=value` items to ignore nodes for maintenance, it did not support repeated keys.

## Solution

Now `maintenanceIgnoreNodeLabels` supports repeated keys with different values by making a `map[string][]string` to store them.

## Testing

See unit tests.

## Release Notes

Fix bug in maintenanceIgnoreNodeLabels to support same key with different values

### DIFF
--- a/internal/controller/clustercontroller/controller.go
+++ b/internal/controller/clustercontroller/controller.go
@@ -108,6 +108,35 @@ func (r SlurmClusterReconciler) ReconcileControllers(
 					return nil
 				},
 			},
+			utils.MultiStepExecutionStep{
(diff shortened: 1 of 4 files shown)


Uburro:
> A while ago https://github.com/nebius/soperator/issues/1330 were introduced to help with image pull issues, now it we don't need that anymore.

I'm not sure I agree with that.

ChessProfessor (approved):

PR #1822, opened 2025-11-23
SCHED-417, SCHED-430, SCHED-437, SCHED-439, SCHED-441, SCHED-429: Pre-release 1.23 fixes related to extensive checks

rdjjke:
## Problem
1. `health-checker` in extensive checks can't collect CUDA version because it's running without PATH from /etc/profile
2. job step id is written to `job_id` log label, and job id is written to `job_array_id` log label
3. `passive-checks` step in `extensive-check` job always fails with an error because it runs as a normal user (can't access kernel log)
4. K8s node conditions set after extensive check failures have the wrong Reason and invalid JSON in the Message
5. Extensive checks always pass because they search for errors in a string that is always empty
6. After a reproducible issue found by a passive check, extensive checks won't run because the same passive check fails in Prolog / HealthCheckProgram, making it impossible for the slurm job to start.

## Solution
1. Use `bash -l` instead of `bash`
2. Adjusted extensive check job step output format
3. Execute `health-checker` with `sudo`
4. Fixed K8s node conditions' Reason and Message
5. Instead of parsing job steps' stdouts, read their output files
6. Introduced new filter `skip_for_reservation_prefixes` for passive checks

## Testing
TODO

## Release Notes
<Nothing>


Uburro (dismissed):

helm/soperator-activechecks/scripts/extensive-check.sh
```diff
@@ -152,24 +195,20 @@ funcs_to_test=(
...
     NODE_NAME=$(hostname)
-    COMPUTE_INSTANCE_ID=$(scontrol show node "$NODE_NAME" -o 2>/dev/null | sed -n 's/.*InstanceId=\([^ ]*\).*/\1/p')
-    COMMENT=$(jq -cn --arg run "$LAST_RUN_ID" --arg inst "$COMPUTE_INSTANCE_ID" \
-          '{health_checker_run_id: $run, compute_instance_id: $inst}')
-    echo "Setting node comment: $COMMENT"
-    sudo scontrol update NodeName=$NODE_NAME Comment="$COMMENT"
-
-    echo "$test reported failure (exit code $TEST_EXIT_CODE)."
+    echo "Setting comment on node $NODE_NAME"
+    COMPUTE_INSTANCE_ID=$(scontrol show node worker-0 --json | jq -r '.nodes[0].instance_id')
```
theyoprst: `worker-0` -> `$NODE_NAME`?

helm/soperator-activechecks/scripts/extensive-check.sh
```diff
@@ -152,24 +195,20 @@ funcs_to_test=(
...
+    COMMENT=$(jq -cn --arg run "$HC_RUN_ID" --arg inst "$COMPUTE_INSTANCE_ID" \
+          '{"health_checker_run_id": "$run", "compute_instance_id": "$inst"}')
```
theyoprst: bug: quotes around $run and $inst variables break the var substitution

```sh
➜ jq -cn --arg run "$HC_RUN_ID" --arg inst "$COMPUTE_INSTANCE_ID" \
                    '{"health_checker_run_id": "$run", "compute_instance_id": "$inst"}'
{"health_checker_run_id":"$run","compute_instance_id":"$inst"}
```

It should be:

```sh
➜ jq -cn --arg run "$HC_RUN_ID" --arg inst "$COMPUTE_INSTANCE_ID" \
                    '{"health_checker_run_id": $run, "compute_instance_id": $inst}'
{"health_checker_run_id":"","compute_instance_id":""}
```

rdjjke: Fixed. Thank you

helm/soperator-activechecks/scripts/extensive-check.sh
rdjjke: Fixed. Thank you

theyoprst (approved):

PR #1823, opened 2025-11-24
bump node exporter version to latest stable

ali-sattari:
## Problem

Node exporter v1.6.0 has [this issue](https://github.com/prometheus/node_exporter/issues/3017) with NFSd. Resulting in no metrics being collected.
## Solution

Bumping the version to the latest stable works: [1.10.2](https://github.com/prometheus/node_exporter/releases/tag/v1.10.2) at the time.

## Testing

-
## Release Notes

-

### DIFF
--- a/helm/nfs-server/Chart.yaml
+++ b/helm/nfs-server/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: helm-nfs-server
 description: A Helm chart for NFS server deployment
 type: application
-version: 1.1.0
-appVersion: "1.1.0"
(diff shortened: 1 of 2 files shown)


NFS_VERSION
```diff
@@ -1 +1 @@
-1.1.0
+1.2.0
```
asteny, line 1: How this file helps here? Did you create some commands to the makefile?

ali-sattari: Yes, `make sync-version` works for NFS too.

PR #1900, opened 2025-12-03
Support container customEnv + enable video capability by default

rdjjke:
## Problem
Currently, K8s device plugin only mounts GPU driver libraries related to "compute" and "utility" capabilities (default values).
In some cases, "video" capability is also needed.

## Solution
To enable the "video" capability, we need to add `NVIDIA_DRIVER_CAPABILITIES` env var to `slurmd` container.
To propagate these libraries from the container FS to the jail FS, we also need to make `nvidia-container-cli` respect this env var.

## Testing
1. Create a test cluster
2. Make sure all active checks succeeded
3. Make sure this works from jail on worker nodes: `ldconfig -p | grep nvcuvid`

## Release Notes
Add support for NVENC.

### DIFF
--- a/internal/render/worker/container.go
+++ b/internal/render/worker/container.go
@@ -86,14 +86,17 @@ func renderContainerSlurmd(
 		ImagePullPolicy: container.ImagePullPolicy,
 		Command:         container.Command,
 		Args:            container.Args,
-		Env: renderSlurmdEnv(
-			clusterName,
-			cgroupVersion,
-			clusterType,
-			realMemory,
-			enableGDRCopy,
-			slurmNodeExtra,
-			workerFeatures,
+		Env: append(
+			renderSlurmdEnv(
+				clusterName,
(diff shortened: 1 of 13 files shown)


internal/render/worker/container.go
```diff
@@ -211,16 +214,19 @@ func renderContainerNodeSetSlurmd(
...
-			"",
-			nil,
+		Env: append(
+			renderSlurmdEnv(
+				nodeSet.ParentalCluster.Name,
+				nodeSet.CgroupVersion,
+				utils.Ternary(nodeSet.GPU.Enabled, consts.ClusterTypeGPU, consts.ClusterTypeCPU),
+				realMemory,
+				nodeSet.GPU.Nvidia.GDRCopyEnabled,
+				//
```
asteny, line 224: ?

asteny (approved):

PR #1901, opened 2025-12-04
SCHED-492 Upgrade cuda version and get rid of dcgmi in ansible

ChessProfessor:
## Problem

We want to have two jail images with different cuda base (12 and 13). We already added required dcgmi to cuda base image, now I get rid of it in the jail (since it's already added to cuda base)

### DIFF
--- a/ansible/roles/dcgmi/tasks/main.yml
+++ b/ansible/roles/dcgmi/tasks/main.yml
@@ -1,7 +1,7 @@
 ---
 
 - name: Detect CUDA toolkit version with nvcc
-  ansible.builtin.shell: "bash -lc 'nvcc -V'"
+  ansible.builtin.shell: "/usr/local/cuda/bin/nvcc -V"
   args:
     executable: /bin/bash
   register: dcgmi_nvcc_version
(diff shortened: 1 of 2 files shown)


ansible/roles/dcgmi/tasks/main.yml
asteny, line 1: I dont know why you moved dcgmi to the base cuda image, but we can remove role here, because we need it for upgrading versions. After [this PR ](https://github.com/nebius/ml-containers/pull/25/files)we will have 2 places for dcgmi version...

asteny (changes requested):

ansible/roles/dcgmi/tasks/main.yml
ChessProfessor: You mean we can't remove role here? 
There are two different dcgmi packages for different cudas, and since we want to support jail with both cudas I moved all cuda-related to cuda_base image. If we want to leave dcgmi role in ansible we have to support both dcgmi packages for different cuda version...

asteny: If we want to support dcgmi inside the jail we need to control package version and so on, and we cant remove this role (thats why I wrote it)

PR #1913, opened 2025-12-08
SCHED-490 Add a flag to disable extensive check

ChessProfessor:
## Problem

Now we don't have a reliable way to disable extensive check but still have set-unhealthy after any HC failure

## Solution

Adding disable-extensive-check flag for it

### DIFF
--- a/cmd/soperatorchecks/main.go
+++ b/cmd/soperatorchecks/main.go
@@ -112,7 +112,7 @@ func main() {
 		logFormat                   string
 		logLevel                    string
 		enabledNodeReplacement      bool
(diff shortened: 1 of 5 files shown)


cmd/soperatorchecks/main.go
```diff
@@ -112,6 +112,7 @@ func main() {
 		logFormat                   string
 		logLevel                    string
 		enabledNodeReplacement      bool
+		disableExtensiveCheck       bool
```
asteny: maybe enableExtensiveCheck?

PR #1927, opened 2025-12-11
feat: Adding node metrics to track node status in Slurm

mateusclira-nv:
## Problem

The SLURM exporter doesn't have some Node metrics needed for comprehensive dashboards and monitoring:

1. **Node State Flags**: The `slurm_node_info` metric lacks flag-based state indicators (COMPLETING, FAIL, PLANNED) that are needed for accurate node state tracking and alerting.

2. **CPU & Memory Metrics**: No separate metrics for Total, Allocated, and Idle resources at the node level, making it difficult to:
   - Monitor resource utilization trends
   - Create capacity planning dashboards
   - Track resource availability over time

3. **Partition-Level Aggregation**: Missing a metric to map nodes to partitions, preventing partition-level resource aggregation using PromQL joins.

4. **Job-Level Resource Metrics**: No simple metrics for allocated CPU and memory per job, limiting job-level resource monitoring and cost analysis.

## Solution

Added the following new Prometheus metrics to the SLURM exporter:

### Node State Flags (added to `slurm_node_info`)
- `state_is_completing`: Boolean flag indicating COMPLETING state
- `state_is_fail`: Boolean flag indicating FAIL state  
- `state_is_planned`: Boolean flag indicating PLANNED state

### Node-Level CPU Metrics
- `slurm_node_cpus_total`: Total number of CPUs on the node
- `slurm_node_cpus_allocated`: Number of CPUs currently allocated
- `slurm_node_cpus_idle`: Number of idle CPUs
- `slurm_node_cpus_effective`: Effective CPUs (excluding specialized CPUs)

### Node-Level Memory Metrics
- `slurm_node_memory_total_bytes`: Total memory in bytes
- `slurm_node_memory_allocated_bytes`: Allocated memory in bytes
- `slurm_node_memory_free_bytes`: Free memory in bytes
- `slurm_node_memory_effective_bytes`: Effective memory in bytes (excluding specialized memory)

### Partition Mapping
- `slurm_node_partition{node_name, partition}`: Maps nodes to their partitions, enabling partition-level aggregation via PromQL joins

### Job-Level Resource Metrics
- `slurm_job_cpus{job_id}`: Number of CPUs allocated to the job
- `slurm_job_memory_bytes{job_id}`: Memory in bytes allocated to the job

**Implementation Details:**
- Extended `slurmapi.Node` and `slurmapi.Job` structs to parse additional fields from SLURM API responses
- Updated `MetricsCollector` to expose new metrics with proper labels
- Added helper functions for memory unit conversion (MB to bytes)
- All metrics follow existing naming conventions and label patterns

## Testing

All unit tests pass (make test)
E2E tests will be run by creating this PR
## Release Notes

Feature: Added comprehensive resource metrics

### DIFF
--- a/pkg/valuesgen/generated/soperatorfluxcd/values.go
+++ b/pkg/valuesgen/generated/soperatorfluxcd/values.go
@@ -1590,7 +1590,7 @@ func NewDefaults() Values {
 			ReleaseName: "soperator-storageclasses",
 			Timeout:     "5m",
 			Values:      nil,
-			Version:     "1.23.0",
+			Version:     "1.23.1",
 		},
 		Tailscale: ValuesTailscale{
 			DriftDetection: ValuesTailscaleDriftDetection{


internal/slurmapi/node.go
```diff
@@ -128,6 +164,7 @@ var baseStates = []api.V0041NodeState{
 	api.V0041NodeStateALLOCATED,
 	api.V0041NodeStateERROR,
 	api.V0041NodeStateMIXED,
+	api.V0041NodeStateCOMPLETING,
```
theyoprst: This contradicts the Slurm state model. As discussed and documented in https://github.com/SchedMD/slurm/blob/master/slurm/slurm.h.in:

1. The node state is a 32-bit integer where the lowest 4 bits (0x0000000f) encode exactly 6 mutually exclusive base states: IDLE, DOWN, ALLOCATED, ERROR, MIXED, UNKNOWN
2. COMPLETING is a flag bit, not a base state:
#define NODE_STATE_COMPLETING SLURM_BIT(10) /* node is completing allocated job */
3. A node can be IDLE+COMPLETING simultaneously (e.g., State=IDLE+COMPLETING+DYNAMIC_NORM+NOT_RESPONDING)

It would be great to add detailed comments here about this, I can do it later

internal/slurmapi/node.go
```diff
@@ -72,10 +83,20 @@ func NodeFromAPI(node api.V0041Node) (Node, error) {
 		States:      nodeStates,
 		Partitions:  *node.Partitions,
 		Tres:        *node.Tres,
+		TresUsed:    valueOrDefault(node.TresUsed),
```
theyoprst: `TresUsed` is not used (it's set but is never read). Maybe remove?

internal/exporter/collector.go
```diff
@@ -339,15 +373,144 @@ func (c *MetricsCollector) slurmNodeMetrics(slurmNodes []slurmapi.Node) iter.Seq
 				strconv.FormatBool(node.IsDrainState()),
 				strconv.FormatBool(node.IsMaintenanceState()),
 				strconv.FormatBool(node.IsReservedState()),
+				strconv.FormatBool(node.IsCompletingState()),
```
theyoprst: Potentially it could be required adding all possible state flags (there are a couble of dozen nowadays), and we can hit the Victoria Metrics limit of 30 labels. So it's better to report not "true"/"false", but "true"/"" (empty label === no label). It's my historical mistake to report "true". Let's report empty value for new flags? You can use some helper like

```go
  func boolToLabelValue(b bool) string {
        if b {
                return "true"
        }
        return ""
  }
```

But please keep `strconv.FormatBool` for current flags for backward-compatibility.

internal/exporter/collector.go
```diff
@@ -339,15 +373,144 @@ func (c *MetricsCollector) slurmNodeMetrics(slurmNodes []slurmapi.Node) iter.Seq
...
+	if node.CPUs != nil {
+		return float64(*node.CPUs), true
+	}
+	if tres, err := slurmapi.ParseTrackableResources(node.Tres); err == nil && tres.CPUCount > 0 {
+		return float64(tres.CPUCount), true
+	}
+	return 0, false
+}
+
+func nodeCPUAllocated(node slurmapi.Node) (float64, bool) {
```
theyoprst: The Node struct in the `slurmapi` package already has methods like `IsDrainState()`, `IsMaintenanceState()`, `BaseState()`, so adding resource accessor methods would be consistent.
Could you convert these function to the methods?

theyoprst:
Merged, thanks @mateusclira-nv !
`values.yaml` fix was necessary as well, we broke the builds it accidentally before the New Year

PR #1928, opened 2025-12-11
Adjust rmem and wmem network sysctls by default

asteny:
## Problem

sysctl defaults are too low. Network speed is low.


Uburro (dismissed):

helm/nodeconfigurator/values.yaml
```diff
@@ -1,10 +1,16 @@
-nodeConfigurator:
+customContainer:
   enabled: false
+  hostNetwork: true
```
Uburro: it's podSpec not container spec

PR #1980, opened 2026-01-02
Allow set procMount

### LINKED ISSUE #1510: hostUsers: false incompatible with NFS storage - idmap mounts not supported
**Issue Description**:
PR #1482 set `hostUsers: false` for pods to address security concerns with `procMount: Unmasked`. However, this breaks deployments using NFS storage due to lack of idmap mount support.

**Problem**:
When `hostUsers: false` is set, Kubernetes attempts to use idmap mounts for secure UID/GID mapping. NFS filesystems don't support idmap mounts

https://kubernetes.io/blog/2025/04/25/userns-enabled-by-default/#everything-you-wanted-to-know-about-user-namespaces-in-kubernetes
```
Most popular file-systems are supported, the notable absence that isn't supported yet is NFS.
```

From kubelet:
```
Error: container create failed: mount_setattr `/mnt/jail` (maybe the file system used doesn't support idmap mounts on this kernel?): Invalid argument
```

andriishestakov:
Resolves #1510


Uburro:
@andriishestakov Hi. Thanks for the PR. Could you rebase it, plz ?

PR #1991, opened 2026-01-06
SCHED-696: update umbrella chart for logging

itechdima:
## Problem

In some cases we need to be able to use tokens from nodes for public login.

## Solution

Add option to use token from the node.

## Release Notes

Feature: add `publicEndpointTokenKind` value for observability to choose what token to use for public logging.


helm/soperator-fluxcd/templates/opentelemetry-collector-events.yaml
```diff
@@ -83,7 +83,11 @@ spec:
           endpoint: 0.0.0.0:13133
         {{- if $hasPublicEndpoint }}
         bearertokenauth:
+        {{- if eq .Values.observability.publicEndpointTokenKind "secret" }}
           filename: "/o11ytoken/accessToken"
+        {{- else }}
```
theyoprst: How about checking for alternative name? Like 

```
{{- else if eq .Values.observability.publicEndpointTokenKind "node" }}
```

Or "node-local". It would be easier to find this alternative key usage in the future.
But it's up to you.

theyoprst (dismissed):

PR #2010, opened 2026-01-12
SCHED-394 Build and populate jail for both cuda 12 and 13

ChessProfessor:
## Problem

Currently we cannot update jail image in 1.23 using cuda13 for B300, adding such functionality (for 1.23 ONLY)


images/jail/jail.dockerfile
```diff
@@ -1,9 +1,10 @@
 # syntax=docker.io/docker/dockerfile-upstream:1.20.0
 
-FROM cr.eu-north1.nebius.cloud/soperator/cuda_base:12.9.0-ubuntu24.04-nccl2.26.5-1-17be0c5 AS cuda
```
asteny, line 3: Add a comment plz https://github.com/nebius/ml-containers/pull/44

PR #2043, opened 2026-01-15
SCHED-775: Fix image for wait-for-soperatorchecks-srun-ready

theyoprst:
## Problem

`wait-for-soperatorchecks-srun-ready` job fails blocking all further slurm jobs:

```
runuser: failed to execute srun: No such file or directory
```

Root cause: it usess wrong image `k8s_check_job` instead of `slurm_check_job`.

## Solution

The `mustMerge` function gives precedence to the first argument, but the code was passing default dict first (which contains k8s_check_job image for k8s jobs.

The Fix: swap mustMerge arguments so user-specified parameters take precedence.

## Testing

Add a test that image can be overriden for the job.

## Release Notes

None


.github/workflows/one_job.yml
```diff
@@ -276,6 +276,22 @@ jobs:
           echo "Building and pushing image of the slurm_check_job ${IMAGE_VERSION}"
           make docker-build-and-push DOCKER_BUILD_ARGS="${DOCKER_BUILD_ARGS}" UNSTABLE="${UNSTABLE}" IMAGE_NAME=slurm_check_job DOCKERFILE=slurm_check_job/slurm_check_job.dockerfile IMAGE_VERSION=${IMAGE_VERSION}
 
+          echo "=== DEBUG: Verifying slurm_check_job image contents ==="
```
asteny: I believe you add this just to test it right now and we dont want to merge that)

theyoprst: Yes, don't worry, it will not be merged this way :-)

PR #2052, opened 2026-01-19
SCHED-785: Plug-in SPANK plugins properly in slurm job active checks

theyoprst:
## Problem

In slurm job active check container, `/usr/lib/slurm` misses required plugins which are mentioned in `/etc/slurm/plugstack.conf`:
- `chroot.so` (required)
- `spank_pyxis.so` (required)
- `spanknccldebug.so` (optional)

## Solution

On workers and login nodes these files are bind-mounted in 

## Testing

None

## Release Notes

Loading spanknccldebug.so error is no longer appearing in slurm active check jobs:

```
srun: error: plugin_load_from_file: dlopen(spanknccldebug.so): spanknccldebug.so: cannot open shared object file: No such file or directory
```

### DIFF
--- a/images/slurm_check_job/slurm_check_job.dockerfile
+++ b/images/slurm_check_job/slurm_check_job.dockerfile
@@ -34,6 +34,13 @@ RUN apt-get update && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
 
+# Install NCCL debug plugin
+COPY images/common/spank-nccl-debug/src /usr/src/soperator/spank/nccld-debug
+COPY images/common/scripts/install_nccld_debug_plugin.sh /opt/bin/
+RUN chmod +x /opt/bin/install_nccld_debug_plugin.sh && \
+    /opt/bin/install_nccld_debug_plugin.sh && \
+    rm /opt/bin/install_nccld_debug_plugin.sh
+
 # Install kubectl
 RUN ARCH="$(uname -m | sed 's/x86_64/amd64/; s/aarch64/arm64/')" && \
     KUBECTL_VERSION="$(curl -Ls https://dl.k8s.io/release/stable.txt)" && \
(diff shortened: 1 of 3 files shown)


images/login/sshd.dockerfile
```diff
@@ -60,6 +60,12 @@ RUN apt-get update && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
 
+# Create single folder with slurm plugins for all architectures
+RUN mkdir -p /usr/lib/slurm && \
```
asteny: There is ansible role with slurm https://github.com/nebius/soperator/blob/a8cde4ff0c33093047ab5f6e1c43b272bc8512e7/ansible/roles/slurm-divert/tasks/main.yml#L12

images/slurm_check_job/slurm_check_job.dockerfile
```diff
@@ -28,12 +28,18 @@ RUN chown 0:0 /etc/enroot/enroot.conf && \
...
 ARG PYXIS_VERSION=0.21.0
-# Install slurm pyxis plugin \
+# Install slurm pyxis plugin
 RUN apt-get update && \
     apt -y install nvslurm-plugin-pyxis=${SLURM_VERSION}-${PYXIS_VERSION}-1 && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
 
+# Create single folder with slurm plugins for all architectures
+RUN mkdir -p /usr/lib/slurm && \
```
asteny: We do not need that, because this libraries will be bind mounted from containers https://github.com/nebius/soperator/blob/410c1532752a5e92cfd5df4c03850ac2a9dddaf4/images/common/scripts/complement_jail.sh#L149

asteny (changes requested):

.github/workflows/one_job.yml
```diff
@@ -300,6 +300,65 @@ jobs:
           echo "Removing jail rootfs tar archive to speedup further docker builds."
           rm -f images/jail_rootfs*.tar
 
+      - name: Test SPANK plugins are accessible
```
asteny: I dont think that we need such tests in CI

asteny (approved):

PR #2131, opened 2026-02-04
[SCHED-885] Fix initContainers order in pods

theyoprst:
## Solution

Custom init containers can now specify runsBefore to indicate they should run before specific system init containers.


api/v1/slurmcluster_types.go
```diff
@@ -566,6 +566,19 @@ type VolumeSource struct {
...
+// It allows specifying dependencies between init containers to control execution order.
+type InitContainer struct {
+	corev1.Container `json:",inline"`
+
+	// RunsBefore specifies names of system init containers that must run after this one.
+	// Can only reference system init containers (e.g., "munge", "wait-for-accounting").
+	// Custom containers maintain their relative order from the array.
+	//
+	// +kubebuilder:validation:Optional
+	RunsBefore []string `json:"runsBefore,omitempty"`
```
Uburro, line 579: Let's do not do that. I would prefer user kuberntetes spec way

PR #2136, opened 2026-02-05
fix retrigger checks

itechdima:

### DIFF
--- a/helm/soperator-activechecks/templates/_helpers.tpl
+++ b/helm/soperator-activechecks/templates/_helpers.tpl
@@ -210,7 +210,7 @@ jobContainer:
 {{- end }}
 {{- if $env }}
   env:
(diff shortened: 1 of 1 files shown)


helm/soperator-activechecks/scripts/retrigger-checks.py
```diff
@@ -1,11 +1,10 @@
-import os
 import subprocess
 import logging
 import time
 import json
 import datetime
 
-NS = os.environ["NAMESPACE"]
+NS = "{{ .Release.Namespace }}"
```
asteny: Lets put that not inside code
Maybe like https://github.com/nebius/soperator/blob/main/helm/soperator-activechecks/templates/_helpers.tpl#L177

asteny (approved):

PR #2141, opened 2026-02-05
[SCHED-804] Deprecate and make optional slurmNodes.worker field in SlurmCluster CRD

theyoprst:
## Problem

We still use deprecated SlurmCluster.worker:
- set it in terraform
- use these data in controller

## Solution

Make SlurmCluster Worker field truly optional, next changes were done:

- Change Worker from value type to pointer (*SlurmNodeWorker)
- Fix CRD validation
- Add nil guards throughout controllers, values, and render packages
- Wrap worker section in Helm chart with conditional
- Handle SlurmdPort per-node in NodeSets mode
- Conditionally emit global SlurmdPort only in legacy mode
- UseDefaultAppArmorProfile usage was fixed

Also stop specifying worker in terraform:
https://github.com/nebius/nebius-solutions-library/pull/780

## Testing

Tested manually: dev cluster is provisioned successfully.

## Release Notes

Deprecated worker data in SlurmCluster CRD are not used by the soperator controller anymore.

### DIFF
--- a/api/v1/slurmcluster_types.go
+++ b/api/v1/slurmcluster_types.go
@@ -1007,7 +1007,7 @@ type SlurmNodeWorkerVolumes struct {
 	// JailSubMounts represents the sub-mount configurations within the jail volume
 	//
 	// +kubebuilder:validation:Optional
-	JailSubMounts []NodeVolumeMount `json:"jailSubMounts"`
+	JailSubMounts []NodeVolumeMount `json:"jailSubMounts,omitempty"`
 
 	// CustomMounts represents the custom mount configurations
 	//
@@ -1087,7 +1087,7 @@ type SlurmNodeLoginVolumes struct {
 	// JailSubMounts represents the sub-mount configurations within the jail volume
 	//
 	// +kubebuilder:validation:Optional
-	JailSubMounts []NodeVolumeMount `json:"jailSubMounts"`
+	JailSubMounts []NodeVolumeMount `json:"jailSubMounts,omitempty"`
 
 	// CustomMounts represents the custom mount configurations
 	//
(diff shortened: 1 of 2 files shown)


internal/render/common/configmap.go
```diff
@@ -455,7 +465,13 @@ func addSlurmConfigProperties(res *renderutils.PropertiesConfig, config interfac
...
 
 func generateCGroupConfig(cluster *values.SlurmCluster) renderutils.ConfigFile {
-	defaultLines := strings.Split(renderDefaultCGroupConfig(cluster.NodeWorker.CgroupVersion), "\n")
+	cgroupVersion := cluster.NodeWorker.CgroupVersion
+	if cgroupVersion == "" {
+		// TODO: it's better to use "autodetect", but right now we have some checks in the entrypoint
+		// that rely on "v2" value.
+		cgroupVersion = consts.CGroupV2
+	}
+	defaultLines := strings.Split(renderDefaultCGroupConfig(cgroupVersion), "\n")
```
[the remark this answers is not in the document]
theyoprst: We use only v2 now, cgroups from node groups should be deprecated

internal/render/common/configmap.go
```diff
@@ -455,7 +465,13 @@ func addSlurmConfigProperties(res *renderutils.PropertiesConfig, config interfac
 }
 
 func generateCGroupConfig(cluster *values.SlurmCluster) renderutils.ConfigFile {
-	defaultLines := strings.Split(renderDefaultCGroupConfig(cluster.NodeWorker.CgroupVersion), "\n")
+	cgroupVersion := cluster.NodeWorker.CgroupVersion
```
Uburro: Maybe should we move it to slurmcluster.spec this values? @rdjjke what do you think?

rdjjke: But `cluster.NodeWorker` is `null`, isn't it?
So yes, I think we should move it to `cluster.CgroupVersion`

theyoprst: Done

Uburro (dismissed):

Uburro (approved):

PR #2183, opened 2026-02-19
SCHED-897 Remove empty libnvidia.so files in populate_jail

ChessProfessor:
## Problem

After cluster upgrade there might be some empty .so leftovers which can lead to broken pyxis

## Solution

Removing these files in populate_jail

## Testing

- Empty .so files should be deleted on populating jail
- After deleting populate_jail k8s job it shouldn't be recreated when there are non-terminated login or worker pods

### DIFF
--- a/images/jail/populate_jail_entrypoint.sh
+++ b/images/jail/populate_jail_entrypoint.sh
@@ -30,7 +30,9 @@ remove_empty_lib_mount_targets() {
     echo "Removing empty library files that were used as bind-mount targets on the previous cluster"
     ARCH_LIST="x86_64 aarch64"
     for arch in $ARCH_LIST; do
-        find "/mnt/jail/lib/${arch}-linux-gnu" \
+        dir="/mnt/jail/lib/${arch}-linux-gnu"
+        [ -d "$dir" ] || continue
+        find "$dir" \
             -maxdepth 1 -type f -empty -name '*.so*' -print |
         while IFS= read -r file; do
             echo "Removing $file"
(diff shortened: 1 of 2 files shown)


images/jail/populate_jail_entrypoint.sh
```diff
@@ -23,6 +23,22 @@ populate_jail_rootfs() {
...
 
+remove_empty_lib_mount_targets() {
+    echo "Removing the flag file that shows that GPU library bind-mount targets exist"
+    rm -f "/mnt/jail/etc/gpu_libs_installed.flag"
+
+    echo "Removing empty library files that were used as bind-mount targets on the previous cluster"
+    ARCH_LIST="x86_64 aarch64"
+    for arch in $ARCH_LIST; do
+        find "/mnt/jail/lib/${arch}-linux-gnu" \
+            -maxdepth 1 -type f ! -type l -empty -print |
```
theyoprst: The PR title is not quite correct: we remove any empty file. But maybe we should to rename only *.so files? In that case, `-name '*.so*'` should be eadded

theyoprst: `! -type l` isn't it redundant? We already have `-type f` which already excludes symlinks according to an LLM

internal/controller/clustercontroller/populate_job.go
```diff
@@ -21,13 +22,16 @@ import (
...
 
+const populateJailRequeueDuration = 10 * time.Second
+
 // ReconcilePopulateJail reconciles and wait all resources necessary for deploying Populate Jail Job
 func (r SlurmClusterReconciler) ReconcilePopulateJail(
 	ctx context.Context,
 	clusterValues *values.SlurmCluster,
 	cluster *slurmv1.SlurmCluster,
-) error {
+) (bool, error) {
```
theyoprst: It is hard to guess what bool means. Maybe use ctrl.Result of bool instead? It's widely used, search for `(ctrl.Result, error)`

PR #2184, opened 2026-02-19
docs: add Tailscale documentation (README  and docs page)

dasbatta:
## Summary
Adds public documentation for enabling Tailscale access to Soperator login pods.

## Changes
- Expand README "Tailscale support" section with high-level usage
- Add `docs/tailscale.md` with verbose steps

### DIFF
--- a/docs/tailscale.md
+++ b/docs/tailscale.md
@@ -90,6 +90,10 @@ Wait for login pods to restart:
 ```bash
 kubectl -n soperator get pods -l app.kubernetes.io/component=login -w
 ```
+
+> Note: The Kubernetes Secret referenced by `TS_KUBE_SECRET` is created/updated automatically by the Tailscale container to persist its state.
+> In this setup `TS_KUBE_SECRET` is set to the pod name (`metadata.name`), so each login pod typically uses a Secret with the same name as the pod.
+> Deleting that Secret forces the pod to re-authenticate (a new auth URL will appear in logs).
 ---
 
 ## 3) Authenticate login pods


docs/tailscale.md
```diff
@@ -0,0 +1,101 @@
+# Tailscale support (login pods)
+
+Soperator can optionally enable [Tailscale](https://github.com/tailscale/tailscale) on login pods (`login-0`, `login-1`) so you can SSH to the cluster over your Tailnet.
+
+When enabled via Terraform, Soperator:
```
asteny: Terraform is a Nebius specific thing, which will be removed soon because of Managed solution. We are writing docs for using Soperator as is without Terraform

docs/tailscale.md
```diff
@@ -0,0 +1,101 @@
...
+```
+
+In the output, look for:
+
+```text
+To authenticate, visit:
+  https://login.tailscale.com/a/...
+```
+
+A Tailnet admin must open each URL and approve/authenticate both devices (`login-0` and `login-1`) in the Tailscale admin console.
```
asteny: You can have 1 login node or 21

docs/tailscale.md
dasbatta: Noted. Let me update.

docs/tailscale.md
```diff
@@ -0,0 +1,155 @@
...
+      "name":"tailscale",
+      "image":"ghcr.io/tailscale/tailscale:latest",
+      "imagePullPolicy":"Always",
+      "restartPolicy":"Always",
+      "securityContext":{"privileged":true},
+      "env":[
+        {"name":"POD_NAME","valueFrom":{"fieldRef":{"fieldPath":"metadata.name"}}},
+        {"name":"POD_UID","valueFrom":{"fieldRef":{"fieldPath":"metadata.uid"}}},
+        {"name":"TS_DEBUG_FIREWALL_MODE","value":"auto"},
+        {"name":"TS_KUBE_SECRET","valueFrom":{"fieldRef":{"fieldPath":"metadata.name"}}},
```
Uburro, line 81: At which step was this secret created? Maybe we should explicitly mention that the Secret is created automatically by Tailscale. What do you think?

dasbatta: The Kubernetes Secret referenced by `TS_KUBE_SECRET` is created/updated automatically by the Tailscale container to persist its state outside of Container/pod. Let me add a note There as well

dasbatta: Added a note @Uburro 
> Note: The Kubernetes Secret referenced by `TS_KUBE_SECRET` is created/updated automatically by the Tailscale container to persist its state.
> In this setup `TS_KUBE_SECRET` is set to the pod name (`metadata.name`), so each login pod typically uses a Secret with the same name as the pod.
> Deleting that Secret forces the pod to re-authenticate (a new auth URL will appear in logs)

PR #2192, opened 2026-02-20
SCHED-987: Delete dynamic workers from code

Uburro:
## Problem
Dynamic workers were part of the codebase but are no longer supported or needed. Their presence added unnecessary complexity and maintenance burden without providing value to current users.
## Solution
Removed all code related to dynamic workers, including their types, controllers, and any associated configuration or reconciliation logic. This simplifies the codebase and reduces the surface area for potential issues.
## Testing
Verified that the existing functionality works correctly without dynamic workers by deploying the operator and confirming that standard worker provisioning is unaffected. No regressions observed.
## Release Notes
Breaking: Dynamic workers support has been removed. If your cluster configuration references dynamic workers, please migrate to the standard worker configuration before upgrading.

### DIFF
--- a/internal/controller/clustercontroller/login.go
+++ b/internal/controller/clustercontroller/login.go
@@ -199,7 +199,6 @@ func (r SlurmClusterReconciler) ReconcileLogin(
 						&clusterValues.Secrets,
 						clusterValues.VolumeSources,
 						&clusterValues.NodeLogin,
-						true,
 					)
 					if err != nil {
 						stepLogger.Error(err, "Failed to render")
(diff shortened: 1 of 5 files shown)


internal/controller/clustercontroller/login.go
```diff
@@ -202,7 +199,7 @@ func (r SlurmClusterReconciler) ReconcileLogin(
 						&clusterValues.Secrets,
 						clusterValues.VolumeSources,
 						&clusterValues.NodeLogin,
-						nodeSetsEnabled,
+						true,
```
itechdima: mb remove this argument at all from the `ReconcileLogin`?

PR #2201, opened 2026-02-24
SCHED-1008: Fix IB topology for GPU nodes

Uburro:
## Problem
When a worker node is registered in Slurm via dynamic topology, scontrol show node reports only one switch level:
Topology=default:root:leaf
However, the Kubernetes node has two topology tiers in its labels:
```
topology.nebius.com/tier-1=leaf
topology.nebius.com/tier-2=spine
```
This means the spine switch (tier-2) was silently dropped, leaving Slurm unaware of the full network hierarchy and unable to optimize job placement across the IB fabric correctly.

## Solution
Fixed _format_tier_topology in worker_init.py to include all topology tiers in the Slurm switch path instead of only the lowest tier.

## Testing

1. All unit tests passed
2. Manually 

## Release Notes
Fix: Worker nodes now register the full switch hierarchy (spine → leaf) in Slurm dynamic topology, enabling correct topology-aware scheduling across multi-level IB fabrics. No configuration changes required.

### DIFF
--- a/images/worker/worker_init.py
+++ b/images/worker/worker_init.py
@@ -204,15 +204,19 @@ def format_slurm_topology(topology: str) -> str:
     Format topology string for Slurm --conf option.
 
     Input formats:
-      - JSON: '{"tier-1":"switch1","tier-2":"rack1"}' -> "topology=default:root:switch1"
-        (uses lowest tier number as the leaf switch/block)
+      - JSON: '{"tier-1":"switch1","tier-2":"rack1"}' -> "topology=default:root:rack1:switch1"
+        (builds full switch hierarchy: highest tier first, leaf last)
       - "default:switch1" -> "topology=default:root:switch1"
       - "default:sw_root:s1:s2" -> "topology=default:sw_root:s1:s2" (intermediate switches already present)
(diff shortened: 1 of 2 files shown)


images/worker/worker_init.py
```diff
@@ -283,10 +290,10 @@ def _format_tier_topology(parts: dict) -> str:
...
-        lowest_tier_key = tier_keys[0][1]
-        leaf_switch = parts[lowest_tier_key]
-        return f"topology=default:root:{leaf_switch}"
+        # Sort descending: highest tier first (spine/root-side), lowest last (leaf)
+        tier_keys.sort(key=lambda x: x[0], reverse=True)
+        switches = [parts[k] for _, k in tier_keys]
+        return f"topology=default:root:{':'.join(switches)}"
 
     if parts:
         first_value = next(iter(parts.values()))
```
rdjjke, line 299: Do I understand it correctly, that this case isn't fixed?

Uburro: will be delete in nexy main release

rdjjke: We decided to remove it completely, but in main, not in 3.0.1

PR #2215, opened 2026-02-25
SCHED-1015: add retries values for all Helm Releases

Uburro:
## Problem
install and upgrade sections in all FluxCD HelmRelease templates were hardcoded, making it impossible to tune CRD handling strategy or remediation retries without modifying templates directly.

## Solution
Extracted install and upgrade blocks into values.yaml for all HelmRelease resources in soperator-fluxcd. Templates now render these sections via toYaml.

## Testing
helm template ./helm/soperator-fluxcd  # renders without errors
helm unittest ./helm/soperator-fluxcd   # all tests pass

## Release Notes
Feature: install and upgrade settings (CRD strategy, remediation retries) for all FluxCD HelmReleases are now configurable via values.yaml. No breaking changes.

### DIFF
--- a/helm/soperator-fluxcd/values.yaml
+++ b/helm/soperator-fluxcd/values.yaml
@@ -133,7 +133,7 @@ mariadbOperator:
       remediation:
         retries: 3
     upgrade:
-    crds: CreateReplace
+      crds: CreateReplace
     remediation:
       retries: 3
       remediateLastFailure: true


helm/soperator-fluxcd/values.yaml
```diff
@@ -93,7 +118,25 @@ mariadbOperator:
...
+    remediation:
+      retries: 3
+      remediateLastFailure: true
   crds:
+    install:
+      crds: CreateReplace
+      remediation:
+        retries: 3
+    upgrade:
+    crds: CreateReplace
```
theyoprst: wrong indent

Uburro: thanks

PR #2295, opened 2026-03-10
SCHED-1024 SSSD for LDAP integration

asteny:
## Problem

We needed to add full SSSD support to slurm-operator to enable LDAP-backed user resolution and SSH key lookup. Before this change, the operator did not provide a consistent way to run SSSD as a sidecar, there was no complete model for managing sssd.conf and LDAP CA inputs, and the controller could not resolve LDAP users required by Slurm.

## Solution

This PR adds optional SSSD sidecar support for login, worker, and controller pods. It introduces shared API and values fields for SSSD sidecar configuration, sssd.conf Secret references, LDAP CA ConfigMap references, and sssdDebugLevel. It also adds default sssd.conf Secret generation when no external Secret is provided, mounts LDAP CA from ConfigMap without triggering automatic pod rollouts, and includes a proper SSSD bootstrap command for reliable startup. SSH AuthorizedKeysCommand is now rendered only when SSSD is enabled, and Helm charts were updated so SSSD can be configured from a single place and remains disabled by default.


internal/render/common/container.go
```diff
@@ -41,6 +41,14 @@ func RenderContainerMunge(container *values.Container, opts ...RenderOption) cor
 		AppArmorProfile: ParseAppArmorProfile(container.AppArmorProfile),
 	}
 
+	volumeMounts := []corev1.VolumeMount{
```
ali-sattari: This is not used below in returned `corev1.Container`. Is it redundant?

images/controller/slurmctld_entrypoint.sh
```diff
@@ -13,6 +13,9 @@ echo "Bind-mount REST JWT secret key from K8S secret"
 touch /var/spool/slurmctld/jwt_hs256.key
 mount --bind /mnt/rest-jwt-key/rest_jwt.key /var/spool/slurmctld/jwt_hs256.key
 
+echo "Bind mount pipes for sssd"
+mount --bind /var/lib/sss/pipes var/lib/sss/pipes
```
theyoprst, line 17: It should be a slash before second `var`, isn't it?

rdjjke: I don't understand this line as well.

It mounts `/var/lib/sss/pipes` to `./var/lib/sss/pipes` in the working directory.
But what is the working directory?

images/common/scripts/complement_jail.sh
```diff
@@ -49,6 +49,15 @@ pushd "${jaildir}"
     echo "Bind-mount /etc/hosts"
     mount --bind /etc/hosts etc/hosts
 
+    echo "Bind mount pipes for sssd"
+    mount --bind /var/lib/sss/pipes var/lib/sss/pipes
+
+    echo "Bind mount sssd.conf if exists"
+    if [[ -f /etc/sssd/sssd.conf ]]; then
+      touch etc/sssd/sssd.conf
+      mount --bind /etc/sssd/sssd.conf etc/sssd/sssd.conf
```
rdjjke: Can we bind-mount the parent directory (`/etc/sssd`)? So that after `sssd.conf` file changes, there's no need to restart pods and re-run `complement_jail.sh`

images/controller/slurmctld_entrypoint.sh
asteny: removed

images/common/scripts/complement_jail.sh
asteny: ok

PR #2344, opened 2026-03-20
SCHED-957: docker http proxy

faucct:
## Problem

When user runs docker containers using `srun --mem 1G docker run`, the containers don't inherit the job limits and may crash other jobs on the node.

## Solution

I implicitly proxy Slurm Jobs's cgroups to Docker Container's `.HostConfig.CgroupParent` as a default. This is being done by pointing the Docker client to our HTTP Proxy via `DOCKER_HOST`, which copies the default from a `Cgroup-Parent` HTTP header, supplied via `DOCKER_CUSTOM_HEADERS`.
https://nebius.atlassian.net/wiki/spaces/SCHED/pages/1518403683/RFC+020+implicitly+set+docker+run+--cgroup-parent+in+srun
Right now the proxy is vibe-coded in golang, but this ended up being too large, so I plan to rewrite it in a smaller NGinx config.
The proxy also runs in the `slurmd` container and I will probably move it into a separate one.

## Testing

E2E tests and manual testing that setting Slurm memory limits makes the container to hit them as well.

## Release Notes

Feature: Docker containers now inherit the Cgroup of Slurm Jobs to respect the limits.

### DIFF
--- a/images/worker/nginx/docker_proxy.js
+++ b/images/worker/nginx/docker_proxy.js
@@ -42,7 +42,7 @@ function buildRequestBody(body, cgroupParent) {
 function handleCreate(r) {
     const cgroupParent = normalizeHeaderValue(r.headersIn['Cgroup-Parent']);
     if (!cgroupParent) {
-        r.internalRedirect('@docker_upstream');
+        r.internalRedirect('/_docker_upstream');
         return;
     }
 
@@ -50,12 +50,12 @@ function handleCreate(r) {
     try {
         updatedBody = buildRequestBody(r.requestText, cgroupParent);
     } catch (e) {
-        r.internalRedirect('@docker_upstream');
+        r.internalRedirect('/_docker_upstream');
         return;
     }
 
     if (updatedBody === null) {
-        r.internalRedirect('@docker_upstream');
+        r.internalRedirect('/_docker_upstream');
         return;
     }
 
--- a/images/worker/nginx/soperator-docker-proxy.conf
+++ b/images/worker/nginx/soperator-docker-proxy.conf
@@ -41,7 +41,7 @@ http {
             internal;
             proxy_http_version 1.1;
             proxy_buffering off;
-            proxy_request_buffering on;
+            proxy_request_buffering off;
             proxy_read_timeout 1d;
             proxy_send_timeout 1d;
             proxy_set_header Host docker;
@@ -50,22 +50,10 @@ http {
             proxy_pass http://docker_daemon$docker_request_uri;
         }
 
-        location @docker_upstream {
-            proxy_http_version 1.1;
-            proxy_buffering off;
-            proxy_request_buffering on;
-            proxy_read_timeout 1d;
-            proxy_send_timeout 1d;
-            proxy_set_header Host docker;
-            proxy_set_header Upgrade $http_upgrade;
-            proxy_set_header Connection $connection_upgrade;
-            proxy_pass http://docker_daemon;
-        }
-
         location / {
             proxy_http_version 1.1;
             proxy_buffering off;
-            proxy_request_buffering on;
+            proxy_request_buffering off;
             proxy_read_timeout 1d;
             proxy_send_timeout 1d;
             proxy_set_header Host docker;


helm/soperator-custom-configmaps/config-files/supervisord.conf
```diff
@@ -51,3 +51,19 @@ stopasgroup=true ; Send SIGTERM to all child processes of supervisord
 killasgroup=true ; Send SIGKILL to all child processes of supervisord
 stopsignal=SIGTERM ; Signal to send to the program to stop it
 stopwaitsecs=10 ; Wait for the process to stop before sending a SIGKILL
+
+[program:soperator-docker-proxy]
```
Uburro: could  we add this proxy in sidecar, not the same container?

faucct: Will do next week.

docker-proxy/main.go
faucct, line 1: Will be rewritten to a shorter NGINX config.

.github/workflows/e2e_test.yml
```diff
@@ -511,7 +506,7 @@ jobs:
             echo "No running sconfigcontroller pod found, skipping jail files collection"
           fi
 
-      - name: Zip Jail
+      - name: Tar Jail
```
asteny: Do you want to tar all Jail? 20+ GB? If not write here plz a description

asteny: DO you want to tar 20+ Gib? If not, plz write a description here

faucct: Is not relevant and already merged:
https://github.com/nebius/soperator/pull/2358

images/worker/slurmd.dockerfile
```diff
@@ -2,6 +2,13 @@
 
 ARG SLURM_VERSION
 
+FROM golang:1.26 AS docker-proxy
```
itechdima: in the docker-proxy go.mod
`go 1.22.2`
and here `golang:1.26`
and in Soperator we have even `1.25` :) 

Let's have `1.25` everywhere (I know there's a plans to rewrite it to nginx, but to not bring inconsistency(

faucct: I will just install the nginx to this image, so this won't be here. The changes are here, as I am waiting for E2E:
https://github.com/nebius/soperator/compare/feature/docker-http-proxy...feature/docker-http-proxy-wip

helm/soperator-custom-configmaps/config-files/supervisord.conf
faucct: Done.

docker-proxy/main.go
faucct: Done.

images/worker/slurmd.dockerfile
faucct: Done.

helm/slurm-cluster/slurm_scripts/task_prolog.sh
```diff
@@ -22,3 +33,12 @@ fi
 if [ -v SLURM_JOB_ID ]; then
     echo export "SLURM_JOB_ID_SOPERATOR_TASK_PROLOG=${SLURM_JOB_ID:-}"
 fi
+
+if ! /usr/bin/python3 -c "import sys; sys.exit(0)" >/dev/null 2>&1; then
```
Uburro: Why do we need this check? And why we exit with 0 if failed?

faucct: `exit 0` is the way it is done in other prologs, but this check is indeed not needed anymore, so I will delete it.

internal/render/worker/container.go
```diff
@@ -267,6 +268,20 @@ func renderContainerNodeSetSlurmd(
...
+		Image:           nodeSet.ContainerSlurmd.Image,
+		ImagePullPolicy: nodeSet.ContainerSlurmd.ImagePullPolicy,
+		Command:         []string{"/opt/bin/slurm/docker_proxy_nginx_entrypoint.sh"},
+		VolumeMounts: []corev1.VolumeMount{
+			renderVolumeMountRuntime(),
+		},
+		TerminationMessagePath:   corev1.TerminationMessagePathDefault,
+		TerminationMessagePolicy: corev1.TerminationMessageReadFile,
+	}
+}
```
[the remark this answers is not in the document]
faucct: It works, because the `/var/run` is symlinked to `/run` in the used docker image. I would rather keep it like this instead of having different `/var/run` and `/run` or mounting twice.

internal/render/worker/volume.go
```diff
@@ -101,6 +102,15 @@ func renderSupervisordConfigMap(name string) corev1.Volume {
 	}
 }
 
+func renderVolumeRuntime() corev1.Volume {
+	return corev1.Volume{
+		Name: consts.VolumeNameRuntime,
+		VolumeSource: corev1.VolumeSource{
+			EmptyDir: &corev1.EmptyDirVolumeSource{},
```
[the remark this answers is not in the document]
faucct: By default the containers `/run` is overlayfs backed by host disk, so this is fine I guess.

helm/slurm-cluster/slurm_scripts/task_prolog.sh
```diff
@@ -5,6 +5,17 @@ set -euxo pipefail
...
+    cgroup=$(cat /proc/self/cgroup)
+    cgroup="${cgroup#0::}"
+    # drop basename until it is user
+    while [ -n "$cgroup" ] && [ "${cgroup##*/}" != "user" ]; do
+    cgroup="${cgroup%/*}"
+    done
+    if [ -n "$cgroup" ]; then
+        echo export "DOCKER_HOST=unix:///var/run/soperator-docker.sock"
+        echo export "DOCKER_CUSTOM_HEADERS=Cgroup-Parent=$cgroup"
+    fi
```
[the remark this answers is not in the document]
faucct: We only have cgroup V2.

Uburro (approved):

PR #2316, opened 2026-03-13
SCHED-795, SCHED-797: TaskProlog should check for recursive CPU bindings

faucct:
## Problem

When accidentally running `sbatch` inside an interactive session, it is ignoring its `--cpus-per-task` setting:

```
cat > /sleep.sbatch <<BASH
#!/bin/bash
#SBATCH --cpus-per-task=10

sleep 100
BASH

srun --cpus-per-task=1 --time=30:00 --pty /bin/bash
```

```
root@worker-1:~# sbatch /sleep.sbatch 
Submitted batch job 273
root@worker-1:~# ssh worker-0
Welcome to Soperator cluster 

You are on node worker-0 (Ubuntu 24.04.3 LTS x86_64) 

System information as of Fri Mar 13 10:24:44 UTC 2026:
  CPU load: 0.1%	Memory usage: 4.5%	Processes: 18
  File system usage:
     Size Use% FSType   Directory
     512G   3% virtiofs /
   494.9G   3% ext4     |-/tmp
      91G   0% nfs4     |-/home
      55G   0% tmpfs    |-/mnt/memory
     512G   0% virtiofs |-/mnt/data
  1006.9G   0% ext4     |-/mnt/local-data

Slurm nodes:
  PARTITION   CPUS  MEMORY   GRES                                NODES  NODELIST                  STATE       REASON                                            
  main        16    56320    (null)                              2      worker-[0-1]              mixed       none                                              

Job queue:
  JOBID       PARTITION   NAME                    USER            STATE     TIME     NODES   NODELIST(REASON)                                  
  273         main        sleep.sbatch            root            RUNNING   0:21     1       worker-0                                          
  272         main        bash                    root            RUNNING   0:24     1       worker-1                                          

No other users are currently logged in 
Last login: Fri Mar 13 10:22:28 2026 from 10.22.52.78
root@worker-0:~# pgrep sleep
455344
root@worker-0:~# cat /proc/455344/environ
SHELL=/bin/bashSLURM_STEP_NUM_TASKS=2OMPI_MCA_orte_precondition_transports=0000011000000000-0000011000000000SLURM_JOB_USER=rootSLURM_TASKS_PER_NODE=1SLURM_JOB_UID=0SLURM_TIME_FORMAT=%Y-%m-%d %H:%M:%S.%ZSLURM_CPU_BIND=verbose,mask_cpu:0x0001,0x0002
```

## Testing
Created a test cluster with the mentioned config. Now I can see the warning in `.out`:

```
[...]

PR #2350, opened 2026-03-23
Fix E2E for 2316

faucct:
https://github.com/nebius/soperator/pull/2316

### DIFF
--- a/images/common/scripts/complement_jail.sh
+++ b/images/common/scripts/complement_jail.sh
@@ -135,6 +135,8 @@ pushd "${jaildir}"
 
     echo "Bind-mount /etc/enroot, /usr/share/enroot and /usr/lib/enroot"
     mkdir -p etc/enroot usr/share/enroot usr/lib/enroot
+    mkdir -p /etc/enroot/mounts.d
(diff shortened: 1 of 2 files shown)


images/common/scripts/complement_jail.sh
```diff
@@ -135,6 +135,8 @@ pushd "${jaildir}"
 
     echo "Bind-mount /etc/enroot, /usr/share/enroot and /usr/lib/enroot"
     mkdir -p etc/enroot usr/share/enroot usr/lib/enroot
+    mkdir -p /etc/enroot/mounts.d
+    echo '/opt/slurm_scripts/task_prolog.sh' | sudo tee /etc/enroot/mounts.d/task_prolog.fstab
```
itechdima, line 139: `taskProlog` option in set to Optional in SlurmCluster Custom Resource
Will it work if we don't have any task prolog? (just to double check)

faucct: The code looks like the file will always be there, just won't be used by slurm, so this code is fine.

faucct:
The E2E have passed except `Add build info to job summary`, but this feels like some unrelated flap.

PR #2408, opened 2026-04-09
SCHED-1397: Skip GPU active checks on CPU-only clusters

theyoprst:
## Problem

GPU-oriented active checks (`cuda-samples`, `dcgmi-diag-*`, `nccl-*`, `gpu-fryer`, etc.) fail on CPU-only clusters because sbatch cannot satisfy the GPU gres request. The check is reported as Failed, the `wait-for-active-checks` Helm hook loops on the error, and `terraform apply` times out.

Additionally, `docker-memory-limit-oomkilled` fails on CPU nodes because the docker daemon defaults to the NVIDIA Container Runtime which can't initialize NVML. And two acceptance test scenarios unconditionally run `nvidia-smi`, which doesn't exist on CPU workers.

### Active Checks — auto-detect GPU from sbatch scripts

- `slurm_check_job_entrypoint.sh` now **auto-detects GPU requirements** by grepping the sbatch script (`/opt/bin/sbatch.sh`) for `#SBATCH --gpus-per-node`, `--gpus`, `--gres=gpu`, or `-G` directives. If found and `/etc/slurm/slurm.conf` has no `Gres=gpu` token (CPU-only cluster), the entrypoint annotates the owning K8s Job with `slurm-skipped-reason` and exits 0. No CRD field or env var plumbing needed — the detection is purely at the entrypoint level.
- `ActiveCheckJobReconciler` picks up the annotation (after the K8s Job reaches terminal state), writes `LastRunStatus=Skipped`, and marks the Job final so reconciles short-circuit.
- `ActiveCheckReconciler.dependenciesReady` treats `Skipped` as a satisfied prerequisite so downstream checks still run.
- `wait-for-checks-job` recognises `Skipped` as a terminal success state.
- `docker-memory-limit-oomkilled.sh` now uses `--runtime=runc` to bypass the NVIDIA Container Runtime, making it work on both CPU and GPU nodes (the test validates Slurm memory limits, not GPU).

### Acceptance Tests

- GPU-specific scenarios (`package_installation`, `node_replacement`) are tagged `@gpu` in the feature files.
- `discoverCluster` now queries `sinfo -hN -p main -o '%N|%G'` to detect GPU workers at discovery time (`WorkerRef.HasGPU`).
- The godog runner excludes `@gpu` scenarios via `Tags: "~@gpu"` when no GPU workers are found (CPU-only clusters).
- GPU-specific steps use `AnyGPUWorker()` instead of `AnyWorker()`, so on mixed clusters they target GPU workers only.

## Testing

- Unit suite `internal/controller/soperatorchecks/activecheck_skipped_test.go` covers the skipped annotation path, idempotency, and the dependency resolver.
- GPU e2e: https://github.com/nebius/soperator/actions/runs/24192921815
- CPU e2e: https://github.com/nebius/soperator/actions/runs/24245920893
- Mixed e2e: https://github.com/nebius/soperator/actions/runs/24247372394

## Release Notes

Feature: GPU-only active checks are now automatically skipped on CPU-only clusters. The entrypoint detects GPU requirements from sbatch script directives — no per-check configuration needed.


api/v1alpha1/activecheck_types.go
```diff
@@ -110,6 +110,15 @@ type ActiveCheckSpec struct {
...
 	FailureReactions *Reactions `json:"failureReactions,omitempty"`
+
+	// RequiresGPU declares that this check needs GPU workers. When true, the
+	// slurm_check_job entrypoint inspects /etc/slurm/slurm.conf and exits
+	// early (marking the run Skipped) if no GPU gres is available, without
+	// attempting to submit a slurm job. Only the slurmJob path honors this
+	// flag today.
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=false
+	RequiresGPU bool `json:"requiresGpu,omitempty"`
```
ChessProfessor: Maybe let's move it to SlurmJobSpec? Or do you think we can somehow use it for k8s checks later too?

helm/soperator-activechecks/templates/wait-for-checks-job.yaml
```diff
@@ -66,6 +66,8 @@ spec:
 
                     if [ "$status" = "Complete" ]; then
                       echo "$name succeeded."
+                    elif [ "$status" = "Skipped" ]; then
+                      echo "$name skipped (no GPU on this cluster)."
```
ChessProfessor: Not sure we should specify reason (no gpu) here

helm/soperator-activechecks/values.yaml
```diff
@@ -180,6 +186,7 @@ checks:
       sbatchScriptFile: "scripts/ensure-healthy-nodes.sh"
   extensive-check:
     enabled: true
+    requiresGpu: true
```
ChessProfessor: That means extensive check will always be skioped for cpu only clusters and node may stuck in reservation...
Not sure what to do, extensive check flow probably shouldn't be enabled for cpu-only clusters at all, but for now it is...

images/slurm_check_job/slurm_check_job_entrypoint.sh
ChessProfessor, line 86: Is it hard to add parsing of the list of gpu nodes from config and launch only on them inside this script?
Because currently we just try to launch on every node and ignore all errors.

theyoprst: Thanks for catching this. A couple of clarifications:

- `slurm_submit_jobs.sh` already handles per-node sbatch failures — see lines 69–79. A CPU node that rejects `--gpus-per-node=8` is logged and skipped; the loop continues to the next node. Not quite "ignore all errors".
- The original ticket (SCHED-1397) targets CPU-only clusters, not mixed ones. On CPU-only the entrypoint gate short-circuits before `slurm_submit_jobs.sh` even runs. On mixed clusters (GPU+CPU workers) the gate intentionally does not fire (`Gres=gpu` is present in `slurm.conf`), and todays behavior (try every node, per-node errors for CPU workers, GPU workers succeed) works.
- GPU-node filtering is a nice optimization for mixed clusters (reduces log noise and wasted sbatch calls). I will open a follow-up ticket for it rather than expand scope here.

ChessProfessor: Yes, let's create a ticket to add that too, because for now we do not notice any sbatch submission errors since they are just logged in the job logs and ignored then

internal/controller/soperatorchecks/slurm_nodes_controller.go
```diff
@@ -301,6 +301,15 @@ func (c *SlurmNodesController) processHealthCheckFailed(
 
 	logger.V(1).Info("Creating a slurm reservation for drained node with [HC] reason")
 
+	// TODO(SCHED-1397): on CPU-only clusters extensive-check is Skipped
```
ChessProfessor: I also created this one not to forget to plan it - https://nebius.atlassian.net/browse/SCHED-1417

ChessProfessor (dismissed):

images/slurm_check_job/slurm_check_job_entrypoint.sh
theyoprst: https://nebius.atlassian.net/browse/SCHED-1421

internal/controller/soperatorchecks/slurm_nodes_controller.go
theyoprst: Updated in the code

helm/soperator-activechecks/values.yaml
theyoprst: https://nebius.atlassian.net/browse/SCHED-1417

images/slurm_check_job/slurm_check_job_entrypoint.sh
```diff
@@ -37,6 +37,26 @@ echo "Create directory for slurm job outputs"
...
+    if [[ -f /etc/slurm/slurm.conf ]] && ! grep -q '^[^#]*Gres=gpu' /etc/slurm/slurm.conf; then
+        SKIP_REASON="no GPU nodes in slurm.conf"
+        echo "$SKIP_REASON — marking check '$ACTIVE_CHECK_NAME' as Skipped"
+
+        K8S_JOB_NAME=$(kubectl get pod "$K8S_POD_NAME" -n "$K8S_POD_NAMESPACE" \
+            -o jsonpath='{.metadata.ownerReferences[?(@.kind=="Job")].name}')
+        if [[ -n "$K8S_JOB_NAME" ]]; then
+            kubectl annotate job "$K8S_JOB_NAME" \
+                -n "$K8S_POD_NAMESPACE" \
+                slurm-skipped-reason="$SKIP_REASON" --overwrite || true
```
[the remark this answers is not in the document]
theyoprst: Good catch — applied. Changed `|| true` to `|| { echo "..."; exit 1 }` so a failed annotation now fails the Job, consistent with the existing `slurm-job-id` annotation path below. Also added an explicit check for empty `K8S_JOB_NAME` (same pattern as the non-skip path).

internal/controller/soperatorchecks/activecheck_jobs_controller.go
```diff
@@ -160,6 +160,37 @@ func (r *ActiveCheckJobReconciler) Reconcile(
...
 
 	if activeCheck.Spec.CheckType == "slurmJob" {
+		// Skipped: the entrypoint detected the cluster has no GPU and exited
+		// early without calling sbatch. The owning Job is in Complete state.
+		if reason, ok := k8sJob.Annotations[consts.ActiveCheckSkippedReasonAnnotation]; ok {
+			// Idempotency: don't re-process if this Job is already marked final.
+			if k8sJob.Annotations[K8sAnnotationSoperatorChecksFinalStateTime] != "" {
+				return ctrl.Result{}, nil
+			}
+
```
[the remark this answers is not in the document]
theyoprst: Good point — applied. Added a `getK8sJobStatus` guard that requeues (10s) until the Job reaches `Complete` or `Failed` before processing the skip annotation. This prevents the microsecond race between the annotation write and the pod exit, and is consistent with the existing `Failed` branch which also gates on terminal status.

Uburro (approved):

PR #2421, opened 2026-04-10
SCHED-1402: Requeue when populate jail job exists but has not completed yet

theyoprst:
## Problem

When a populate-jail Job already exists (e.g. from a previous reconciliation cycle), `ReconcilePopulateJail` returned success without checking whether the Job had actually completed. This let the reconciler proceed to create login/worker pods while restic restore was still running — restic's `--delete` pass could then remove files that `complement_jail.sh` had already written, causing ~30% e2e failure rate (`dlopen(chroot.so): No such file or directory`).

## Solution

`ReconcilePopulateJail` now checks `Status.Succeeded` when the Job already exists:
- `Succeeded > 0` → return normally, reconciliation continues
- `Succeeded == 0` → return `notReady = true`, caller early-returns before `ReconcileLogin`/`ReconcileWorker` and requeues after 10s

This is intentionally separate from PR #2183's `populateJailRequeue` mechanism (maintenance overwrite mode), which must let the rest of the reconciliation run so that login/worker StatefulSets get scaled to zero.

## Testing

[E2E runs](https://github.com/nebius/soperator/actions/workflows/e2e_test.yml?query=branch%3ASCHED-1402%2Ffix-node-replacement-2)

## Release Notes

Fix: populate-jail reconciliation no longer skips incomplete jobs, preventing broken node replacements caused by a race between restic restore and login/worker pod creation.

### DIFF
--- a/internal/controller/clustercontroller/populate_job.go
+++ b/internal/controller/clustercontroller/populate_job.go
@@ -30,9 +30,10 @@ func (r SlurmClusterReconciler) ReconcilePopulateJail(
 	ctx context.Context,
 	clusterValues *values.SlurmCluster,
 	cluster *slurmv1.SlurmCluster,
-) (ctrl.Result, error) {
+) (res ctrl.Result, notReady bool, err error) {
 	logger := log.FromContext(ctx)
 	populateJailRequeue := false
+	populateJailNotReady := false
 
 	reconcilePopulateJailImpl := func() error {
 		return utils.ExecuteMultiStep(ctx,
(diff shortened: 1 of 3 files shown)


internal/controller/clustercontroller/populate_job.go
```diff
@@ -78,7 +79,12 @@ func (r SlurmClusterReconciler) ReconcilePopulateJail(
 							}
 							return nil
 						}
-						stepLogger.V(1).Info("Skipping creation: Populate jail Job already exists")
+						if desired.Status.Succeeded == 0 {
```
asteny, line 87: Could it work if there is Populate jail Job with Succeeded status, but you want to create new one for repopulate the jail? For example with `maintenance = "downscaleAndOverwritePopulateJail"`

asteny (approved):

internal/controller/clustercontroller/populate_job.go
theyoprst: Added support for this use case too, thanks

PR #2428, opened 2026-04-14
SCHED-1295 Cluster Creation Acceptance

ChessProfessor:
## Problem

Current cluster creation test is not fully covers scenario from confluence.

## Solution

Adding more checks to cluster creation test

## Testing

E2e run gpu - https://github.com/nebius/soperator/actions/runs/24400490397
cpu - https://github.com/nebius/soperator/actions/runs/24446945252

### DIFF
--- a/internal/e2e/acceptance/framework/exec.go
+++ b/internal/e2e/acceptance/framework/exec.go
@@ -6,8 +6,8 @@ import (
 )
 
 const (
-	DefaultRetryAttempts = 3
-	DefaultRetryDelay    = 5 * time.Second
+	DefaultRetryAttempts = 5
+	DefaultRetryDelay    = 10 * time.Second
 )
 
 type Exec interface {
--- a/internal/e2e/acceptance/framework/exec.go
+++ b/internal/e2e/acceptance/framework/exec.go
@@ -12,6 +12,7 @@ const (
 
 type Exec interface {
 	AnyWorker() (WorkerRef, error)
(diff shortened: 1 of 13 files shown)


theyoprst (commented):
couple of minor things:

1. typo in constant at `cluster_creation.go:27` - `clusterCreationPerNodeSmokeTimout` (missing `e` in Timeout)

2. `WorkerRef.HasGPU` (`model.go:5`) is never set - workers are created as `WorkerRef{Name: name}` in `discoverCluster`, and `classifyWorkers` populates `state.GPUWorkers` instead of setting the field. so the loop at `model.go:33-38` in `HasGPUWorkers()` is dead code (function works via the first `GPUWorkers` check, but the field can be removed)

3. minor: `runner.go:103-110` - after `line = strings.TrimSpace(line)`, the next line does `name := strings.TrimSpace(line)` - `name` is always equal to `line` at that point, second check redundant

internal/e2e/acceptance/framework/exec.go
```diff
@@ -6,20 +6,34 @@ import (
 )
 
 const (
-	DefaultJailRetryAttempts = 5
-	DefaultJailRetryDelay    = 10 * time.Second
+	DefaultRetryAttempts = 3
+	DefaultRetryDelay    = 5 * time.Second
```
theyoprst: I would keep 5 attempts with 10s timeout :-)

theyoprst (approved):

PR #2443, opened 2026-04-15
SCHED-1320 Make Internal SSH test up to date

ChessProfessor:
## Problem

1. Current internal ssh test is not fully compatible with the scenario from confluence.
2. There are no information about step/test duration in the logs

## Solution

1. Make it compatible
2. Add hooks to track execution time.

## Testing

E2e test - https://github.com/nebius/soperator/actions/runs/24453565524

### DIFF
--- a/internal/e2e/acceptance/features/internal_ssh.feature
+++ b/internal/e2e/acceptance/features/internal_ssh.feature
@@ -1,5 +1,6 @@
 Feature: Internal SSH
   Scenario: A regular user can SSH to a worker without extra options
     Given a regular user account exists on the login node
+    And the selected worker host key is not present for that user
     When the user SSHs from the login node to a worker
     Then the connection succeeds without extra SSH options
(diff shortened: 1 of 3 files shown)


theyoprst (commented):
lgtm. one nit - in `theUserSSHsFromTheLoginNodeToAWorker`, `worker` is already shell-quoted (output of `ShellQuote`), then the whole inner command is `ShellQuote`'d again. works correctly since `ShellQuote` escapes embedded single-quotes, but the worker name ends up double-quoted at the shell level. could simplify by using `s.targetWorker.Name` directly in the inner `fmt.Sprintf` and quoting only once at the outer `su -c` level. not a blocker.

theyoprst:
optional follow-up idea: godog has a `cucumber` formatter that outputs JSON with per-step `result.duration` (nanoseconds) - machine-readable timings for free. would be a one-liner in `runner.go`:

```go
Format: "pretty,cucumber:results.json",
```

to make it useful in CI you'd also need to upload the artifact - something like:

```yaml
- name: Upload acceptance test results
  if: always()
  uses: actions/upload-artifact@v7
  with:
    name: acceptance-results
    path: results.json
    retention-days: 7
```

not blocking, just easy to add if you want the data for trend analysis later.

theyoprst (approved):

PR #2446, opened 2026-04-15
SCHED-1299 Make package installation test up to date

ChessProfessor:
## Problem

Current scenario installs jq instead of nvitop

## Solution

Fixing that

## Testing

E2E


ChessProfessor:
Created a ticket to fix the problem here - https://nebius.atlassian.net/browse/SCHED-1498

PR #2447, opened 2026-04-15
SCHED-1302 Add RUN_UNSTABLE_TESTS flag

ChessProfessor:
## Problem

Not up to date node replacement test after PoC
It also flaps a lot, so we need to have a way to skip it

## Solution

Adding --gpus-per-node=8 to make it up to date with confluence scenario
Adding RUN_UNSTABLE_TEST flag to github action with defailt false

### DIFF
--- a/.github/workflows/e2e_test.yml
+++ b/.github/workflows/e2e_test.yml
@@ -18,6 +18,11 @@ on:
         required: false
         default: ""
         type: string
+      run_unstable_tests:
+        description: "Run acceptance scenarios tagged @unstable"
+        required: false
+        default: false
+        type: boolean
 
 permissions:
   contents: read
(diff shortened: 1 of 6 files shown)


.github/workflows/e2e_test.yml
```diff
@@ -18,6 +18,11 @@ on:
         required: false
         default: ""
         type: string
+      run_unstable_tests:
+        description: "Run acceptance scenarios tagged @unstable"
+        required: false
```
theyoprst, line 23: Better use strings: the boolean expressions with booleans may look weird here (not remember when, probably in case of negation or comparison of two bools

theyoprst (dismissed):

theyoprst (approved):

PR #2448, opened 2026-04-16
SCHED-1297 SCHED-1298 Docker container test

ChessProfessor:
## Problem

No docker and enroot container tests in acceptance yet

## Solution

Adding them

## Testing

E2e - https://github.com/nebius/soperator/actions/runs/24846459159


internal/e2e/acceptance/features/enroot_containers.feature
```diff
@@ -0,0 +1,17 @@
+Feature: Enroot containers
```
theyoprst, line 1: Maybe merge docker+enroot under one feature? Like `containers.feature`? Since they are both about containers support

ChessProfessor: I split these checks onto two PRs, we may consider it in the enroot one: https://github.com/nebius/soperator/pull/2473

internal/e2e/acceptance/steps/docker_containers.go
```diff
@@ -0,0 +1,221 @@
...
+
+	"github.com/cucumber/godog"
+
+	"nebius.ai/slurm-operator/internal/e2e/acceptance/framework"
+)
+
+const (
+	dockerContainersFeatureFile = "docker_containers.feature"
+
+	dockerImage = "cr.eu-north1.nebius.cloud/soperator/active_checks:12.9.0-ubuntu24.04-nccl_tests2.16.4-3935b93"
```
asteny: Is it a good design to hardcoded such thing as dockerImage, dockerARP and so on? Can we have such thing customizable?

internal/e2e/acceptance/steps/container_helpers.go
```diff
@@ -0,0 +1,139 @@
...
+	if waitTimeout <= 0 {
+		return nil
+	}
+	if err := waitForJobGone(ctx, exec, jobID, waitTimeout); err != nil {
+		return fmt.Errorf("wait for job %s to finish: %w", jobID, err)
+	}
+	return nil
+}
+
+func runWorkerCommand(ctx context.Context, exec framework.Exec, worker, command string) (string, error) {
```
asteny: dont get why you are doing something with ssh...

internal/e2e/acceptance/steps/package_installation.go
asteny, line 57: jq already installed https://github.com/nebius/ml-containers/blob/eb10e9bcf3825d5cf0b940faf44e6d74f09bfc69/ansible/roles/common-packages/defaults/main.yml#L10
So in this case you are doing nothing or upgrade it if there is new version

internal/e2e/acceptance/steps/docker_containers.go
ChessProfessor: I also don't like it, planned to consider changing it when tests are working, right now it's WiP
Do you have any suggestions how can we make it customizable and simple to use?

internal/e2e/acceptance/steps/container_helpers.go
ChessProfessor: I need to launch some of the commands from worker node (like sbatch), that's why I ssh to it from login and then execute the command. Do you know how to do that better?

internal/e2e/acceptance/steps/package_installation.go
ChessProfessor: Yep, we have a task to fix that - https://nebius.atlassian.net/browse/SCHED-1498
Current scenario setup with nvitop doesn't work, so I just switched to random package to make scenario work for now and created this bug. It should be nvitop eventually in this test
For now we can consider this scenario unstable

internal/e2e/acceptance/features/enroot_containers.feature
ChessProfessor: Enroot scenario is back to this PR because I managed to make it work :) 
Still I don't really like the idea of merging the features. For now we have very brittle split for scenarios and the source of truth is actually confluence page, so I want to make the tests split as close as possible to it, until we move source of truth to something more manageable than confluence. Let's maybe keep it as is for now?

internal/e2e/acceptance/steps/docker_containers.go
asteny: Maybe environment variables or config file

internal/e2e/acceptance/steps/container_helpers.go
asteny: Ok...
Maybe REST can be an option. Not sure that it supports all the options, but if we use it, that mean that we test it)

internal/e2e/acceptance/steps/docker_containers.go
theyoprst: > Do you have any suggestions how can we make it customizable and simple to use?

Don't we have some jobs in the cluster with such image? Maybe get it from there using kube api?

theyoprst (commented):
Few things, all minor:

1. **Stability — retries missing on slurm RPCs.** This runs every few minutes, so any transient slurmrestd hiccup flakes the run. No retry on:
   - `sbatch` submit (`docker_containers.go:81`, `enroot_containers.go:283`, `:341`)
   - `scancel` (`container_helpers.go:92`)
   - `runWorkerCommand` SSH path (`container_helpers.go:109`)

   `ExecJailWithDefaultRetry` / the `WithRetry` variant exists for a reason — use it consistently for any network-facing call. Only the cleanup `docker stop` / best-effort paths should skip retries (and even there, retry-once is safer).

2. **`waitForTreeEntriesOnWorker` is duplicated verbatim** in `docker_containers.go:155` and `enroot_containers.go:368` — differ only in the timeout constant. Hoist to `container_helpers.go` taking workers + timeout as args.

3. **`dockerContainersFromThatJobAreNoLongerRunning` (docker_containers.go:133) is tautological** — it does the cleanup itself and then waits for it. TODO acknowledges it, but as-is the scenario can't catch regressions of SCHED-1497 since it bypasses the thing it's supposed to assert. Worth making the workaround louder (step/scenario name, or a flag) so flipping back once the bug is fixed is mechanical.

4. **`stopContainersByNamePrefix` silently swallows errors** at docker_containers.go:214 (`continue`) and :218 (`_, _ = ...`). Cleanup shouldn't fail the test, but a `Logf` on failure will save hours of flake debugging later.

5. Nit: error-message casing is mixed — "cancel Docker job" (L176) vs "docker job id is empty" (L97, L157, L183). Per AGENTS.md "Docker" as a proper noun wants the uppercase — pick one.

Structure is solid otherwise — shared helpers in `container_helpers.go`, consistent job-lifecycle utilities, and the diagnostic dump in the enroot failure path is really nice for triaging flakes.

internal/e2e/acceptance/steps/docker_containers.go
ChessProfessor: For now I put it to envs, but we can try to get it from ActiveCheck CRs for example

ChessProfessor:
To @theyoprst review:

All done except for
  - No retry added for sbatch submit yet. I decided to avoid naive submit retries to prevent accidental duplicate jobs, I'm not sure right now how to retry safely. Added TODO comment though
  - No behavioral change to the SCHED-1497 workaround path yet, for now we do docker stop, later we will remove it.

theyoprst (dismissed):
Refactor looks good — `framework/slurm_client.go` and the scope-based `exec.Controller().Run()` / `exec.Worker(name).Run()` API are cleaner, and `WaitForTreeEntriesOnWorker` is now shared. Build/vet green.

Findings from last round: #2 (dedup), #4 (silent errors), #5 (casing) all fixed. #1 (retries) is mostly in — `scancel` + ssh paths retry now. `sbatch` retry is the interesting one. If you want to make it safe, a pattern that works: pick a deterministic job name per scenario attempt, and before sbatch do `squeue -h --name=<jobname> -o '%i'` — if non-empty, a prior attempt already succeeded and you adopt that job ID instead of re-submitting. That makes the sbatch step idempotent at the slurm layer. Not a blocker, just an option.

Two tiny nits:

1. `framework.SlurmPollInterval` is now used for non-slurm polling too (e.g., `WaitForTreeEntriesOnWorker`). The name reads wrong in those call sites — maybe `DefaultPollInterval` in `helpers.go`.
2. `docker_containers.go:136-141` — `theDockerNCCLJobIsCancelled`'s `if err != nil { return err }` wrapper is dead weight; collapse to `return s.cancelCurrentJob(ctx)`.

Fine to ship once you feel good about sbatch idempotency (or just leave the TODO and file a follow-up).

theyoprst (approved):

PR #2464, opened 2026-04-20
SCHED-1481: Configure NCCL Inspector Pre-Configuration plugin

dstaroff:
## Problem
Despite the plugin is installed into Jail and containers, it's not specified in the `plugstack.conf`.

## Solution
This PR introduces the configuration of the Sneaky Leprechaun plugin via CRD and `slurm-cluster` Helm chart the same way as for Pyxis and SNCCLD.

## Testing
Tested on dev cluster.

## Release Notes
Feature: Sneaky Leprechaun configuration via CRD and Helm chart.


api/v1/slurmcluster_types.go
```diff
@@ -231,6 +231,12 @@ type PlugStackConfig struct {
 	// +kubebuilder:default={ required: false, enabled: false, logLevel: "INFO", outputToFile: true, outputToStdOut: false, outputDirectory: "/opt/soperator-outputs/nccl_logs" }
 	NcclDebug PluginConfigNcclDebug `json:"ncclDebug,omitempty"`
 
+	// NcclInspectorPreConf represents the NCCL Inspector Pre-Configuration SPANK plugin configuration.
+	//
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default={ required: false, enabled: false, profilerPlugin: "/usr/lib/x86_64-linux-gnu/libnccl-profiler-inspector.so", dumpDir: "/opt/soperator-outputs/nccl_profiles/%j/%s", dumpVerbose: false, dumpThreadIntervalMicroseconds: 1000000 }
```
asteny, line 237: Is it possible make a symlink in deb package or after installation from /usr/lib/x86_64... to the /usr/lib/libnccl-profiler-inspector and do not configure that?

dstaroff: It's be possible. It just requires updating the whole pipeline from `slurm-deb-packages` -> `deb registry` -> `ml-containers` -> `soperator images` -> `here`

But it could be even SO file, built by users, having arbitrary name and placed under arbitrary directory.

For now I just want to configure it via TF.

Uburro (approved):

PR #2497, opened 2026-04-28
SCHED-1146: rebooter: reduce memory usage when checking pod eviction on large clusters

Uburro:
## Problem

On large clusters the rebooter's `AreAllPodsEvicted()` called `r.List()` against the controller-runtime informer cache, which loads and retains full `Pod` objects for every pod in the cluster regardless of the field selector - memory grows linearly with pod count.

## Solution

Query the Kubelet API through kube-apiserver instead of the informer cache.  The pod field indexer and pod watch are removed entirely, so the manager no longer caches pods at all. RBAC updated: `pods get/list/watch` → `nodes/proxy get`.

## Testing

Added regression coverage for stale node updates and verified the rebooter test package passes.

## Release Notes

Reduced rebooter memory usage on large clusters by replacing informer-cache pod listing with direct per-node Kubelet API queries.

### DIFF
--- a/internal/rebooter/apiserver_node_proxy.go
+++ b/internal/rebooter/apiserver_node_proxy.go
@@ -1,0 +1,51 @@
+package rebooter
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+)
+
+// NodePodsFetcher fetches the pods running on a specific node without going
+// through the controller-runtime informer cache.
+type NodePodsFetcher interface {
+	GetPodsOnNode(ctx context.Context, nodeName string) (*corev1.PodList, error)
+}
+
+// apiserverNodeProxy implements NodePodsFetcher via the API server's
+// nodes/proxy sub-resource. This avoids direct kubelet access entirely -
+// TLS and auth are handled by the API server using its own credentials,
+// independent of kubelet serving-cert configuration.
+type apiserverNodeProxy struct {
+	clientset kubernetes.Interface
+}
+
+func (p *apiserverNodeProxy) GetPodsOnNode(ctx context.Context, nodeName string) (*corev1.PodList, error) {
+	raw, err := p.clientset.CoreV1().RESTClient().Get().
+		Resource("nodes").
+		Name(nodeName).
+		SubResource("proxy").
+		Suffix("pods").
+		DoRaw(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("node proxy request failed for %s: %w", nodeName, err)
+	}
+	var podList corev1.PodList
+	if err := json.Unmarshal(raw, &podList); err != nil {
+		return nil, fmt.Errorf("decode proxy response for %s: %w", nodeName, err)
+	}
+	return &podList, nil
+}
+
+// NewAPIServerNodePodsFetcher creates a NodePodsFetcher that proxies through the API server.
+func NewAPIServerNodePodsFetcher(config *rest.Config) (NodePodsFetcher, error) {
+	clientset, err := kubernetes.NewForConfig(config)
+	if err != nil {
+		return nil, fmt.Errorf("create kubernetes client: %w", err)
+	}
+	return &apiserverNodeProxy{clientset: clientset}, nil
+}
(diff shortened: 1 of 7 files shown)


asteny (dismissed):

theyoprst (commented):
> Claude Code review (PR author requested), not a human review. Take with a grain of salt.

couple of things worth a look:

1. PR description says "stream-decoded, short-circuits on first blocking pod" but `kubelet_client.go:33` uses `DoRaw(ctx)` which `io.ReadAll`s the full body, then `json.Unmarshal`s into a `PodList`. memory win is real (one node's pods, GC'd quickly) but not via streaming. either fix the description or switch to `Stream(ctx)` + `json.NewDecoder` if streaming actually matters here.

2. `cmd/rebooter/main.go:117` lowers default `reconcile-timeout` from 3m to 1m. with the Node watch removed, this is now the ceiling on how fast the rebooter reacts to a new drain/reboot signal. not mentioned in the PR — was this intentional?

3. behavior delta worth flagging in release notes: the new code reads the kubelet's `/pods` (via apiserver proxy), the old code did API-server LIST by `spec.nodeName`. kubelet view excludes pods scheduled-but-not-yet-running and includes pods deleted at the API but not yet acked by kubelet. probably an improvement for drain semantics, but it's a real semantic shift.

4. test gap: `TestAreAllPodsEvicted` mocks `NodePodsFetcher` cleanly but nothing exercises the actual `apiserverNodeProxy` wire path. an `httptest.Server` returning a `PodList` JSON would lock down the wire format. acceptable for a memory-targeted refactor, just noting.

5. style nits per `CLAUDE.md` — error messages should start lowercase and skip "failed to". new error wraps in `reconcile.go:444` (`failed to patch object`), `:535` (`failed to set node condition for node`), `:671` (`failed to patch object status`). pre-existing in this file but the PR adds more.

the `RetryOnConflict` + `APIReader.Get`-inside-the-closure pattern for the Patch rewrites is clean. good call dropping the Node informer alongside the pod cache; the stale-node regression tests are meaningful.

internal/rebooter/apiserver_node_proxy.go
theyoprst, line 1: kubelet_client.go contains NodePodsFetcher interface and apiserverNodeProxy impl. Looks like it should be called in some other way?

Uburro: kubelet_client.go ->  apiserver_node_proxy.go
NewNodePodsFetcher -> NewAPIServerNodePodsFetcher is it ok ?

Uburro:
> > Claude Code review (PR author requested), not a human review. Take with a grain of salt.
> 
> couple of things worth a look:
> 
> 1. PR description says "stream-decoded, short-circuits on first blocking pod" but `kubelet_client.go:33` uses `DoRaw(ctx)` which `io.ReadAll`s the full body, then `json.Unmarshal`s into a `PodList`. memory win is real (one node's pods, GC'd quickly) but not via streaming. either fix the description or switch to `Stream(ctx)` + `json.NewDecoder` if streaming actually matters here.
> 2. `cmd/rebooter/main.go:117` lowers default `reconcile-timeout` from 3m to 1m. with the Node watch removed, this is now the ceiling on how fast the rebooter reacts to a new drain/reboot signal. not mentioned in the PR — was this intentional?
> 3. behavior delta worth flagging in release notes: the new code reads the kubelet's `/pods` (via apiserver proxy), the old code did API-server LIST by `spec.nodeName`. kubelet view excludes pods scheduled-but-not-yet-running and includes pods deleted at the API but not yet acked by kubelet. probably an improvement for drain semantics, but it's a real semantic shift.
> 4. test gap: `TestAreAllPodsEvicted` mocks `NodePodsFetcher` cleanly but nothing exercises the actual `apiserverNodeProxy` wire path. an `httptest.Server` returning a `PodList` JSON would lock down the wire format. acceptable for a memory-targeted refactor, just noting.
> 5. style nits per `CLAUDE.md` — error messages should start lowercase and skip "failed to". new error wraps in `reconcile.go:444` (`failed to patch object`), `:535` (`failed to set node condition for node`), `:671` (`failed to patch object status`). pre-existing in this file but the PR adds more.
> 


1. fixed
2. we have an reconcile-timeout in previous realisation 
3. it's ok
4. fixed
5. fixed

> the `RetryOnConflict` + `APIReader.Get`-inside-the-closure pattern for the Patch rewrites is clean. good call dropping the Node informer alongside the pod cache; the stale-node regression tests are meaningful.

glad to hear it

internal/rebooter/apiserver_node_proxy.go
Uburro: fixed

theyoprst (approved):

PR #2499, opened 2026-04-29
SCHED-1396: Reduce unnecessary Slurm API calls in pod_ephemeral_storage check low-usage path

Uburro:

### DIFF
--- a/cmd/rebooter/main.go
+++ b/cmd/rebooter/main.go
@@ -114,7 +114,7 @@ func main() {
 		"If set, HTTP/2 will be enabled for the metrics and webhook servers")
 	flag.StringVar(&logFormat, "log-format", "json", "Log format: plain or json")
 	flag.StringVar(&logLevel, "log-level", "debug", "Log level: debug, info, warn, error, dpanic, panic, fatal")
-	flag.DurationVar(&reconcileTimeout, "reconcile-timeout", 3*time.Minute, "The maximum duration allowed for a single reconcile if some error occurs")
+	flag.DurationVar(&reconcileTimeout, "reconcile-timeout", 1*time.Minute, "The maximum duration allowed for a single reconcile if some error occurs")
 	flag.DurationVar(&cacheSyncTimeout, "cache-sync-timeout", 2*time.Minute, "The maximum duration allowed for caching sync")
 	flag.Parse()
 
@@ -142,6 +142,11 @@ func main() {
 	tlsOpts := []func(*tls.Config){}
 	if !enableHTTP2 {
 		tlsOpts = append(tlsOpts, disableHTTP2)
+	}
+
+	nodeName := os.Getenv(consts.RebooterNodeNameEnv)
+	if nodeName == "" {
+		cli.Fail(setupLog, fmt.Errorf("%s environment variable is not set", consts.RebooterNodeNameEnv), "unable to determine rebooter node name")
 	}
 
 	webhookServer := webhook.NewServer(webhook.Options{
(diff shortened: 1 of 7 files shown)


theyoprst (commented):
optimization is reasonable and matches what the ticket asked for. tests pass locally including `-race`. couple of things before merge:

1. cache key is built three different ways. lookup uses `pod.Namespace + "/" + pod.Name` (`pod_ephemeral_storage_check.go:606`), drain uses `slurmClusterName.Namespace + "/" + slurmNodeName` (line 712), undrain same (line 663). they happen to agree only because `slurmNode.Name == pod.Name` by convention - if that ever drifts, drain writes one key and lookup reads another, optimization silently stops working. would extract a `cacheKey(namespace, name)` helper used in all three sites.

2. log capitalization fix is partial. you upper-cased lines 644 / 664 / 717 but missed `logger.Info("draining slurm node")` at line 688 and `logger.Info("slurm node", "nodeStates", ...)` at line 430. either capitalize all of them or drop the message changes from this PR.

3. PR body is empty (just template). copy a few lines from the ticket so the change is reviewable without jira.

nit: no direct test for the cache=`true` → fetch → undrain path or the new `Store(key, false)` branch at line 617. not a blocker.

theyoprst (commented):
missed this one in my first pass, adding now.

[P2] negative cache is set-and-stick - `pod_ephemeral_storage_check.go:607-608`

once a pod is stored as `false` the early return makes that conclusion permanent for the lifetime of the process. resume path breaks if the node later enters DRAIN with the `pod_ephemeral_storage` reason without a successful `Store(..., true)` here.

concrete failure mode:

1. low-usage reconcile - cache stores `false`
2. usage spikes, `handleHighStorageUsage` -> `drainSlurmNode`. slurm applies the drain but the client returns err (e.g. response timeout after slurm wrote the state). we return early, cache stays `false`
3. usage drops below resume threshold. `handleLowStorageUsage` short-circuits on the cached `false`, never re-reads slurm. node stays drained until the process restarts

same hole if any other actor (manual `scontrol update`, another controller) sets a drain reason starting with `<HC> pod_ephemeral_storage`.

few options, pick one:

- TTL on `false` entries (re-check after e.g. 5 min) - simplest, bounded staleness
- on `drainSlurmNode` error path, `Delete(key)` defensively so the next reconcile re-reads - covers the timeout case but not the external-actor case
- drop negative caching entirely and positive-cache only, which is what the ticket originally proposed (Option A: a set of nodes we drained). lookup becomes `if _, drained := Load(key); !drained { return nil }`. accepts the one-time-per-node restart cost the ticket explicitly called out

theyoprst (commented):
[design] before merging - do we need a per-pod cache here at all?

`slurmapi.Client.ListNodes` already exists and the sibling controller in this same package uses it (`slurm_nodes_controller.go:138`). one `ListNodes` per cluster gives you all node states at once, which is exactly what this check needs.

if `PodEphemeralStorageCheck` kept a shared `map[clusterNN]map[nodeName]Node` snapshot with a short TTL (~30s) refreshed via `ListNodes`, you'd get:

- 1 `ListNodes` per cluster per TTL instead of N `GetNode` per cycle, much lower load on `slurmctld`
- no `sync.Map`, no `Store(true/false)`/`Delete` bookkeeping
- no `false`-staleness bug from the previous comment (snapshot self-invalidates)
- no startup storm (single `ListNodes` per cluster, not N `GetNode`)
- high-usage path benefits too (currently it still pays full `GetNode` per reconcile)
- consistent with how the sibling controller in the same package already works

SCHED-1396 listed only `sync.Map` (Option A) and annotation (Option B), but the in-repo precedent is the snapshot approach. would propose redoing this PR that way, or scoping it as a follow-up if you need this change to land fast for the 2k PoC.

PR #2500, opened 2026-04-30
SCHED-1396: Reduce unnecessary Slurm API calls in pod_ephemeral_storage check low-usage path

Uburro:

### DIFF
--- a/internal/controller/soperatorchecks/pod_ephemeral_storage_check.go
+++ b/internal/controller/soperatorchecks/pod_ephemeral_storage_check.go
@@ -314,15 +314,19 @@ func (r *PodEphemeralStorageCheck) initSlurmClientAndGetNode(
 		Namespace: pod.Namespace,
 	}
 
-	if err := r.InitSlurmAPIClients(slurmClusterNamespacedName, slurmClusterName); err != nil {
+	if err := r.ensureNodeCache(ctx, slurmClusterNamespacedName, slurmClusterName); err != nil {
 		return types.NamespacedName{}, slurmapi.Node{}, fmt.Errorf("initializing Slurm API clients: %w", err)
 	}
 
-	slurmNode, err := r.getSlurmNode(ctx, slurmClusterNamespacedName, pod.Name)
-	if err != nil {
-		return types.NamespacedName{}, slurmapi.Node{}, fmt.Errorf("getting Slurm node: %w for pod %s/%s", err, pod.Namespace, pod.Name)
+	nc, found := r.slurmAPIClients.GetNodeCache(slurmClusterNamespacedName)
+	if !found {
+		return types.NamespacedName{}, slurmapi.Node{}, fmt.Errorf("node cache not found for cluster %v", slurmClusterNamespacedName)
 	}
 
+	slurmNode, found := nc.GetNode(pod.Name)
+	if !found {
+		return types.NamespacedName{}, slurmapi.Node{}, fmt.Errorf("slurm node %s not found in cache for cluster %v", pod.Name, slurmClusterNamespacedName)
+	}
 	return slurmClusterNamespacedName, slurmNode, nil
 }
 
@@ -561,40 +565,39 @@ func (r *PodEphemeralStorageCheck) getSlurmClusterName(ctx context.Context, name
 	return slurmClusterName, nil
 }
 
-// InitSlurmAPIClients initializes Slurm API clients for the given Slurm cluster
-func (r *PodEphemeralStorageCheck) InitSlurmAPIClients(
-	slurmClusterNamespacedName types.NamespacedName, slurmClusterName string) error {
-	if _, found := r.slurmAPIClients.GetClient(slurmClusterNamespacedName); found {
-		return nil // Client already exists, no need to create a new one
-	}
-
-	jwtToken := jwt.NewToken(r.Client).For(slurmClusterNamespacedName, "root").WithRegistry(jwt.NewTokenRegistry().Build())
-	slurmAPIServer := fmt.Sprintf("http://%s.%s:6820", naming.BuildServiceName(consts.ComponentTypeREST, slurmClusterName), slurmClusterNamespacedName.Namespace)
-	slurmAPIClient, err := slurmapi.NewClient(slurmAPIServer, jwtToken, slurmapi.DefaultHTTPClient())
-	if err != nil {
-		return fmt.Errorf("creating slurm api client: %w", err)
-	}
-	r.slurmAPIClients.AddClient(slurmClusterNamespacedName, slurmAPIClient)
-	return nil
-}
-
-func (c *PodEphemeralStorageCheck) getSlurmNode(
+// ensureNodeCache creates the Slurm API client and starts the node cache background
+// goroutine for the given cluster if they do not already exist, then blocks until
+// the first successful ListNodes refresh completes.
+func (r *PodEphemeralStorageCheck) ensureNodeCache(
 	ctx context.Context,
-	slurmClusterName types.NamespacedName,
-	slurmNodeName string,
-) (slurmapi.Node, error) {
-
-	slurmAPIClient, found := c.slurmAPIClients.GetClient(slurmClusterName)
-	if !found {
-		return slurmapi.Node{}, fmt.Errorf("slurm cluster %v not found", slurmClusterName)
+	slurmClusterNamespacedName types.NamespacedName,
+	slurmClusterName string,
+) error {
+	// Always wait for readiness, even when cache already exists — a concurrent
+	// reconcile may have just created it and the first refresh is still in flight.
+	if nc, found := r.slurmAPIClients.GetNodeCache(slurmClusterNamespacedName); found {
+		return nc.WaitReady(ctx)
 	}
 
-	node, err := slurmAPIClient.GetNode(ctx, slurmNodeName)
-	if err != nil {
-		return slurmapi.Node{}, fmt.Errorf("get node: %w", err)
+	if _, found := r.slurmAPIClients.GetClient(slurmClusterNamespacedName); !found {
+		jwtToken := jwt.NewToken(r.Client).For(slurmClusterNamespacedName, "root").WithRegistry(jwt.NewTokenRegistry().Build())
+		slurmAPIServer := fmt.Sprintf("http://%s.%s:6820", naming.BuildServiceName(consts.ComponentTypeREST, slurmClusterName), slurmClusterNamespacedName.Namespace)
+		slurmAPIClient, err := slurmapi.NewClient(slurmAPIServer, jwtToken, slurmapi.DefaultHTTPClient())
+		if err != nil {
+			return fmt.Errorf("creating slurm api client: %w", err)
+		}
+		r.slurmAPIClients.AddClient(slurmClusterNamespacedName, slurmAPIClient)
 	}
 
-	return node, nil
+	nc := r.slurmAPIClients.EnsureNodeCache(
+		slurmClusterNamespacedName,
+		r.reconcileTimeout,
+		log.Log.WithName("NodeCache").WithValues("cluster", slurmClusterNamespacedName),
+	)
+	if nc == nil {
+		return fmt.Errorf("no slurm API client for cluster %v", slurmClusterNamespacedName)
+	}
+	return nc.WaitReady(ctx)
 }
 
 func (r *PodEphemeralStorageCheck) handleLowStorageUsage(ctx context.Context, pod *corev1.Pod) error {
--- a/internal/slurmapi/node_cache.go
+++ b/internal/slurmapi/node_cache.go
@@ -1,0 +1,127 @@
+package slurmapi
+
+import (
+	"context"
+	"sync/atomic"
+	"time"
+
+	"github.com/go-logr/logr"
+)
+
+const (
+	nodeCacheRequestTimeout   = 2 * time.Minute
+	nodeCacheStalenessTimeout = 3 * time.Minute
+)
+
+// cacheSnapshot is an immutable point-in-time copy of the node list with its
+// capture timestamp. Stored and read via atomic.Pointer — no locks required.
+type cacheSnapshot struct {
+	nodes map[string]Node
+	at    time.Time
+}
+
+// NodeCache maintains a continuously-refreshed in-memory copy of all Slurm nodes
+// for a single cluster. A background goroutine calls ListNodes every refreshInterval
+// and stores the result atomically so readers never block. If the cache goes stale
+// (no successful refresh for nodeCacheStalenessTimeout), the snapshot is cleared so
+// callers receive proper "not found" errors instead of operating on outdated state.
+type NodeCache struct {
+	client          Client
+	logger          logr.Logger
+	refreshInterval time.Duration
+
+	snapshot atomic.Pointer[cacheSnapshot]
+
+	ready chan struct{}
+}
+
+// StartNodeCache starts a background goroutine that refreshes the node list every
+// refreshInterval. Use WaitReady to block until the first successful refresh.
+func StartNodeCache(ctx context.Context, client Client, refreshInterval time.Duration, logger logr.Logger) *NodeCache {
+	nc := &NodeCache{
+		client:          client,
+		logger:          logger,
+		refreshInterval: refreshInterval,
+		ready:           make(chan struct{}),
+	}
+	go nc.run(ctx)
+	return nc
+}
+
+// WaitReady blocks until the first successful ListNodes refresh completes or ctx is done.
+func (nc *NodeCache) WaitReady(ctx context.Context) error {
+	select {
+	case <-nc.ready:
+		return nil
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+}
+
+// GetNode returns the cached node with the given name, or false if not found.
+func (nc *NodeCache) GetNode(name string) (Node, bool) {
+	s := nc.snapshot.Load()
+	if s == nil {
+		return Node{}, false
+	}
+	node, ok := s.nodes[name]
+	return node, ok
+}
+
+func (nc *NodeCache) run(ctx context.Context) {
+	ticker := time.NewTicker(nc.refreshInterval)
+	defer ticker.Stop()
+
+	// Initial refresh loop: keep retrying until one refresh succeeds,
+	// then signal readiness and move to the steady-state loop.
+	for {
+		if nc.refresh(ctx) {
+			close(nc.ready)
+			break
+		}
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+		}
+	}
+
+	// Steady-state loop: refresh on every tick.
+	for {
+		select {
+		case <-ctx.Done():
+			return
+		case <-ticker.C:
+			nc.refresh(ctx)
+		}
+	}
+}
+
+// refresh performs one ListNodes call and updates the snapshot. Returns true on success.
+func (nc *NodeCache) refresh(ctx context.Context) bool {
+	if s := nc.snapshot.Load(); s != nil && time.Since(s.at) > nodeCacheStalenessTimeout {
+		nc.logger.Error(nil, "Slurm node cache is stale, clearing cached data",
+			"lastRefreshTime", s.at,
+			"elapsed", time.Since(s.at),
+		)
+		nc.snapshot.Store(nil)
+		// After clearing, s.at is gone — the stale check won't fire again until
+		// the next successful refresh followed by another full staleness period.
+	}
+
+	reqCtx, cancel := context.WithTimeout(ctx, nodeCacheRequestTimeout)
+	defer cancel()
+
+	nodes, err := nc.client.ListNodes(reqCtx)
+	if err != nil {
+		nc.logger.Error(err, "Failed to list Slurm nodes")
+		return false
+	}
+
+	nodeMap := make(map[string]Node, len(nodes))
+	for _, n := range nodes {
+		nodeMap[n.Name] = n
+	}
+	nc.snapshot.Store(&cacheSnapshot{nodes: nodeMap, at: time.Now()})
+	return true
+}
--- a/cmd/soperatorchecks/main.go
+++ b/cmd/soperatorchecks/main.go
@@ -17,7 +17,6 @@ limitations under the License.
 package main
 
 import (
-	"context"
 	"crypto/tls"
 	"errors"
 	"flag"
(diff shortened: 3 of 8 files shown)


internal/slurmapi/node_cache.go
```diff
@@ -0,0 +1,110 @@
...
+		case <-ticker.C:
+			nc.refresh(ctx)
+		}
+	}
+}
+
+func (nc *NodeCache) refresh(ctx context.Context) {
+	nc.mu.Lock()
+	lastRefresh := nc.lastRefreshTime
+	nc.mu.Unlock()
```
theyoprst: We don't need a mutex: `lastRefreshTime` is used and updated in a single goroutine, no need for synchronization, this part can be simplified.

internal/slurmapi/node_cache.go
```diff
@@ -0,0 +1,110 @@
...
+	for _, n := range nodes {
+		nodeMap[n.Name] = n
+	}
+	nc.nodes.Store(&nodeMap)
+
+	nc.mu.Lock()
+	nc.lastRefreshTime = time.Now()
+	nc.mu.Unlock()
+
+	nc.once.Do(func() { close(nc.ready) })
```
theyoprst: You can put `close(nc.ready)` after the first `nc.refresh(ctx)`, and remove `nc.once` completely.
Easier, same effect.

internal/slurmapi/node_cache.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+}
+
+func (nc *NodeCache) refresh(ctx context.Context) {
+	nc.mu.Lock()
+	lastRefresh := nc.lastRefreshTime
+	nc.mu.Unlock()
+
+	if !lastRefresh.IsZero() && time.Since(lastRefresh) > nodeCacheStalenessTimeout {
+		panic(fmt.Sprintf("slurm node cache is stale: last successful refresh was %v ago", time.Since(lastRefresh)))
```
theyoprst: Sorry, it was my bad idea to panic here. It will kill all controllers in this binary.
Let's rather log some message and store nil map in nodes, then ephemeral storage check will return proper errors

internal/slurmapi/node_cache.go
```diff
@@ -0,0 +1,110 @@
...
+	}
+
+	reqCtx, cancel := context.WithTimeout(ctx, nodeCacheRequestTimeout)
+	defer cancel()
+
+	nodes, err := nc.client.ListNodes(reqCtx)
+	if err != nil {
+		return
```
theyoprst: Can we pass logger here and log errors? It's not good that we lost them

internal/controller/soperatorchecks/pod_ephemeral_storage_check.go
```diff
@@ -314,15 +314,20 @@ func (r *PodEphemeralStorageCheck) initSlurmClientAndGetNode(
...
-		return types.NamespacedName{}, slurmapi.Node{}, fmt.Errorf("getting Slurm node: %w for pod %s/%s", err, pod.Namespace, pod.Name)
+	nc, found := r.slurmAPIClients.GetNodeCache(slurmClusterNamespacedName)
+	if !found {
+		return types.NamespacedName{}, slurmapi.Node{}, fmt.Errorf("node cache not found for cluster %v", slurmClusterNamespacedName)
 	}
 
+	slurmNode, found := nc.GetNode(pod.Name)
+	if !found {
+		log.FromContext(ctx).V(1).Info("Slurm node not found in cache, treating as not drained",
+			"pod", pod.Name, "cluster", slurmClusterNamespacedName)
```
theyoprst: Previously we returned error (404) if node not found, now we return Node{}, nil. Maybe keep returning error?

internal/controller/soperatorchecks/pod_ephemeral_storage_check.go
```diff
@@ -561,40 +566,35 @@ func (r *PodEphemeralStorageCheck) getSlurmClusterName(ctx context.Context, name
 	return slurmClusterName, nil
 }
 
-// InitSlurmAPIClients initializes Slurm API clients for the given Slurm cluster
+// InitSlurmAPIClients initializes the Slurm API client and node cache for the given
+// cluster, then blocks until the first ListNodes refresh completes so that callers
+// can immediately read node state from the cache.
 func (r *PodEphemeralStorageCheck) InitSlurmAPIClients(
```
theyoprst: InitSlurmAPIClients is now misnamed. It does client creation and cache spin-up and readiness wait. Either rename to EnsureSlurmInfra / EnsureNodeCacheReady, or split.

internal/slurmapi/node_cache.go
theyoprst: Well, I was wrong, better keep it :-)
First refresh can fail...

theyoprst: I like this claude code suggestion:

```go
  func (nc *NodeCache) run(ctx context.Context) {
      // initial refresh loop — keep trying until one succeeds
      for {
          if nc.refresh(ctx) {
              close(nc.ready)
              break
          }
          select {
          case <-ctx.Done():
              return
          case <-ticker.C:
          }
      }

      // steady-state refresh loop
      ticker := time.NewTicker(nodeCacheRefreshInterval)
      defer ticker.Stop()
      for {
          select {
          case <-ctx.Done():
              return
          case <-ticker.C:
              nc.refresh(ctx)
          }
      }
  }
```

internal/slurmapi/node_cache.go
```diff
@@ -0,0 +1,122 @@
...
+import (
+	"context"
+	"sync/atomic"
+	"time"
+
+	"github.com/go-logr/logr"
+)
+
+const (
+	nodeCacheRefreshInterval  = 15 * time.Second
```
theyoprst: It should be equal to reconcileTimeoutPodEphemeralStorageCheck, ideally reused.
Not sure how hard it would be to pass the value though all levels of indirection

internal/slurmapi/node_cache.go
```diff
@@ -0,0 +1,122 @@
...
+		case <-ticker.C:
+			nc.refresh(ctx)
+		}
+	}
+}
+
+// refresh performs one ListNodes call and updates the cache. Returns true on success.
+func (nc *NodeCache) refresh(ctx context.Context) bool {
+	if !nc.lastRefreshTime.IsZero() && time.Since(nc.lastRefreshTime) > nodeCacheStalenessTimeout {
+		nc.logger.Error(nil, "Slurm node cache is stale, clearing cached data",
```
theyoprst, line 103: Passing nil err is unusal. Why not .Info()?

internal/slurmapi/node_cache.go
```diff
@@ -0,0 +1,122 @@
...
+			"lastRefreshTime", nc.lastRefreshTime,
+			"elapsed", time.Since(nc.lastRefreshTime),
+		)
+		nc.nodes.Store(nil)
+		nc.lastRefreshTime = time.Time{} // reset so this fires exactly once per stale event
```
theyoprst:
```
  8. lastRefreshTime = time.Time{} after clearing — node_cache.go:103. The comment "fires exactly once per stale event" is correct but the mental
  model is subtle: zero now means both "never succeeded" and "succeeded once, went stale, cleared." Consider either:

  - a separate bool cleared to disambiguate, OR
  - keep zero-as-init and just track whether the snapshot pointer is nil; nc.nodes.Load() == nil already encodes "stale or never loaded" as one
  state, which is what callers care about. You may not need lastRefreshTime reset at all if you key the staleness check on snapshot age stored
  alongside the snapshot.

  A cleaner pattern stuffs the timestamp into the snapshot itself:

  type cacheSnapshot struct {
      nodes map[string]Node
      at    time.Time
  }
  nc.nodes atomic.Pointer[cacheSnapshot]

  Then refresh only needs to read nc.nodes.Load() to decide staleness, no separate field. Single source of truth, no reset-to-zero bookkeeping. Worth
   considering.
```

Didn't read it through, but maybe it can improve the readability here

theyoprst:
LGTM, gut gemacht!

BTW, one thing can be improved though, you can improve it if you want but it's optional.
Now the caching goroutine is not cancelled and just dies with the app. It is not a problem, but it would be a problem if we had tests (but we don't have ones, which is another minor thing) -- tests should stop this goroutine, otherwise there will be a gorouine leak. To improve the situation, we could pass a cancellable context instead of context.Background(), and cancel is on SIGTERM. Here is more detailed explanation from Claude Code:

```
⏺ Yes, do the same thing as the exporter. The pattern there is the right one — let me show what it actually is and how to apply it here.

  What the exporter does

  Looking at the actual code:

  cmd/exporter/main.go:243:
  ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)

  internal/exporter/exporter.go:71-87:
  func (e *Exporter) Start(ctx context.Context, addr string) error {
      ...
      go e.collectionLoop(ctx)
      ...
  }

  func (e *Exporter) collectionLoop(ctx context.Context) {
      ticker := time.NewTicker(e.params.CollectionInterval)
      defer ticker.Stop()
      e.collector.updateState(ctx)  // initial
      for {
          select {
          case <-ctx.Done():
              return
          case <-e.stopCh:
              return
          case <-ticker.C:
              e.collector.updateState(ctx)
          }
      }
  }

  The pattern: a signal-aware context is created once at the top, plumbed all the way down, and the long-running goroutine selects on ctx.Done() to
  exit cleanly on SIGTERM. That's exactly what we want for the cache.

  How to apply it to soperatorchecks

  cmd/soperatorchecks/main.go:390 already calls ctrl.SetupSignalHandler() — that's the controller-runtime equivalent of signal.NotifyContext and
  returns a context that's cancelled on SIGINT/SIGTERM. The current code uses it only for mgr.Start(...) and discards the value, so controllers and
  the cache can't see it. Capture and plumb it:

  // cmd/soperatorchecks/main.go
  ctx := ctrl.SetupSignalHandler()

  slurmAPIClients := slurmapi.NewClientSet(ctx)   // pass lifetime ctx in
  ... wire up controllers ...
  if err := mgr.Start(ctx); err != nil { ... }

  // internal/slurmapi/client_set.go
  type ClientSet struct {
      slurmAPIClients map[types.NamespacedName]Client
      nodeCaches      map[types.NamespacedName]*NodeCache
      lifetimeCtx     context.Context  // cancelled on SIGTERM
      mux             *sync.Mutex
  }

  func NewClientSet(lifetimeCtx context.Context) *ClientSet {
      return &ClientSet{
          slurmAPIClients: make(map[types.NamespacedName]Client),
          nodeCaches:      make(map[types.NamespacedName]*NodeCache),
          lifetimeCtx:     lifetimeCtx,
          mux:             &sync.Mutex{},
      }
  }

  // EnsureNodeCache no longer takes ctx — uses cs.lifetimeCtx
  func (cs *ClientSet) EnsureNodeCache(name types.NamespacedName, refreshInterval time.Duration, logger logr.Logger) *NodeCache {
      cs.mux.Lock()
      defer cs.mux.Unlock()
      if nc, ok := cs.nodeCaches[name]; ok {
          return nc
      }
      client, ok := cs.slurmAPIClients[name]
      if !ok {
          return nil
      }
      nc := StartNodeCache(cs.lifetimeCtx, client, refreshInterval, logger)
      cs.nodeCaches[name] = nc
      return nc
  }

  // pod_ephemeral_storage_check.go — drop context.Background()
  nc := r.slurmAPIClients.EnsureNodeCache(
      slurmClusterNamespacedName,
      r.reconcileTimeout,
      log.Log.WithName("NodeCache").WithValues("cluster", slurmClusterNamespacedName),
  )

  That's it. Cache goroutines now exit cleanly on SIGTERM. Tests can pass their own context to NewClientSet, cancel it, and verify the goroutine
  exits — which is also a prerequisite for properly testing the cache without leaks.

  Why this matches what the exporter does

  Same shape: signal-aware ctx born at the top, descends into the goroutines, drives shutdown via ctx.Done(). The controller-runtime difference is
  cosmetic — ctrl.SetupSignalHandler() returns a context.Context, same as signal.NotifyContext, just packaged for the manager.

      cs.nodeCaches[name] = nc
      return nc
  }                                                                                                                                                    nc := StartNodeCache(cs.lifetimeCtx, client, refreshInterval, logger)                                                                            cs.nodeCaches[name] = nc
      return nc                                                                                                                                    }
                                                                                                                                                   // pod_ephemeral_storage_check.go — drop context.Background()
          return nil
      }                                                                                                                                                         nc := StartNodeCache(cs.lifetimeCtx, client, refreshInterval, logger)
      cs.nodeCaches[name] = nc
      return nc                                                                                                                                                 }                                                                                                                                                     }

                                                                                                                                                         // EnsureNodeCache no longer takes ctx — uses cs.lifetimeCtx                                                                                              func (cs *ClientSet) EnsureNodeCache(name types.NamespacedName, refreshInterval time.Duration, logger logr.Logger) *NodeCache {
      cs.mux.Lock()
      defer cs.mux.Unlock()                                                                                                                                     if nc, ok := cs.nodeCaches[name]; ok {                                                                                                                        return nc
      }                                                                                                                                                         client, ok := cs.slurmAPIClients[name]                                                                                                                    if !ok {
          return nil                                                                                                                                            }
      nc := StartNodeCache(cs.lifetimeCtx, client, refreshInterval, logger)                                                                                     cs.nodeCaches[name] = nc                                                                                                                                  return nc                                                                                                                                             }
                                                                                                                                                         // pod_ephemeral_storage_check.go — drop context.Background()
  nc := r.slurmAPIClients.EnsureNodeCache(
      slurmClusterNamespacedName,                                                                                                                               r.reconcileTimeout,                                                                                                                                       log.Log.WithName("NodeCache").WithValues("cluster", slurmClusterNamespacedName),
  )
                                                                                                                                                         That's it. Cache goroutines now exit cleanly on SIGTERM. Tests can pass their own context to NewClientSet, cancel it, and verify the goroutine exits —    which is also a prerequisite for properly testing the cache without leaks.

  Why this matches what the exporter does
  Same shape: signal-aware ctx born at the top, descends into the goroutines, drives shutdown via ctx.Done(). The controller-runtime difference is
  cosmetic — ctrl.SetupSignalHandler() returns a context.Context, same as signal.NotifyContext, just packaged for the manager.
  Why not manager.Runnable

  There's a more idiomatic controller-runtime pattern — register a manager.Runnable whose Start(ctx) owns the cache lifecycle. It's cleaner conceptually
  (cache becomes a first-class subsystem the manager runs alongside reconcilers, with leader-election support if you ever need it), but:

  - It's more invasive — new type, new registration in main, the cache decouples from ClientSet
  - The cache currently is created lazily inside Reconcile; a Runnable would push it earlier, which raises questions about "what cluster do I cache for
  before any reconcile has named one?"
  - The exporter doesn't use it either, so doing this here would also drift from the exporter's pattern

  For now, plumbing the ctx through ClientSet matches the exporter and is the smaller change. If later this controller grows to manage many caches with
  explicit lifecycle, revisit.

  So the concrete answer

  Drop the context.Background(), take a lifetimeCtx in NewClientSet, derive cache goroutines from that. Same model as Exporter.Start(ctx) →
  collectionLoop(ctx) → select { <-ctx.Done(): }. 5–10 lines of code.
```

theyoprst (dismissed):

Uburro:
> LGTM, gut gemacht!
> 
> BTW, one thing can be improved though, you can improve it if you want but it's optional. Now the caching goroutine is not cancelled and just dies with the app. It is not a problem, but it would be a problem if we had tests (but we don't have ones, which is another minor thing) -- tests should stop this goroutine, otherwise there will be a gorouine leak. To improve the situation, we could pass a cancellable context instead of context.Background(), and cancel is on SIGTERM. Here is more detailed explanation from Claude Code:
> 
> ```
> ⏺ Yes, do the same thing as the exporter. The pattern there is the right one — let me show what it actually is and how to apply it here.
> 
>   What the exporter does
> 
>   Looking at the actual code:
> 
>   cmd/exporter/main.go:243:
>   ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
> 
>   internal/exporter/exporter.go:71-87:
>   func (e *Exporter) Start(ctx context.Context, addr string) error {
>       ...
>       go e.collectionLoop(ctx)
>       ...
>   }
> 
>   func (e *Exporter) collectionLoop(ctx context.Context) {
>       ticker := time.NewTicker(e.params.CollectionInterval)
>       defer ticker.Stop()
>       e.collector.updateState(ctx)  // initial
>       for {
>           select {
>           case <-ctx.Done():
>               return
>           case <-e.stopCh:
>               return
>           case <-ticker.C:
>               e.collector.updateState(ctx)
>           }
>       }
>   }
> 
>   The pattern: a signal-aware context is created once at the top, plumbed all the way down, and the long-running goroutine selects on ctx.Done() to
>   exit cleanly on SIGTERM. That's exactly what we want for the cache.
> 
>   How to apply it to soperatorchecks
> 
>   cmd/soperatorchecks/main.go:390 already calls ctrl.SetupSignalHandler() — that's the controller-runtime equivalent of signal.NotifyContext and
>   returns a context that's cancelled on SIGINT/SIGTERM. The current code uses it only for mgr.Start(...) and discards the value, so controllers and
>   the cache can't see it. Capture and plumb it:
> 
>   // cmd/soperatorchecks/main.go
>   ctx := ctrl.SetupSignalHandler()
> 
>   slurmAPIClients := slurmapi.NewClientSet(ctx)   // pass lifetime ctx in
>   ... wire up controllers ...
>   if err := mgr.Start(ctx); err != nil { ... }
> 
>   // internal/slurmapi/client_set.go
>   type ClientSet struct {
>       slurmAPIClients map[types.NamespacedName]Client
>       nodeCaches      map[types.NamespacedName]*NodeCache
>       lifetimeCtx     context.Context  // cancelled on SIGTERM
>       mux             *sync.Mutex
>   }
> 
>   func NewClientSet(lifetimeCtx context.Context) *ClientSet {
>       return &ClientSet{
>           slurmAPIClients: make(map[types.NamespacedName]Client),
>           nodeCaches:      make(map[types.NamespacedName]*NodeCache),
>           lifetimeCtx:     lifetimeCtx,
>           mux:             &sync.Mutex{},
>       }
>   }
> 
>   // EnsureNodeCache no longer takes ctx — uses cs.lifetimeCtx
>   func (cs *ClientSet) EnsureNodeCache(name types.NamespacedName, refreshInterval time.Duration, logger logr.Logger) *NodeCache {
>       cs.mux.Lock()
>       defer cs.mux.Unlock()
>       if nc, ok := cs.nodeCaches[name]; ok {
>           return nc
>       }
>       client, ok := cs.slurmAPIClients[name]
>       if !ok {
>           return nil
>       }
>       nc := StartNodeCache(cs.lifetimeCtx, client, refreshInterval, logger)
>       cs.nodeCaches[name] = nc
>       return nc
>   }
> 
>   // pod_ephemeral_storage_check.go — drop context.Background()
>   nc := r.slurmAPIClients.EnsureNodeCache(
>       slurmClusterNamespacedName,
>       r.reconcileTimeout,
>       log.Log.WithName("NodeCache").WithValues("cluster", slurmClusterNamespacedName),
>   )
> 
>   That's it. Cache goroutines now exit cleanly on SIGTERM. Tests can pass their own context to NewClientSet, cancel it, and verify the goroutine
>   exits — which is also a prerequisite for properly testing the cache without leaks.
> 
>   Why this matches what the exporter does
> 
>   Same shape: signal-aware ctx born at the top, descends into the goroutines, drives shutdown via ctx.Done(). The controller-runtime difference is
>   cosmetic — ctrl.SetupSignalHandler() returns a context.Context, same as signal.NotifyContext, just packaged for the manager.
> 
>       cs.nodeCaches[name] = nc
>       return nc
>   }                                                                                                                                                    nc := StartNodeCache(cs.lifetimeCtx, client, refreshInterval, logger)                                                                            cs.nodeCaches[name] = nc
>       return nc                                                                                                                                    }
>                                                                                                                                                    // pod_ephemeral_storage_check.go — drop context.Background()
>           return nil
>       }                                                                                                                                                         nc := StartNodeCache(cs.lifetimeCtx, client, refreshInterval, logger)
>       cs.nodeCaches[name] = nc
>       return nc                                                                                                                                                 }                                                                                                                                                     }
> 
>                                                                                                                                                          // EnsureNodeCache no longer takes ctx — uses cs.lifetimeCtx                                                                                              func (cs *ClientSet) EnsureNodeCache(name types.NamespacedName, refreshInterval time.Duration, logger logr.Logger) *NodeCache {
>       cs.mux.Lock()
>       defer cs.mux.Unlock()                                                                                                                                     if nc, ok := cs.nodeCaches[name]; ok {                                                                                                                        return nc
>       }                                                                                                                                                         client, ok := cs.slurmAPIClients[name]                                                                                                                    if !ok {
>           return nil                                                                                                                                            }
>       nc := StartNodeCache(cs.lifetimeCtx, client, refreshInterval, logger)                                                                                     cs.nodeCaches[name] = nc                                                                                                                                  return nc                                                                                                                                             }
>                                                                                                                                                          // pod_ephemeral_storage_check.go — drop context.Background()
>   nc := r.slurmAPIClients.EnsureNodeCache(
>       slurmClusterNamespacedName,                                                                                                                               r.reconcileTimeout,                                                                                                                                       log.Log.WithName("NodeCache").WithValues("cluster", slurmClusterNamespacedName),
>   )
>                                                                                                                                                          That's it. Cache goroutines now exit cleanly on SIGTERM. Tests can pass their own context to NewClientSet, cancel it, and verify the goroutine exits —    which is also a prerequisite for properly testing the cache without leaks.
> 
>   Why this matches what the exporter does
>   Same shape: signal-aware ctx born at the top, descends into the goroutines, drives shutdown via ctx.Done(). The controller-runtime difference is
>   cosmetic — ctrl.SetupSignalHandler() returns a context.Context, same as signal.NotifyContext, just packaged for the manager.
>   Why not manager.Runnable
> 
>   There's a more idiomatic controller-runtime pattern — register a manager.Runnable whose Start(ctx) owns the cache lifecycle. It's cleaner conceptually
>   (cache becomes a first-class subsystem the manager runs alongside reconcilers, with leader-election support if you ever need it), but:
> 
>   - It's more invasive — new type, new registration in main, the cache decouples from ClientSet
>   - The cache currently is created lazily inside Reconcile; a Runnable would push it earlier, which raises questions about "what cluster do I cache for
>   before any reconcile has named one?"
>   - The exporter doesn't use it either, so doing this here would also drift from the exporter's pattern
> 
>   For now, plumbing the ctx through ClientSet matches the exporter and is the smaller change. If later this controller grows to manage many caches with
>   explicit lifecycle, revisit.
> 
>   So the concrete answer
> 
>   Drop the context.Background(), take a lifetimeCtx in NewClientSet, derive cache goroutines from that. Same model as Exporter.Start(ctx) →
>   collectionLoop(ctx) → select { <-ctx.Done(): }. 5–10 lines of code.
> ```

https://github.com/nebius/soperator/pull/2500/changes/6e9d01f2497c6ae3ccd34b82714342d51395af6c

theyoprst (approved):

theyoprst (approved):

PR #2508, opened 2026-05-06
SCHED-1213: Move active-check GetJobsByID to /slurmdb/v0.0.41/job/{id}

Uburro:
## Problem

Active-check single-job polling slammed slurmctld with REQUEST_JOB_INFO_SINGLE traffic (~736k calls / ~17 min cumulative scheduler time on b300) even though the data needed lives in slurmdbd.

## Solution

Repointed the active-check `GetJobsByID` lookup at slurmdbd's `/slurmdb/v0.0.41/job/{id}` (renamed to `GetJobsByIDFromAccounting`) and added an empty-result requeue to handle slurmctld→slurmdbd propagation lag.

## Testing

`go test ./internal/slurmapi/... ./internal/controller/soperatorchecks/...` is green 
e2e and mually

## Release Notes

Improvement: active-check job-status polling moves off slurmctld onto slurmdbd, eliminating its dominant contribution to `REQUEST_JOB_INFO_SINGLE` on busy clusters; no CRD/config change.

### DIFF
--- a/internal/controller/soperatorchecks/activecheck_jobs_controller.go
+++ b/internal/controller/soperatorchecks/activecheck_jobs_controller.go
@@ -260,11 +260,23 @@ func (r *ActiveCheckJobReconciler) Reconcile(
 
 		requeue := false
 		latestHandledFinalStateTime := lastHandledFinalStateTime
+		var notFoundInAccountingIDs []string
 		for _, slurmJobID := range ids {
-			slurmJobs, err := slurmAPIClient.GetJobsByID(ctx, slurmJobID)
+			slurmJobs, err := slurmAPIClient.GetJobsByIDFromAccounting(ctx, slurmJobID)
 			if err != nil {
 				logger.Error(err, "failed to get slurm job status")
 				return ctrl.Result{}, err
+			}
+
+			// slurmdbd may not yet have a record for a just-submitted job; the propagation
+			// from slurmctld is asynchronous. Treat an empty result as "not visible yet" and
+			// requeue, otherwise the loop would silently mark the run Complete with no jobs seen.
+			// Accumulate the IDs and log once after the loop so a misconfigured/down slurmdbd
+			// surfaces in logs instead of degrading silently into an infinite requeue.
+			if len(slurmJobs) == 0 {
+				notFoundInAccountingIDs = append(notFoundInAccountingIDs, slurmJobID)
+				requeue = true
+				continue
 			}
 
 			for _, slurmJob := range slurmJobs {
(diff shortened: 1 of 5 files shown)


internal/controller/soperatorchecks/activecheck_jobs_controller.go
```diff
@@ -261,12 +261,20 @@ func (r *ActiveCheckJobReconciler) Reconcile(
...
 				logger.Error(err, "failed to get slurm job status")
 				return ctrl.Result{}, err
 			}
 
+			// slurmdbd may not yet have a record for a just-submitted job; the propagation
+			// from slurmctld is asynchronous. Treat an empty result as "not visible yet" and
+			// requeue, otherwise the loop would silently mark the run Complete with no jobs seen.
+			if len(slurmJobs) == 0 {
+				requeue = true
+				continue
```
theyoprst, line 279: What about logging that no jobs found? Maybe not for each job id, but accumulate some counter (or list of jobs), and log the message like "Not found %d jobs in slurmdbd".
Otherwise if accounting doesn't work at all, our controller will be silent

Uburro: done

theyoprst:
agreeing with codex - this is a real silent-drop bug.

scenario: extensive-check run with `ids=["101","102","103"]`, all finished in slurm. slurmdbd has propagated 102 and 103 but not yet 101 (the ctld→dbd lag the new doc talks about). end times: 101=100, 102=200, 103=300.

reconcile #1:
1. 101 → empty → `requeue=true; continue`
2. 102 → terminal, processed, `latestHandledFinalStateTime=200`
3. 103 → terminal, processed, `latestHandledFinalStateTime=300`
4. after loop, the watermark write at `activecheck_jobs_controller.go:327-335` fires unconditionally: `K8sAnnotationSoperatorChecksFinalStateTime=300`

reconcile #2 (after requeue): `lastHandledFinalStateTime=300`. 101 now visible, EndTime=100. filter at `:291` `100 <= 300 → continue`. 101 silently dropped, fail/cancel reaction never runs.

why it's specific to the accounting path: on slurmctld a finished job stayed visible until purge, so "saw end_time for some IDs but not others" did not happen within one run. on slurmdbd the API can be temporarily inconsistent across IDs in the same `ids` array - the global watermark, designed to suppress re-handling of the same ID, now suppresses first-time handling of a different ID whose end_time predates one already seen.

minimal fix - gate the watermark write on `!requeue`:

```go
if !requeue && latestHandledFinalStateTime > lastHandledFinalStateTime {
    ...
}
```

trade-off: when only some IDs are visible, present jobs' reactions re-fire on the next reconcile because the watermark didn't move. cleaner alternatives are per-ID watermarks or only running reactions once every ID is observed, but the one-line gate closes the silent-drop hole.

regression test worth adding: 2 IDs, first returns `[]slurmapi.Job{}`, second returns a terminal job with an earlier `EndTime`. expect `Requeue=true` AND that `K8sAnnotationSoperatorChecksFinalStateTime` was not written.

theyoprst (dismissed):

theyoprst (approved):

ChessProfessor (approved):

PR #2513, opened 2026-05-08
NOTIC: Support command execution in soperator_instance_login.sh utility

rdjjke:
## Problem
Now, it's pretty easy to execute a specific command on a list of worker nodes:
```
scontrol show hostnames worker-[0-128] | parallel 'ssh {} hostname'
```
However, if you need to execute something on instances under these worker nodes, it's not so simple.
Soperator provides the utility script `soperator_instance_login.sh` for opening a login shell on an instance under a specific worker, but it can't be used for executing commands in a non-interactive mode.

## Solution
Support command execution in `soperator_instance_login.sh`, so the following works:
```
scontrol show hostnames worker-[0-128] | parallel 'soperator_instance_login.sh -w {} -c hostname'
```

## Testing
Tested on a dev cluster.

## Release Notes
`soperator_instance_login.sh` utility can now accept a bash command that will be executed on the K8s node.

### DIFF
--- a/ansible/roles/soperator-scripts/files/fs_usage.sh
+++ b/ansible/roles/soperator-scripts/files/fs_usage.sh
@@ -6,9 +6,9 @@ usage() {
     echo "Get Soperator filesystem usage." >&2
     echo "" >&2
     echo "usage: ${0} [-s] [-l] [-h]" >&2
-    echo "  -s - only show shared filesystems"
-    echo "  -l - only show local filesystems"
-    echo "  -m - only show in-memory filesystems"
+    echo "  -s - only show shared filesystems" >&2
+    echo "  -l - only show local filesystems" >&2
+    echo "  -m - only show in-memory filesystems" >&2
     exit 1
 }
 
--- a/ansible/roles/soperator-scripts/files/soperator_instance_login.sh
+++ b/ansible/roles/soperator-scripts/files/soperator_instance_login.sh
@@ -3,18 +3,26 @@
 set -e # Exit immediately if any command returns a non-zero error code
 
 usage() {
-    echo "Run an interactive login shell on the instance associated with a specific worker node." >&2
+    echo "Run an interactive login shell or execute a command on the instance associated with a specific worker node." >&2
     echo "" >&2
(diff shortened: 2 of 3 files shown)


theyoprst (commented):
main thing - `ssh -t` is wrong on the new `-c` branch (line 46). forcing a pty for non-interactive commands:

- adds `\r` to every output line, so the example `> /tmp/{}-rdma-stat.out` produces a JSON file with CRs
- merges stderr into stdout via the tty
- prints `Pseudo-terminal will not be allocated because stdin is not a terminal` when run under `parallel`

drop `-t` (or `-T`) on the command branch, keep it for the interactive shell.

nits:

1. `command` shadows a bash builtin, rename to `cmd`
2. `echo "examples:"` on line 11 is missing `>&2` while every other line in that block has it - inconsistent with the cleanup the PR is doing in the other two scripts

theyoprst (commented):
thanks for the fixes. one more thing I missed in the first pass:

shell quoting through ssh is broken for any `$cmd` that contains shell metachars. the local `"$cmd"` quoting only protects against word-splitting locally - ssh joins the remote-command args with spaces and the remote shell re-parses. with the documented example `cmd='hostname; rdma stat -j'`, the remote actually sees:

```
sudo chroot /run/nvidia/driver nsenter -t 1 -m -u -i -n bash -c hostname; rdma stat -j
```

remote parses as two statements - `bash -c hostname` (runs inside chroot) and `rdma stat -j` (runs in the outer remote shell, outside chroot/nsenter). any `;`, `|`, `&`, redirection, or whitespace-in-arg in `$cmd` lands in the wrong context. the example happens to mostly work because `hostname` and `rdma stat -j` give similar output inside/outside the container, but for commands that actually need the chroot/nsenter env only the first token runs there.

quote once for the remote shell:

```bash
ssh "${worker_name}" "sudo chroot /run/nvidia/driver nsenter -t 1 -m -u -i -n bash -c $(printf %q "$cmd")"
```

theyoprst (approved):
lgtm, all addressed

theyoprst (approved):

PR #2527, opened 2026-05-13
optimise active checks

itechdima:
## Problem

Problems:
- Active check controller list all slurm jobs even though they could be already completed on the previous reconcilation loop.
- Too noisy logs on big clusters.

## Solution

Add extra annotation for unhandled slurm jobs and iterate over it.
Reduce log verbosity.

## Release Notes

Fix: reduce number of slurm api calls in active check controller.

### DIFF
--- a/helm/slurm-cluster/slurm_scripts/pyxis_caching_importer.sh
+++ b/helm/slurm-cluster/slurm_scripts/pyxis_caching_importer.sh
@@ -8,6 +8,7 @@ readonly cmd="$1"
 readonly cache_dir="${ENROOT_CONTAINER_IMAGES_CACHE_DIR:-/var/cache/enroot-container-images}"
 readonly squashfs_temp_path="${cache_dir}/${SLURM_JOB_ID}.${SLURM_STEP_ID}.sqsh"
 readonly lock_path="${squashfs_temp_path}.lock"
+trap 'rm -f "${lock_path}"' EXIT
 
 # Since it's not an ephemeral squashfs file, we can use compression.
 export ENROOT_SQUASH_OPTIONS="-comp zstd -Xcompression-level 3 -b 1M"
--- a/internal/controller/soperatorchecks/activecheck_jobs_controller.go
+++ b/internal/controller/soperatorchecks/activecheck_jobs_controller.go
@@ -253,7 +253,7 @@ func (r *ActiveCheckJobReconciler) Reconcile(
 			oldUnhandledIDs = []string{}
 		}
 		allIDs := strings.Split(allSlurmJobIDs, ",")
-		unhandledIds := make([]string, 0)
+		unhandledIdsMap := make(map[string]struct{})
 		firstJobId := allIDs[0]
 
 		var jobName string
@@ -282,7 +282,8 @@ func (r *ActiveCheckJobReconciler) Reconcile(
 			slurmJobs, err := slurmAPIClient.GetJobsByIDFromAccounting(ctx, slurmJobID)
 			if err != nil {
 				logger.Error(err, "failed to get slurm job status", "slurm_job_id", slurmJobID)
-				unhandledIds = append(unhandledIds, slurmJobID)
+				unhandledIdsMap[slurmJobID] = struct{}{}
+				requeue = true
 				continue
 			}
 
@@ -293,7 +294,7 @@ func (r *ActiveCheckJobReconciler) Reconcile(
 			// surfaces in logs instead of degrading silently into an infinite requeue.
 			if len(slurmJobs) == 0 {
 				notFoundInAccountingIDs = append(notFoundInAccountingIDs, slurmJobID)
-				unhandledIds = append(unhandledIds, slurmJobID)
+				unhandledIdsMap[slurmJobID] = struct{}{}
 				requeue = true
 				continue
 			}
@@ -306,8 +307,8 @@ func (r *ActiveCheckJobReconciler) Reconcile(
 
 				// Job is not yet finished
 				if !slurmJob.IsTerminalState() || slurmJob.EndTime == nil {
+					unhandledIdsMap[slurmJobID] = struct{}{}
 					requeue = true
-					unhandledIds = append(unhandledIds, slurmJobID)
 					continue
 				}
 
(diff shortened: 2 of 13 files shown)


helm/slurm-cluster/slurm_scripts/pyxis_caching_importer.sh
```diff
@@ -8,6 +8,7 @@ readonly cmd="$1"
 readonly cache_dir="${ENROOT_CONTAINER_IMAGES_CACHE_DIR:-/var/cache/enroot-container-images}"
 readonly squashfs_temp_path="${cache_dir}/${SLURM_JOB_ID}.${SLURM_STEP_ID}.sqsh"
 readonly lock_path="${squashfs_temp_path}.lock"
+trap 'rm -f "${lock_path}"' EXIT
```
theyoprst, line 11: Is it necessary?

internal/controller/soperatorchecks/activecheck_jobs_controller.go
```diff
@@ -231,13 +236,25 @@ func (r *ActiveCheckJobReconciler) Reconcile(
...
+			// Most likely controller was updated, but the previous job is still running.
+			oldUnhandledSlurmJobIDs = allSlurmJobIDs
+		}
+
+		oldUnhandledIDs := strings.Split(oldUnhandledSlurmJobIDs, ",")
+		if oldUnhandledSlurmJobIDs == "" {
+			oldUnhandledIDs = []string{}
+		}
+		allIDs := strings.Split(allSlurmJobIDs, ",")
+		unhandledIds := make([]string, 0)
```
theyoprst: > `unhandledIds` can accumulate duplicates — at `activecheck_jobs_controller.go:281-348`, each iteration of the outer loop calls `GetJobsByIDFromAccounting(slurmJobID)`. If that returns >1 record (e.g., a slurm job array with multiple sub-records for one ID),   the inner loop appends slurmJobID to unhandledIds once per non-terminal sub-job. Across reconciles this compounds (N→N², N³…). A map[string]struct{} or a "append once per outer iteration" guard fixes it.

internal/controller/soperatorchecks/activecheck_jobs_controller.go
```diff
@@ -332,18 +352,20 @@ func (r *ActiveCheckJobReconciler) Reconcile(
...
 			// Maybe we could delete the job because it will not be processed anymore
 			// Otherwise, we will have many of these jobs and they will keep being listed in every Reconcile()
-			k8sJobPatch := client.MergeFrom(k8sJob.DeepCopy())
 			k8sJob.Annotations[K8sAnnotationSoperatorChecksFinalStateTime] = fmt.Sprintf("%d", latestHandledFinalStateTime)
-			if err := r.Job.Patch(ctx, k8sJob, k8sJobPatch); err != nil {
-				return ctrl.Result{}, fmt.Errorf("failed to patch k8s Job: %w", err)
-			}
+		}
+
+		if err := r.Job.Patch(ctx, k8sJob, k8sJobPatch); err != nil {
```
theyoprst, line 372: > Job patch is now unconditional — k8sJobPatch := client.MergeFrom(...) is created and r.Job.Patch is always called, even when neither annotation actually changed. The merge-patch body should be {} in that case (no-op on the apiserver), so this is mostly
   cosmetic, but you could skip the patch when oldUnhandledSlurmJobIDs == strings.Join(unhandledIds, ",") and latestHandledFinalStateTime == lastHandledFinalStateTime to avoid the apicall.

internal/controller/soperatorchecks/activecheck_jobs_controller.go
```diff
@@ -261,11 +278,12 @@ func (r *ActiveCheckJobReconciler) Reconcile(
...
 		latestHandledFinalStateTime := lastHandledFinalStateTime
 		var notFoundInAccountingIDs []string
-		for _, slurmJobID := range ids {
+		for _, slurmJobID := range oldUnhandledIDs {
 			slurmJobs, err := slurmAPIClient.GetJobsByIDFromAccounting(ctx, slurmJobID)
 			if err != nil {
-				logger.Error(err, "failed to get slurm job status")
-				return ctrl.Result{}, err
+				logger.Error(err, "failed to get slurm job status", "slurm_job_id", slurmJobID)
+				unhandledIds = append(unhandledIds, slurmJobID)
```
theyoprst: codex:

```
• unhandledIds = append(unhandledIds, slurmJobID) is only half of what you need. It preserves the ID in memory so it can be written back to the Job annotation later, but it does not make the controller retry that ID.

  The problem is the surrounding control flow in internal/controller/soperatorchecks/activecheck_jobs_controller.go:282:

  1. On accounting error, the code appends the ID and continues, but it does not set requeue = true (internal/controller/soperatorchecks/activecheck_jobs_controller.go:283).
  2. Later, the only persisted effect is writing unhandled-slurm-job-id back to the Job annotation (internal/controller/soperatorchecks/activecheck_jobs_controller.go:360).
  3. If that annotation value is unchanged, and the final-state watermark is also unchanged, the Job patch is a no-op (internal/controller/soperatorchecks/activecheck_jobs_controller.go:367).
  4. This controller watches Job objects, not ActiveCheck status writes (internal/controller/soperatorchecks/activecheck_jobs_controller.go:67).
  5. Since requeue stayed false, deriveSlurmRunStatus can return Complete (internal/controller/soperatorchecks/activecheck_jobs_controller.go:443).

  So in the all-error case, the ID is preserved, but nothing guarantees the controller will come back and read it again. It can just sit in the annotation forever.

  There is a second, worse case. The preserved ID can still be lost logically because of the watermark check at internal/controller/soperatorchecks/activecheck_jobs_controller.go:315.

  Concrete example:

  - oldUnhandledIDs = [101, 102]
  - lookup for 101 fails temporarily, so it is appended back to unhandledIds
  - lookup for 102 succeeds, and 102 is terminal with EndTime = 2000
  - controller advances latestHandledFinalStateTime to 2000 (internal/controller/soperatorchecks/activecheck_jobs_controller.go:345)
  - later, accounting recovers and 101 becomes visible with EndTime = 1500

  Now 101 is still in unhandled-slurm-job-id, but when it is finally fetched, this check fires:

  if slurmJob.EndTime.Unix() <= lastHandledFinalStateTime {
      continue
  }

  Because 1500 <= 2000, job 101 is skipped as “already handled”, even though its reactions were never executed. That is why “we appended it to unhandledIds” is not enough by itself.

  The missing piece is: an accounting lookup failure must also prevent the run from being treated as complete, typically by forcing a retry path (requeue = true) and by not advancing the handled watermark past unresolved IDs.
```

internal/controller/soperatorchecks/activecheck_jobs_controller.go
itechdima: fixed

internal/controller/soperatorchecks/activecheck_jobs_controller.go
itechdima: not big load

internal/controller/soperatorchecks/activecheck_jobs_controller.go
itechdima: added requeue

theyoprst (approved):

PR #2562, opened 2026-05-29
SCHED-1615 Add support for squashfuse enroot mount

usernamedt:
## Problem
Currently, Soperator clusters have non-shared network disks for storing OCI data for Enroot and Docker (container filesystems and image layers).

For Enroot, images are already cached on the shared filesystem as SquashFS files, but container startup still materializes runtime data by unpacking/mounting them through the normal Enroot flow. This adds startup latency for every job and keeps us dependent on fast local/NBS-backed Enroot data storage.

SquashFS images can be mounted directly as read-only filesystems. Combined with overlayfs using an ephemeral upper layer, this allows near-instant startup when the image is already cached and reduces the need for per-node persistent Enroot runtime storage.

## Solution
Added an opt-in Pyxis setting: plugStackConfig.pyxis.useSquashfuse.

When enabled, Soperator renders use_squashfuse=1 for spank_pyxis.so, allowing Pyxis/Enroot to start directly from importer-produced SquashFS images instead of materializing a full Enroot runtime rootfs.

This also updates the runtime prerequisites:

upgrades Pyxis to 0.24.0
upgrades Enroot to 4.1.2
installs and bind-mounts squashfuse into the jail
enables ENROOT_NATIVE_OVERLAYFS=yes so the writable upper layer uses kernel overlayfs over tmpfs
updates CRDs, Helm defaults/tests, docs, and E2E coverage for the new option
The option defaults to false, preserving existing behavior unless explicitly enabled.

## Testing

```
go test ./internal/render/common -run TestRenderPlugstack
go test ./internal/e2e/acceptance/steps
```

## Release Notes
Feature: Added opt-in direct SquashFS startup for Enroot containers via `plugStackConfig.pyxis.useSquashfuse`, improving startup latency when images are already cached.

No breaking change: the option defaults to false; enabling it requires images with Pyxis 0.24.0, Enroot 4.1.2, and squashfuse support.


itechdima (dismissed):

helm/soperator-custom-configmaps/config-files/enroot.conf
```diff
@@ -1,3 +1,4 @@
 ENROOT_CACHE_PATH          /mnt/image-storage/enroot/cache
 ENROOT_DATA_PATH           /mnt/image-storage/enroot/data
 ENROOT_RUNTIME_PATH        /mnt/image-storage/enroot/runtime
+ENROOT_NATIVE_OVERLAYFS    yes
```
rdjjke, line 8: Does it work when you submit containerized jobs as a normal user (not root)?

images/common/scripts/complement_jail.sh
```diff
@@ -153,6 +153,19 @@ pushd "${jaildir}"
         touch "usr/bin/$filename" && mount --bind "$file" "usr/bin/$filename"
     done
 
+    echo "Bind-mount enroot FUSE helpers"
```
rdjjke: As I understand, you need 3 packages to make squashfuse work:
- squashfuse
- fuse3
- libfuse3-3 (is installed as a dependency of squashfuse)

For some reason, you install the squashfuse package (and libfuse3-3) to login and worker container filesystems (install_enroot.sh is called in their Dockerfiles), and also in jail. The problem is that you put it to the plain jail Dockerfile, which installs just various convenience packages that we don't need to automatically install/upgrade. So if this package is needed, we should put its installation in Ansible (it upgrades jail automatically).

And even though this package is installed in jail already, for some reason, you also bind-mount /usr/bin/squashfuse to jail.

Also, you don't install fuse3 at all. So I don't understand where `/usr/bin/fusermount3` comes from. Is it available on both login and worker nodes? (complement_jail.sh is executed in both, and you don't have conditions on where to execute it, so there could be a potential race condition).

So I would suggest bringing a bit more order here.
1. I suggest making sure we need any of these packages in container filesystems (sshd.dockerfile & slurmd.dockerfile). If not, then let's just install everything via ansible in jail, and that's it. Pyxis is a Slurm plugin, and it has various hooks that are executed by Slurm at specific moments of job starting/ending. And another plugin -- `images/common/chroot-plugin/chroot.c` changes process' rootfs. So some early Pyxis hooks see the container filesystem as the root, and some later hooks see jail. If it's only needed by the later hooks, then we don't need these packages in container filesystems.
2. If any of these are needed in container filesystems, then there are two options:
2.1. Only install packages that are needed for the container filesystem, and bind-mount required stuff to jail instead of installing it via Ansible. Then we should also prevent such bind-mounted files from being overwritten if a user executes the reinstallation of a package (dpkg-divert or a mock package).
2.2. If only a small set of files is needed in container filesystems, we can symlink them from jail to container FS (but it depends on which file it is), then there will be no need for mock packages or diverts.

helm/slurm-cluster/values.yaml
```diff
@@ -157,6 +157,8 @@ mpiConfig:
 plugStackConfig:
   pyxis:
     required: true
+    # Direct SquashFS startup is opt-in. It requires Pyxis >= 0.24.0 and squashfuse in the jail.
+    useSquashfuse: false
```
rdjjke, line 161: Did you make it false by default because pyxis is not installed in old jails?
If so, we can make it be installed automatically via ansible during upgrades, and make it true by default

helm/soperator-custom-configmaps/config-files/enroot.conf
rdjjke, line 8: The overlayfs upper dir is probably stored in tmpfs.
But when you write something to the ephemeral container filesystem, how is the consumed memory accounted? Does it go to the Slurm job's cgroup, or not? Because if not, we should increase the system memory reservation (memory that isn't available for job allocations)

images/common/scripts/complement_jail.sh
```diff
@@ -153,6 +153,19 @@ pushd "${jaildir}"
...
+        /lib/${ALT_ARCH}-linux-gnu/libfuse3.so.3 \
+        /usr/lib/${ALT_ARCH}-linux-gnu/libfuse3.so.3; do
+        if [ -e "$file" ]; then
+            target="${file#/}"
+            mkdir -p "$(dirname "$target")"
+            touch "$target" && mount --bind "$file" "$target"
+        fi
+    done
+
     if ! getcap usr/bin/enroot-mksquashovlfs | grep -q 'cap_sys_admin+pe'; then
```
usernamedt, line 156: `cap_sys_admin` is granted here ^

helm/soperator-custom-configmaps/config-files/enroot.conf
usernamedt: Overlayfs requires `cap_sys_admin`, and looks like they are already being granted here https://github.com/nebius/soperator/pull/2562/changes#r3334018430

rdjjke: OK, thanks

images/common/scripts/complement_jail.sh
usernamedt: > As I understand, you need 3 packages to make squashfuse work:
> 
> squashfuse
> fuse3
> libfuse3-3 (is installed as a dependency of squashfuse)
> For some reason, you install the squashfuse package (and libfuse3-3) to login and worker container filesystems (install_enroot.sh is called in their Dockerfiles), and also in jail. The problem is that you put it to the plain jail Dockerfile, which installs just various convenience packages that we don't need to automatically install/upgrade. So if this package is needed, we should put its installation in Ansible (it upgrades jail automatically).
> 
> And even though this package is installed in jail already, for some reason, you also bind-mount /usr/bin/squashfuse to jail.
> 

I've switched to installing the `squashfuse` via Ansible, should be better now. 

> Also, you don't install fuse3 at all. So I don't understand where /usr/bin/fusermount3 comes from. Is it available on both login and worker nodes? (complement_jail.sh is executed in both, and you don't have conditions on where to execute it, so there could be a potential race condition).

Ansible chart now contains direct declarations of the package dependencies.

helm/slurm-cluster/values.yaml
usernamedt: Since it is a new feature, I think it might be better to test before making it true by default.

helm/soperator-custom-configmaps/config-files/enroot.conf
usernamedt: Actually in my initial PR I've missed that and upper dir and work dir were residing somewhere in `/mnt/image-storage/enroot/runtime`. 
Now I've changed it and `ENROOT_CACHE_PATH` / `ENROOT_DATA_PATH` / `ENROOT_RUNTIME_PATH` are being allocated in tmpfs. Unfortunately, an extra helm value was required to do it.

asteny (dismissed):

helm/slurm-cluster/slurm_scripts/job_tmpfs_recreate.sh
```diff
@@ -30,3 +30,16 @@ else
     echo "Creating job tmpfs directory ($JOB_TMPFS_DIR)"
     mkdir -p "${JOB_TMPFS_DIR:?}" || true
 fi
+
```
rdjjke: The purpose of this script is to create per-job directories in a tmpfs volume /mnt/memory that users can use as they wish.

This tmpfs volume is created globally on each node, so the volume itself is created out of the job memory cgroup and out of the job mount namespace. So the whole approach here is problematic, because these directories aren't cleaned automatically. However, users can see these directories when connect to a worker node by SSH. Anyway, the use case of these directories is different.

What are you doing here is that you create directories inside that tmpfs volume (you don't create tmpfs volumes, you create directories in an existing global tmpfs volume).

And there are two options of how it actually works:
1. Enroot/Pyxis just uses these directories for storing its data. Then, this data is not automatically cleaned when jobs end (and maybe is never cleaned because `job_tmpfs_delete.sh` and `job_tmpfs_delete_leftover.sh` scripts know nothing about these dirs).
2. Enroot/Pyxis create tmpfs volumes on their own and just use these directories as mount paths (so inside one tmpfs, different tmpfs volumes are created). Then everything is fine, but it just doesn't make much sense to put these directories inside of /mnt/memory. We can simply use default paths in jail (real data inside will be node-local anyway, because each node creates its own mountpoint)

So I would suggest checking whether these directories are automatically cleaned and whether mountpoints are visible when you connect to the worker node by SSH (`mountpoint` command can be used for that).
You can also switch to the running job's mount namespace (nsenter command) to see these mountpoints if they are actually created by enroot.

usernamedt: Yes, you're right about tmpfs - I've checked and enroot creates its own tmpfs mount inside the runtime dir. So there is no point in defining custom ones. As for data dir, there are hardly any writes in squashfuse mode so it makes sense to leave it default as well

rdjjke (approved):

PR #2567, opened 2026-06-01
SCHED-1648: Better support for Block topology

dstaroff:
## Problem

Existing `worker-init` init container dynamically registers worker topology as if the Slurm cluster is always has `TopologyPlugin=topology/tree` - it generates worker topology as this: `default:<tier0>:<tier1>:<tier2>:<worker>`. This doesn't work for block topology, as it expects workers to be part of only one block.

## Solution

Provide separate dynamic topology generation based on the topology plugin:
- For `topology/tree` - nothing changes
- For `topology/block` - generate `default:<tier0>:<worker>`
- Other topology plugins are not supported

#### Support for pattern names in `topology.conf`

We used to generate a comma-separated list of nodes, which may hit the CM size limit on really big clusters. Now we use patterns which are length-efficient and easy-readable. Sorted order and possible gaps are managed automatically.

#### Support for `topology` section of SlurmCluster CR to be set by Helm

It was impossible to set e.g. `blockSize` via Helm.

## Release Notes

Feature: `TopologyPlugin=topology/block` is supported for dynamic worker topology setting.
Feature: `topology.conf` uses pattern names for nodes instead of comma-separated list.
Feature: `topology` section is possible to change via `slurm-cluster` Helm chart.

### DIFF
--- a/images/worker/worker_init.py
+++ b/images/worker/worker_init.py
@@ -13,20 +13,22 @@ Environment Variables (wait-controller):
 Environment Variables (wait-topology):
     K8S_NODE_NAME: Kubernetes node name (required, from Downward API)
     TOPOLOGY_CONFIGMAP_PATH: Path to mounted ConfigMap (default: /tmp/slurm/topology-node-labels)
-    TOPOLOGY_PLUGIN: Slurm topology plugin override (default: read from slurm.conf)
     TOPOLOGY_WAIT_TIMEOUT: Max wait time in seconds (default: 180)
     TOPOLOGY_POLL_INTERVAL: Poll interval in seconds (default: 5)
+    SLURM_TOPOLOGY_PLUGIN: Slurm topology plugin override (default: read from slurm.conf)
 """
 
 import argparse
 import json
+import logging
 import os
 import re
 import shutil
 import subprocess
 import sys
 import time
-import logging
+from pathlib import Path
+from typing import Any
 
 # Configure logging
 logging.basicConfig(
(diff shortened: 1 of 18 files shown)


images/worker/worker_init.py
```diff
@@ -72,17 +81,19 @@ def get_hostname() -> str:
...
+    service_name = os.environ.get("K8S_SERVICE_NAME")
     logger.info("Using K8S_SERVICE_NAME=%s for NodeAddr construction", service_name)
-    pod_namespace = os.environ.get("K8S_POD_NAMESPACE") 
+    pod_namespace = os.environ.get("K8S_POD_NAMESPACE")
     logger.info("Using K8S_POD_NAMESPACE=%s for NodeAddr construction", pod_namespace)
-    if not pod_name or not service_name or not pod_namespace: 
-        logger.error( "Environment variables K8S_POD_NAME, K8S_SERVICE_NAME, and K8S_POD_NAMESPACE must be set" ) 
-        sys.exit(1) 
-    node_addr = f"nodeaddr={pod_name}.{service_name}.{pod_namespace}.svc" 
+    if not pod_name or not service_name or not pod_namespace:
```
asteny: If vars are require you can just start getting var as os.environ["K8S_POD_NAMESPACE"]
And you will have exception on the first var

images/worker/worker_init.py
```diff
@@ -180,6 +189,25 @@ def get_topology_poll_interval() -> int:
     return int(interval)
 
 
+def get_topology_plugin(slurm_conf_path: str = SLURM_CONF_PATH) -> str:
```
asteny: https://docs.python.org/3/library/os.path.html ?

images/worker/worker_init.py
```diff
@@ -198,7 +226,10 @@ def topology_conf_contains_hostname(topology_conf_path: str, hostname: str) -> b
 
 
 def wait_for_hostname_in_topology_conf(
-    hostname: str, wait_timeout: int, poll_interval: int, topology_conf_path: str = "/etc/slurm/topology.conf"
+    hostname: str,
+    wait_timeout: int,
+    poll_interval: int,
+    topology_conf_path: str = "/etc/slurm/topology.conf",
```
asteny: os.path?

internal/render/worker/container.go
```diff
@@ -114,6 +115,14 @@ func RenderContainerWorkerInit(
 				Value: "5",
 			},
 		)
+		if topologyPlugin != "" {
+			env = append(env,
+				corev1.EnvVar{
+					Name:  "TOPOLOGY_PLUGIN",
```
asteny: SLURM_TOPOLOGY_PLUGIN? 
As I remember we are collecting vars started with SLURM_*

images/worker/worker_init.py
ali-sattari: Will this work for Slurm ranges like [this](https://github.com/nebius/soperator/pull/2567/changes#diff-32d271d9bf47a575fd0979973088cc266ef2b38a05945d0055511a71d13a17d8R109) as well? e.g. `worker-1` in `worker-[0-5]`?

images/worker/worker_init.py
dstaroff: I think you meant `pathlib.Path` implementing `os.PathLike`

images/worker/worker_init.py
dstaroff: Good catch, it won't work.
Fixed it by expanding ranges on the go while checking `topology.conf` lines

ali-sattari (approved):

asteny (approved):

PR #2573, opened 2026-06-02
SCHED-1522 Add autogen_slurm.conf + add warning comment

ChessProfessor:
## Problem

Customers can edit Slurm config files directly inside jail/login environments, but those files are generated and later overwritten by Soperator controllers. This makes local changes non-persistent and easy to lose without an obvious warning.

## Solution

Added clear managed-file warnings to generated Slurm configs. Kept /etc/slurm/slurm.conf as the default entrypoint, but made it a warning-only stub that includes generated autogen_slurm.conf. Renamed custom_slurm.conf to override_slurm.conf to make customer override intent clearer. Updated related Terraform
wording.

## Testing

Added new test + e2e run https://github.com/nebius/soperator/actions/runs/27006234596


internal/consts/configmap.go
```diff
@@ -14,8 +14,9 @@ const (
 	ConfigMapNameSupervisord       = supervisord
 
 	ConfigMapKeySlurmConfig             = "slurm.conf"
+	ConfigMapKeyAutogenSlurmConfig      = "autogen_slurm.conf"
```
theyoprst: Maybe `slurm.gen.conf`. It looks to be a more generic name. Autgen is some MS prduct: https://www.google.com/search?q=autogen&oq=autogen&gs_lcrp=EgZjaHJvbWUyBggAEEUYOTIHCAEQABiABDIHCAIQABiABDIHCAMQABiABDIHCAQQABiABDIHCAUQABiABDIGCAYQRRg9MgYIBxBFGD3SAQc5MDNqMGo3qAIAsAIA&sourceid=chrome&ie=UTF-8

asteny (dismissed):

theyoprst (approved):

PR #2577, opened 2026-06-03
[SCHED-374] Support ephemeral node states in exporter

Uburro:
## Problem
The Slurm exporter didn't recognize ephemeral cloud/power node states (CLOUD, POWER_DOWN, POWER_DRAIN, POWERED_DOWN, POWERING_DOWN, POWERING_UP, POWER_UP). As a result these states were invisible in slurm_node_info, and routine cloud power-down (which sets DOWN/DRAIN) was wrongly counted as node failures and unavailability — producing false alerts.

Solution
Added Is*State() helpers on slurmapi.Node for the seven power/cloud states and exposed them as state_is_* flag labels on slurm_node_info. Power-managed nodes (powering up/down or powered down) are now excluded from the slurm_node_unavailability_duration_seconds and slurm_node_fails_total metrics, since they reflect the normal cloud lifecycle rather than faults. The CLOUD flag alone is not excluded, so a real failure on an active cloud node is still tracked.

## Testing
make test
manual testing on the live cluster

```
slurm_node_info{address="worker-ephemral-1.soperator-nodeset-svc.soperator.svc.cluster.local",comment="",instance_id="",is_unavailable="",node_name="worker-ephemral-1",reason="",reservation_name="",state_base="IDLE",state_is_cloud="true",state_is_completing="",state_is_drain="false",state_is_fail="",state_is_invalid="",state_is_maintenance="false",state_is_not_responding="",state_is_planned="",state_is_power_down="",state_is_power_drain="",state_is_power_up="",state_is_powered_down="true",state_is_powering_down="",state_is_powering_up="",state_is_reserved="false"} 1
```

## Release Notes
Feature: The Slurm exporter now reports ephemeral cloud/power node states as state_is_* labels on slurm_node_info and no longer counts routine cloud power-down as node failures or unavailability.


theyoprst:
@Uburro could you also add these labels to the exporters's README?

theyoprst (approved):

PR #2585, opened 2026-06-08
SCHED-962: add network topology for nodesets in different fabrics

Uburro:
## Problem
All worker nodes are placed under a single root switch, which tells Slurm that IB
connectivity exists between every node. This is false for heterogeneous clusters where
NodeSets live in different IB fabrics — jobs can be scheduled across fabrics that aren't
actually connected.

## Solution
Make the IB fabric configurable per NodeSet via a new spec.topology.fabric field
(defaults to root), and build one root switch per fabric instead of a single shared root:

NodeSet has a fabric → its workers are grouped under a per-fabric root switch named after the fabric, with their tier-2/tier-1 switches beneath it. The fabric roots stay unconnected, so Slurm never schedules a single job across fabrics. Powered-down / unscheduled nodes (known from the CR even without a pod) go under <fabric>.unknown.
NodeSet has no fabric → falls back to the legacy root / unknown switches, so existing single-fabric clusters produce byte-for-byte identical topology.conf.
fabricByNode (Slurm node → fabric, from each NodeSet's spec.topology.fabric) is threaded
into both the tree (BuildTopologyGraph) and block (BuildTopologyBlocks) builders.
Top-of-path switches are attached to their fabric root only when still parentless after the
whole tree is built, so heterogeneous tier depths don't get an intermediate switch re-parented.

## Testing
New/updated unit tests in topologyconfcontroller: two NodeSet fabrics → two unconnected roots, fabric as the root of the tier-1/tier-2 path, powered-down nodes under <fabric>.unknown, mixed explicit-fabric + defaulted NodeSet, no-fabric stays under root, and per-fabric unknown blocks for block topology.
All existing tree/block topology cases still pass (backward compatible).
go test ./internal/controller/topologyconfcontroller/... passes; CRD/deepcopy regenerated.

##Release Notes
NodeSets can now declare an IB fabric via spec.topology.fabric. Workers are grouped under a
per-fabric root switch so Slurm never schedules a job across physically disconnected fabrics.
NodeSets without a fabric keep the previous single-root behavior.


internal/controller/topologyconfcontroller/topology_graph.go
```diff
@@ -144,30 +150,54 @@ func BuildTopologyGraph(
...
+		addTopSwitch(fabric, topSwitch)
 		for i := range len(pathToRoot) - 1 {
 			graph.AddEdge(pathToRoot[i+1], pathToRoot[i])
 		}
 	}
 
-	// Stage 1: every node not placed on a real switch goes under "unknown".
-	const unknownSwitchName = "unknown"
+	// Stage 1: every node not placed on a real switch goes under the default "unknown" switch. The
+	// gpu-cluster-id of powered-down / unscheduled nodes is unknown, so they default to "root".
```
itechdima: nodes in different fabrics could be powered down
since they're attached to the same `root`, during scheduling slurm can power up them and run a job on non-connected GPU nodes

mb it's also better to add `fabric` field to NodeSet CR, so in case of unknown path we would separate them to different roots (or even attach to the same root but to unknown switch)

itechdima (dismissed):

internal/controller/topologyconfcontroller/topology_graph.go
Uburro: >nodes in different fabrics could be powered down
since they're attached to the same root, during scheduling slurm can power up them and run a job on non-connected GPU nodes

It's dynamically provisioned when the pod starts up

Uburro:
```
$ cat etc/slurm/topology.conf 
SwitchName=0efa3121705957aea7b2213618a1d0c3 Nodes=worker-0
SwitchName=3d9a95aa9b1ce5fb1d9de189269138a1 Switches=606eb9929b92fea4d78fd4dbd5dbd032
SwitchName=606eb9929b92fea4d78fd4dbd5dbd032 Nodes=worker-1
SwitchName=8b1623d3def657b2ac9cffaa3a6cbf13 Switches=0efa3121705957aea7b2213618a1d0c3
SwitchName=computegpucluster-i00m5ed7q9y9am5vd5 Switches=3d9a95aa9b1ce5fb1d9de189269138a1,8b1623d3def657b2ac9cffaa3a6cbf13
SwitchName=root Switches=unknown
SwitchName=unknown Nodes=worker-ephemral-0,worker-ephemral-1
```

```
$ scontrol show topo
SwitchName=0efa3121705957aea7b2213618a1d0c3 Level=0 LinkSpeed=1 Nodes=worker-0
SwitchName=3d9a95aa9b1ce5fb1d9de189269138a1 Level=1 LinkSpeed=1 Nodes=worker-1 Switches=606eb9929b92fea4d78fd4dbd5dbd032
SwitchName=606eb9929b92fea4d78fd4dbd5dbd032 Level=0 LinkSpeed=1 Nodes=worker-1
SwitchName=8b1623d3def657b2ac9cffaa3a6cbf13 Level=1 LinkSpeed=1 Nodes=worker-0 Switches=0efa3121705957aea7b2213618a1d0c3
SwitchName=computegpucluster-i00m5ed7q9y9am5vd5 Level=2 LinkSpeed=1 Nodes=worker-[0-1] Switches=3d9a95aa9b1ce5fb1d9de189269138a1,8b1623d3def657b2ac9cffaa3a6cbf13
SwitchName=root Level=1 LinkSpeed=1 Nodes=worker-ephemral-[0-1] Switches=unknown
SwitchName=unknown Level=0 LinkSpeed=1 Nodes=worker-ephemral-[0-1]
```

Uburro:
```
cat etc/slurm/topology.conf

SwitchName=0efa3121705957aea7b2213618a1d0c3 Nodes=worker-0
SwitchName=8b1623d3def657b2ac9cffaa3a6cbf13 Switches=0efa3121705957aea7b2213618a1d0c3
SwitchName=computegpucluster-i00m5ed7q9y9am5vd5 Switches=8b1623d3def657b2ac9cffaa3a6cbf13,computegpucluster-i00m5ed7q9y9am5vd5.unknown
SwitchName=computegpucluster-i00m5ed7q9y9am5vd5.unknown Nodes=worker-1
SwitchName=root Switches=unknown
SwitchName=unknown Nodes=worker-ephemral-0,worker-ephemral-1
```

rdjjke (approved):

PR #2595, opened 2026-06-10
SCHED-1666 Make acceptance tests for dev clusters

ChessProfessor:
## Problem

Acceptance tests were tightly coupled to the e2e flow, which made them impossible to run against an already existing dev cluster. Manual runs also needed a safer way to target a specific Kubernetes cluster instead of relying on whichever kubectl context happened to be active.

## Solution

Added a standalone acceptance runner for dev-cluster usage with explicit CLI configuration.

Also added documentation for building and running acceptance tests against a dev cluster.

## Testing

E2e passed + local test passed.

### DIFF
--- a/internal/e2e/acceptance/runner.go
+++ b/internal/e2e/acceptance/runner.go
@@ -16,6 +16,7 @@ import (
 	"github.com/cucumber/godog"
 	corev1 "k8s.io/api/core/v1"
 
+	slurmv1alpha1 "nebius.ai/slurm-operator/api/v1alpha1"
 	"nebius.ai/slurm-operator/internal/e2e/acceptance/framework"
 	"nebius.ai/slurm-operator/internal/e2e/acceptance/steps"
 )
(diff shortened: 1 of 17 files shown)


internal/e2e/acceptance/runner.go
```diff
@@ -160,6 +166,56 @@ func discoverCluster(ctx context.Context, w *world, state *framework.ClusterStat
...
+		return "", fmt.Errorf("create report dir %q: %w", reportDir, err)
+	}
+	return fmt.Sprintf("pretty,cucumber:%s,junit:%s",
+		filepath.Join(reportDir, "acceptance.cucumber.json"),
+		filepath.Join(reportDir, "acceptance.junit.xml"),
+	), nil
+}
+
+func discoverExpectedNodeSets(ctx context.Context, w *world, clusterName string) ([]framework.ExpectedNodeSet, error) {
+	output, err := w.Kubectl().RunWithDefaultRetry(ctx, "get", "nodesets", "-n", soperatorNamespace, "-o", "json")
```
theyoprst, line 186: It's tautological test now: `kubectl get nodesets` is run here, and the test then runs the very same command and compare results. Of course they will be the same.

I think it doesn't make sense to check node sets in "acceptance" mode. Just keep expected node sets empty and skip the check if it's empty. ExpectedNodeSets should also be renamed to something like ProfileNodeSets, to underline the fact that they are coming from e2e profile (in raw acceptance mode there is no profile, so we don't expect anything).

Can we do something like this?

theyoprst (approved):

PR #2617, opened 2026-06-17
SCHED-1934: Works around NCCL Inspector

dstaroff:
## Problem
1. It's worth installing the plugin with Ansible instead of bash
2. NCCL Inspector's `DUMP_DIR` should be created beforehand otel-collector startup
3. `deleteAfterRead` requires a feature gate opened
4. SCHED-1594: Old empty dump dirs should be cleaned up to reduce FS listing by otel-collector

## Solution
1. Installation of the SPANK plugin is done via Ansible. Creation of the default dump dir is executed during `manage-jail-state` active check
2. Create provided `DUMP_DIR` during otel-collector pod startup
3. Open the feature gate if `deleteAfterRead`
4. Extend `soperator-outputs-log-cleaner`

## Testing
Tested on dev cluster.

## Release Notes
Feature: Added automatic creation NCCL Inspector profiles directory.
Feature: Added automatic cleanup of NCCL Inspector profile directories.

### DIFF
--- a/ansible/roles/spank-nccl-inspector-preconf/defaults/main.yml
+++ b/ansible/roles/spank-nccl-inspector-preconf/defaults/main.yml
@@ -1,0 +1,5 @@
+---
+
+spank_nccl_inspector_preconf_dump_dir: "/opt/soperator-outputs/nccl_profiles"
+spank_nccl_inspector_preconf_dump_dir_mode: "0777"
+spank_nccl_inspector_preconf_dump_dir_create: true
--- a/ansible/roles/spank-nccl-inspector-preconf/tasks/main.yml
+++ b/ansible/roles/spank-nccl-inspector-preconf/tasks/main.yml
@@ -1,0 +1,21 @@
+---
+
+- name: Ensure NCCL Inspector PreConf SPANK plugin is installed
+  ansible.builtin.apt:
+    name:
+      - spank-nccl-inspector-preconf
+    state: present
+    update_cache: true
+  tags:
+    - spank-nccl-inspector-preconf
+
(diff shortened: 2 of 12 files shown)


helm/soperator-activechecks/scripts/ensure-dir-nccl-profiles.sh
```diff
@@ -0,0 +1,4 @@
+set -ex
+
+echo "Ensuring log directory for NCCL Inspector profiles (${SNCCLPRECON_DUMP_DIR})..."
+chroot /mnt/jail /bin/bash -c "mkdir -p '${SNCCLPRECON_DUMP_DIR}'; chmod 777 '${SNCCLPRECON_DUMP_DIR}'"
```
asteny: Simpler install this plugin with ansible and create this dir there.

dstaroff: How are you going to create an arbitrary directory with ansible?

asteny: same as here, with specifying SNCCLPRECON_DUMP_DIR var to the playbook

asteny: https://docs.ansible.com/projects/ansible/latest/playbook_guide/playbooks_variables.html#key-value-format

dstaroff: Fixed

theyoprst (approved):

PR #2621, opened 2026-06-18
SCHED-1178: Don't run ldconfig on each worker pod start

Uburro:
## Problem
The jail root (and its `etc/ld.so.cache`) is a shared filesystem mounted on every node. On each worker pod start (e.g. node auto-replacement), `complement_jail.sh` ran `ldconfig`, rewriting the shared cache and briefly leaving it empty — so jobs starting on other nodes failed to resolve shared libraries.

## Solution
Rebuild the linker cache only when its contents would actually change: probe a throwaway candidate cache against the current one and run `ldconfig` only if they differ. A real `ldconfig` failure now fails the entry point (container restarts) instead of being logged and ignored; a busy `flock` lock is still a benign skip.

## Testing
`bash -n images/common/scripts/complement_jail.sh` passes. Manual: verified on a live cluster that new worker pods skip `ldconfig` when libraries are unchanged and rebuild on driver/library changes, with no empty-cache window for running jobs.

## Release Notes
Fix: Worker pods no longer rebuild the shared linker cache on every start, eliminating a brief window where jobs on other nodes could fail to start.

### DIFF
--- a/images/common/scripts/complement_jail.sh
+++ b/images/common/scripts/complement_jail.sh
@@ -24,6 +24,12 @@ if [ -z "$jaildir" ] || [ -z "$upperdir" ]; then
     usage
 fi
 
+# Default optional variables so `set -u` doesn't abort: `worker` is only set via -w (unset on
+# login nodes), and the GPU env vars may be absent depending on the cluster/nodeset.
+worker="${worker:-}"
+NODESET_GPU_ENABLED="${NODESET_GPU_ENABLED:-}"
+SLURM_CLUSTER_WITH_GPU="${SLURM_CLUSTER_WITH_GPU:-}"
+
 ALT_ARCH="$(uname -m)"
 log "🔧 Using ALT_ARCH = ${ALT_ARCH}"
 
--- a/images/common/scripts/complement_jail.sh
+++ b/images/common/scripts/complement_jail.sh
@@ -3,7 +3,9 @@
 # Complement jaildir by bind-mounting virtual filesystems, users, and NVIDIA binaries from the host filesystem
 
 set -x # Print actual command before executing it
-set -e # Exit immediately if any command returns a non-zero error code
+set -euo pipefail
+
+log() { { set +x; } 2>/dev/null; echo "[$(date -u +%H:%M:%S)] $*"; set -x; }
 
 usage() { echo "usage: ${0} -j <path_to_jail_dir> -u <path_to_upper_jail_dir> [-w] [-h]" >&2; exit 1; }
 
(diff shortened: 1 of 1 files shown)


images/common/scripts/complement_jail.sh
```diff
@@ -204,18 +204,52 @@ pushd "${jaildir}"
...
+            LDCONFIG_CACHE_KEY=$(basename "$(ls usr/lib/${ALT_ARCH}-linux-gnu/libcuda.so.*.* 2>/dev/null | head -1)")
+            LDCONFIG_CACHE_KEY=${LDCONFIG_CACHE_KEY#libcuda.so.}
         else
-            echo "ldconfig FAILED with exit code ${ldconfig_rc}"
+            LDCONFIG_CACHE_KEY="no-gpu"
+        fi
+
+        if [ "$SLURM_CLUSTER_WITH_GPU" = "true" ] && [ "${NODESET_GPU_ENABLED:-}" != "true" ]; then
+            # CPU worker in a GPU cluster: the real GPU libs are replaced by libdummy stubs
+            # (bind-mounted above), so running ldconfig here would build a wrong cache. The
```
rdjjke: > running ldconfig here would build a wrong cache

It won't. Cache only has paths to the libraries, but `ldconfig` skips empty files. So if libdummy is mounted to library paths, then `ldconfig` will produce a correct cache

images/common/scripts/complement_jail.sh
```diff
@@ -127,95 +129,127 @@ pushd "${jaildir}"
...
+        # ld.so.cache lives in the shared jail root, so running ldconfig rewrites the cache
+        # that every running job on every node depends on, briefly leaving it empty. To avoid
+        # this on each worker pod start (e.g. node auto-replacement), rebuild the shared cache
+        # only when its contents would actually change.
+        #
+        # The check builds a throwaway candidate cache from the libraries currently visible in
+        # the jail and compares it with the cache already in place; the real ldconfig runs only
+        # if they differ. This detects any library change (driver upgrade, OpenMPI, custom
+        # libs) without a version heuristic or a shared marker file.
+        if [ "$SLURM_CLUSTER_WITH_GPU" = "true" ] && [ "${NODESET_GPU_ENABLED:-}" != "true" ]; then
```
rdjjke: I don't think this branch is needed. It just should not be so that executing ldconfig from a CPU-only worker in a GPU cluster breaks the linker cache. If it happens, this should be fixed separately.

Even if you add this safeguard, users can still install a package on such a worker, or execute ldconfig somehow differently, and break the shared linker cache anyway.

images/common/scripts/complement_jail.sh
```diff
@@ -127,95 +129,127 @@ pushd "${jaildir}"
...
+            # /tmp inside the jail is node-local (bind-mounted from the host), so the candidate
+            # cache plus -X (don't update symlinks) and -i (don't touch the shared aux-cache)
+            # keep the shared jail untouched while we probe. The first line of `ldconfig -p`
+            # is a header that embeds the cache path, so compare only the entry lines (=>).
+            current_cache_dump=$(mktemp)
+            candidate_cache_dump=$(mktemp)
+            candidate_cache="tmp/ld.so.cache.candidate.$$"
+
+            log "[ldcache] Probing whether the linker cache is up to date"
+            chroot "${jaildir}" /usr/sbin/ldconfig -p 2>/dev/null | grep '=>' | sort > "$current_cache_dump" || true
```
rdjjke: I suggest adding `set -euo pipefail` instead of `set -e` and the beginning of the script not to hide failures

images/common/scripts/complement_jail.sh
```diff
@@ -127,95 +129,127 @@ pushd "${jaildir}"
...
+            chroot "${jaildir}" /usr/sbin/ldconfig -X -i -C "/${candidate_cache}"
+            chroot "${jaildir}" /usr/sbin/ldconfig -C "/${candidate_cache}" -p 2>/dev/null | grep '=>' | sort > "$candidate_cache_dump" || true
```
rdjjke: What about removing `|| true` endings? I think it's better for the script (and container entrypoint) to fail if something went wrong, and restart the container

images/common/scripts/complement_jail.sh
```diff
@@ -127,95 +129,127 @@ pushd "${jaildir}"
...
+            else
+                log "[ldcache] Linker cache is stale, updating it inside the jail"
+                ldconfig_rc=0
+                flock --nonblock etc/complement_jail_ldconfig.lock -c "chroot \"${jaildir}\" /usr/sbin/ldconfig" || ldconfig_rc=$?
+                if [ "$ldconfig_rc" -eq 0 ]; then
+                    log "[ldcache] ldconfig completed successfully (got flock)"
+                elif [ "$ldconfig_rc" -eq 1 ]; then
+                    log "[ldcache] ldconfig SKIPPED (flock busy)"
+                else
+                    log "[ldcache] ldconfig FAILED with exit code ${ldconfig_rc}"
```
rdjjke: I think the entry point should fail in this case, not just log the error

images/common/scripts/complement_jail.sh
Uburro: don t u mind if I remove `set -x `?

Uburro: or `log() { { set +x; } 2>/dev/null; echo "[$(date -u +%H:%M:%S)] $*"; set -x; }`

images/common/scripts/complement_jail.sh
```diff
@@ -127,95 +129,126 @@ pushd "${jaildir}"
...
         done
-        echo "Bind-mount all GPU-specific empty lib files into the host's libdummy"
+        log "[gpu-wait] Bind-mount all GPU-specific empty lib files into the host's libdummy"
         find "${jaildir}/lib/${ALT_ARCH}-linux-gnu" -maxdepth 1 -type f ! -type l -empty -print0 | while IFS= read -r -d '' file; do
             mount --bind "/lib/${ALT_ARCH}-linux-gnu/libdummy.so" "$file"
         done
     fi
 
     # For worker node only
     if [ -n "$worker" ]; then
```
rdjjke, line 200: What about running this even on login nodes? It should be safe for login nodes as well

Uburro: Why does it needed on login node? If we wait worker node anyway

rdjjke: It should be allowed to execute ldconfig from any node. If we don't run ldconfig on login nodes, we can miss bugs.
And yes, we wait for worker nodes anyway, so why add an additional if statement here if it's 100% alowed to run this on login nodes as well

rdjjke: OK, I was wrong, we can't just remove this if statement. Resolving the comment

rdjjke (dismissed):

rdjjke (approved):

PR #2642, opened 2026-06-24
Increase helm release timeouts

asteny:

### DIFF
--- a/helm/soperator-fluxcd/values.yaml
+++ b/helm/soperator-fluxcd/values.yaml
@@ -166,13 +166,13 @@ observability:
     logs:
       version: 0.149.*
       interval: 15m
-      timeout: 15m
+      timeout: 30m
       install:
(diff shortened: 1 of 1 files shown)


helm/soperator-fluxcd/values.yaml
```diff
@@ -166,13 +166,13 @@ observability:
     logs:
       version: 0.149.*
       interval: 15m
-      timeout: 15m
```
Uburro, line 169: in task we have to increase only in kruise, why did u increase others charts timeout?

asteny: There is a timeouts in other charts. Its in a ticket (https://paste.nebius.dev/paste/4f079ac9-e9fd-4a89-aa53-b5a07bd3c727.html)

Uburro (approved):

PR #2671, opened 2026-07-01
SCHED-1904 Remove extensive checks

asteny:


helm/slurm-cluster/slurm_scripts/gpu_health_check.py.json
```diff
@@ -4,7 +4,7 @@
   "platforms": ["8xH100", "8xH200", "8xB200", "8xB300", "4xGB300"],
   "skip_for_cpu_jobs": true,
   "skip_for_partial_gpu_jobs": false,
-  "skip_for_reservation_prefixes": ["suspicious-node"],
+  "skip_for_reservation_prefixes": [],
```
ChessProfessor, line 7: nit: Maybe we should remove it completely?

ChessProfessor (approved):

helm/slurm-cluster/slurm_scripts/gpu_health_check.py.json
asteny: I dont know... this is a functionality in passive checks. If we want to remove skip_for_reservation_prefixes we need to remove additional code. Maybe we will use this skip in some different place

asteny: I created a separate task for that

PR #2675, opened 2026-07-01
SCHED-1741: Make sure we have Active or Passive check to drain nodes where there are empty libnvidia.so files

IronhandedLayman:
## Problem
Intermittently, Nvidia libraries on worker nodes can be mounted incorrectly, leading to an error state where the library exists but is either not loadable or has a size of 0. This causes problems work ML workloads running on said workers.

## Solution
This patch creates an active check that confirms the existence of Nvidia libraries and their ability to load. If those libraries cannot be loaded, then the worker node drains, thus preventing future workloads from running on that node until the issue is resolved.

## Testing
This patch was tested on an existing 2-node Soperator cluster running 3.0.2, both in positive (libraries existing, passed check) and negative tests (library did not exist, check failed, node drained). All cases worked as expected.

## Release Notes
Adding a new health check to existing clusters may cause worker nodes to drain, because libraries were missing that were unexpected, or maybe not used by previous workloads (but given the list of libraries, that is a vanishingly small probability). While this is intended behavior, it may be surprising.


helm/slurm-cluster/slurm_scripts/check_nvidia_libraries.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
+
+set -euxo pipefail
+
+VERSION="0.2.1"
```
theyoprst: Why is this version here?

helm/slurm-cluster/slurm_scripts/check_nvidia_libraries.sh.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "check_nvidia_libraries",
+  "command": "./check_nvidia_libraries.sh",
+  "platforms": ["any"],
```
asteny: Should it be for GPU only?

helm/slurm-cluster/slurm_scripts/check_nvidia_libraries.sh
asteny: This is a version of what?

asteny:
`This patch was tested on an existing 2-node Soperator cluster running 3.0.2, both in positive (libraries existing, passed check) and negative tests (library did not exist, check failed, node drained). All cases worked as expected.`
If this PR for release 3, you need to change branch to the soperator-release-3

IronhandedLayman:
Going to abandon this PR and remake it since the history got messed up after the attempted rebase.

PR #2682, opened 2026-07-02
SCHED-1041: add jail NVML visibility diagnostics

theyoprst:
## Problem

[SCHED-1041](https://nebius.atlassian.net/browse/SCHED-1041) (52 e2e reproductions): GPU containers fail with `libnvidia-ml.so.1: cannot open shared object file`. That symlink is created by `chroot ldconfig` in `complement_jail.sh`, which one worker runs for the whole shared jail (under flock).

Analysis of run [#28568510809](https://github.com/nebius/soperator/actions/runs/28568510809) and three more reproductions shows the same picture every time: the worker that wins the flock cannot see the NVIDIA libs its own `nvidia-container-cli` just mounted, so its ldconfig creates no symlinks. The other worker sees everything but skips ldconfig (lock busy). Current logs can't tell why the winner's view is broken: stale virtiofs cache, mounts landing in a wrong mount namespace, or a cross-node race.

#2621 (main only) rewrote the ldconfig block but didn't fix this — the bug kept reproducing on main, just with a different error text (`file too short`).

## Solution

Logging only, no functional change:

- `complement_jail.sh`: `dump_nvml_state` prints the lib dir as seen by path and by chroot, the nvidia bind mounts from `/proc/self/mountinfo`, and mount-namespace identity — enough to tell the three suspects apart. `verify_nvml_soname` checks the symlink after the ldconfig phase; a worker that skipped ldconfig first waits for the lock holder to finish (bounded by 5 min). All lines are prefixed `[nvml]` for grepping; code comments explain how to read each outcome. Diagnostics run on every worker in a GPU cluster, since a CPU-nodeset worker can win the flock too.
- `prepull-container-image.sh`: on failure, print the jail lib dir state from inside the job. Bounded with `timeout 30` so diagnostics can't turn a fast FAILED into a job TIMEOUT.

Both helpers are standalone functions and port to main's #2621 block as-is.

## Testing

- shellcheck; all new branches exercised with a fake jail dir, under both `set -e` and `set -euo pipefail`
- `helm template` renders, `helm unittest helm/soperator-activechecks` passes (22/22)
- the bug is intermittent and environment-bound; the logs will be evaluated on the next reproduction
- E2E run: https://github.com/nebius/soperator/actions/runs/28666998123

## Release Notes

Diagnostics: added NVML jail-state logging to worker jail setup and the prepull-container-image check to root-cause missing `libnvidia-ml.so.1` failures (SCHED-1041). No functional change; extra startup wait only while another worker's ldconfig is still running (max 5 minutes).

[SCHED-1041]: https://nebius.atlassian.net/browse/SCHED-1041?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/images/common/scripts/complement_jail.sh
+++ b/images/common/scripts/complement_jail.sh
@@ -5,6 +5,9 @@
 set -x # Print actual command before executing it
 set -e # Exit immediately if any command returns a non-zero error code
 
+# Timestamp traced commands to correlate concurrent workers complementing the shared jail
+PS4='+ [$(date "+%H:%M:%S.%3N")] '
+
 usage() { echo "usage: ${0} -j <path_to_jail_dir> -u <path_to_upper_jail_dir> [-w] [-h]" >&2; exit 1; }
 
 while getopts j:u:wh flag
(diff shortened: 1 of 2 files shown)


images/common/scripts/complement_jail.sh
```diff
@@ -5,6 +5,9 @@
 set -x # Print actual command before executing it
 set -e # Exit immediately if any command returns a non-zero error code
 
+# Timestamp traced commands to correlate concurrent workers complementing the shared jail
+PS4='+ [$(date "+%H:%M:%S.%3N")] '
```
Uburro, line 9: where is this value used ?

images/common/scripts/complement_jail.sh
```diff
@@ -27,6 +30,74 @@ echo "🔧 Using ALT_ARCH = ${ALT_ARCH}"
...
+#   jail, or the CLI mounted into another namespace (check the mountinfo counts below)
+# - 0-size versioned libs -> shared-FS placeholders visible, bind mounts detached
+# - bind mounts present in mountinfo while invisible by path -> stale dentries (virtiofs)
+# - bind mounts absent from mountinfo -> the CLI's mounts landed in another namespace
+# - "self" and "via jail proc" namespaces differ -> the CLI targeted a wrong namespace:
+#   it derives the target from ${jaildir}/proc/<its parent pid>/ns/mnt
+# Must be called with CWD = ${jaildir}.
+# shellcheck disable=SC2010 # ls|grep is deliberate: the probe must exercise readdir
+dump_nvml_state() {
+    echo "=== NVML jail state: ${1} ==="
```
Uburro: Maybe add some prefix to this log for easy grepping, like [NVML], [lib], etc.

images/common/scripts/complement_jail.sh
theyoprst: PS4 is used when Bash tracing is enabled with (like in this script):

```
set -e
```

After that, Bash prints each command before executing it. The value of PS4 is printed at the beginning of each traced line.

So with this PS4, traced commands will look like this:

```
+ [14:23:01.127] echo hello
hello
+ [14:23:01.130] ls
```

Uburro (dismissed):

images/common/scripts/complement_jail.sh
theyoprst: done

Uburro (approved):

PR #2684, opened 2026-07-02
SCHED-1741: Make sure we have Active or Passive check to drain nodes where there are empty libnvidia.so files (amended)

IronhandedLayman:
## Problem

Intermittently, Nvidia libraries on worker nodes can be mounted incorrectly, leading to an error state where the library exists but is either not loadable or has a size of 0. This causes problems work ML workloads running on said workers.

## Solution

This patch creates an active check that confirms the existence of Nvidia libraries and their ability to load. If those libraries cannot be loaded, then the worker node drains, thus preventing future workloads from running on that node until the issue is resolved.

## Testing

This patch was tested on an existing 2-node Soperator cluster running 3.0.2, both in positive (libraries existing, passed check) and negative tests (library did not exist, check failed, node drained). All cases worked as expected. _Even though this is tested against a 3.0.2 soperator cluster, it is expected to work with all versions, including latest._

## Release Notes

Adding a new health check to existing clusters may cause worker nodes to drain, because libraries were missing that were unexpected, or maybe not used by previous workloads (but given the list of libraries, that is a vanishingly small probability). While this is intended behavior, it may be surprising.

### DIFF
--- a/helm/slurm-cluster/slurm_scripts/check_nvidia_libraries.sh.json
+++ b/helm/slurm-cluster/slurm_scripts/check_nvidia_libraries.sh.json
@@ -1,0 +1,17 @@
+{
+  "name": "check_nvidia_libraries",
+  "command": "./check_nvidia_libraries.sh",
+  "platforms": ["8xGPU", "1xGPU", "4xGPU"],
+  "skip_for_cpu_jobs": false,
+  "skip_for_partial_gpu_jobs": false,
+  "skip_for_reservation_prefixes": [],
+  "contexts": ["prolog", "hc_program"],
+  "node_states": ["any"],
+  "on_fail": "drain",
+  "on_ok": "undrain",
+  "reason_base": "[node_problem] $name",
+  "reason_append_details": true,
+  "run_in_jail": true,
+  "log": "slurm_scripts/$worker.$name.$context.out",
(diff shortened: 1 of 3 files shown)


asteny (approved):

helm/slurm-cluster/slurm_scripts/check_nvidia_libraries.sh.json
```diff
@@ -0,0 +1,17 @@
...
+  "command": "./check_nvidia_libraries.sh",
+  "platforms": ["8xGPU", "1xGPU", "4xGPU"],
+  "skip_for_cpu_jobs": false,
+  "skip_for_partial_gpu_jobs": false,
+  "skip_for_reservation_prefixes": [],
+  "contexts": ["prolog", "hc_program"],
+  "node_states": ["any"],
+  "on_fail": "drain",
+  "on_ok": "undrain",
+  "reason_base": "[node_problem] $name",
```
rdjjke, line 12: All nodes drained with `[node_problem]` prefix are being automatically replaced by Soperator automation.
But these drains show issues with hardware, not software.

This check doesn't show hardware issues, so automatic node replacement isn't needed. I suggest to change the prefix to something else, e.g. `[software_problem]`

PR #2694, opened 2026-07-03
SCHED-1951: Optional Docker on clusters

Uburro:
## Problem

Clusters deployed without image-storage disks still get a Docker daemon,
docker-proxy sidecar, and docker CLI in the jail — Docker seems available
but fails with ambiguous errors. The `docker-memory-limit-oomkilled` active
check was added by mistake.

## Solution

- New `docker.enabled` flag (NodeSet CRD + `nodesets` chart, default `true`),
  set by terraform based on image-storage disk presence. When disabled: no
  docker-proxy sidecar, and the jail's `docker`/`dockerd` are masked with a
  clear error message.
- New `docker.enabled` value in `soperator-custom-configmaps` removes the
  dockerd supervisord program.
- Removed the `docker-memory-limit-oomkilled` active check.

## Testing

- `make lint`, `make test` (new unit test for the docker-proxy gating),
`make helmtest`
- manually rendered charts with `docker.enabled: false`
- on the live cluster

## Release Notes

Feature: Docker can be disabled per NodeSet (`docker.enabled`) for clusters
without image-storage disks. Removed the `docker-memory-limit-oomkilled`
active check. Default keeps Docker enabled — no impact on existing clusters.


dstaroff (approved):

images/common/scripts/complement_jail.sh
```diff
@@ -211,10 +212,26 @@ pushd "${jaildir}"
         mount --bind /usr/sbin/slurmd usr/sbin/slurmd
         mount --bind /usr/sbin/slurmstepd usr/sbin/slurmstepd
 
-        log "[docker] Bind-mount dockerd stuff from container to the jail"
-        mkdir -p etc/docker
-        touch etc/docker/daemon.json
-        mount --bind "/etc/docker/daemon.json" "etc/docker/daemon.json"
+        if [ "$SOPERATOR_DOCKER_ENABLED" = "true" ]; then
```
rdjjke, line 215: Will enabling docker on a cluster where it was previously disabled lead to container restarts so that this new branch is executed?

rdjjke (approved):

images/common/scripts/complement_jail.sh
Uburro: yes, because the docker-proxy sidecar  will be added

PR #2698, opened 2026-07-06
Backport: [SCHED-374] Support ephemeral node states in exporter

itechdima:
## Problem

Missing state labels for ephemeral nodes.

## Solution

Cherry-pick exporter from main.

## Release Notes

Feature: support ephemeral node states in exporter


Uburro (approved):

itechdima:
we won't do it for 3.0, only starting from 4.0

PR #2703, opened 2026-07-06
SCHED-1041: Pre-create GPU lib placeholders so a failing nvidia-container-cli cannot break the cluster

theyoprst:
## Problem

At cold cluster bring-up all workers run `nvidia-container-cli` against the shared jail at the same moment. The CLI creates 0-byte placeholder files on the shared jail and bind-mounts the host driver libs over them. Confirmed in 7 e2e reproductions:

1. The workers race to create the same files. The loser fails (`mount error: file creation failed: .../gsp_ga10x.bin: file exists`), its container restarts, and the solo re-run succeeds — this worker always recovers by accident.
2. Before dying, the failing CLI deletes every 0-byte placeholder it had mounted — including the ones under the *other* worker's live bind mounts. That worker's kernel detaches its mounts, and it stays GPU-broken for the pod's lifetime: `prepull-container-image` fails the bring-up, or multi-node GPU jobs fail later with `file too short`.

The second half is not limited to bring-up races: any CLI failure on any worker at any time deletes the placeholders under the whole fleet's mounts.

## Solution

Pre-create the placeholders as small non-empty marker files before the CLI runs (`precreate_gpu_placeholders` in `complement_jail.sh`):

- the file set comes from `nvidia-container-cli list` at runtime, so driver upgrades are handled automatically; the `/dev` and `/run` entries resolve to objects already created by the kernel and system daemons, so the existence check skips them;
- on the first bring-up the missing files are created under a cross-node flock, with the host file's mode; on every later pod start everything already exists and no lock is taken;
- the CLI then finds every file present and skips creation — no race, no crash;
- the CLI's failure cleanup only deletes *empty* files, so a failing CLI can no longer break the other workers, whatever the failure cause.

The CLI result is verified instead of trusted: every mount the CLI created must stay attached and show real content by path (by size — a placeholder passes plain existence checks), whatever file set the enabled driver capabilities select. The `libnvidia-ml.so.1` soname is additionally verified after ldconfig (repaired by a local ldconfig run if missing) when NVML is mounted. On any failure the container exits so kubelet restarts it.

The two places that treated "empty file" as "placeholder" (the CPU-worker libdummy mounts and the populate_jail cleanup) now match by size and versioned lib name (`lib*.so.*` smaller than 64 bytes), covering both old 0-byte and new marker placeholders while leaving the jail image's other small files (ncurses linker scripts, glibc stub archives) alone.

## Testing

- shellcheck + smoke tests with stubbed CLI/flock/stat: cold and warm paths, node-local filtering, failure exits, the size predicates.
- three green e2e cold bring-ups ([28922118210](https://github.com/nebius/soperator/actions/runs/28922118210), [28952699476](https://github.com/nebius/soperator/actions/runs/28952699476), [28973074103](https://github.com/nebius/soperator/actions/runs/28973074103), SB_MAN_H200) together exercised every pre-create path: sole creator, all-present gate, and a worker that queued on the lock mid-creation and found everything present on the re-check. Both CLIs skipped all creations in each run, no anomalies. The `[nvml]` log telemetry stays for verifying this in production.

## Release Notes

Fix: GPU driver lib placeholders in the shared jail are now pre-created as non-empty files, so concurrent `nvidia-container-cli` runs never race to create them and a failing run can no longer delete them from under the other workers' bind mounts — fixing intermittent NVML load failures ("cannot open shared object file" / "file too short") on multi-worker bring-up.

### DIFF
--- a/images/common/scripts/complement_jail.sh
+++ b/images/common/scripts/complement_jail.sh
@@ -30,67 +30,245 @@ echo "🔧 Using ALT_ARCH = ${ALT_ARCH}"
 
 SLURM_LIB_PATH="usr/lib/${ALT_ARCH}-linux-gnu/slurm"
 
-# SCHED-1041 diagnostics.
-# The NVIDIA soname symlinks (e.g. libnvidia-ml.so.1) are created only by the "chroot ldconfig" run below,
-# and sometimes the worker that wins the ldconfig flock has a stale view of the jail lib dir right after
-# its own nvidia-container-cli run, so its ldconfig creates no sonames and GPU containers fail on every node of the cluster.
-# How to read the output:
-# - path/chroot view has no libnvidia-ml at all -> stale dentry/readdir view of the shared
-#   jail, or the CLI mounted into another namespace (check the mountinfo counts below)
-# - 0-size versioned libs -> shared-FS placeholders visible, bind mounts detached
-# - bind mounts present in mountinfo while invisible by path -> stale dentries (virtiofs)
-# - bind mounts absent from mountinfo -> the CLI's mounts landed in another namespace
-# - "self" and "via jail proc" namespaces differ -> the CLI targeted a wrong namespace:
-#   it derives the target from ${jaildir}/proc/<its parent pid>/ns/mnt
-# Must be called with CWD = ${jaildir}. All output is prefixed with [nvml] for grepping.
+# NVML diagnostics dump, [nvml]-prefixed for grepping. Must be called with CWD = ${jaildir}.
+# The failure mechanism these dumps watch for: when one worker's nvidia-container-cli fails, its
+# cleanup unmounts and DELETES every 0-byte mount target it had mounted on the shared jail
+# (nvc_mount.c unmount() -> file_remove()), unlinking the placeholders under the other workers' live bind mounts;
+# their kernels then detach those mounts. precreate_gpu_placeholders below makes the placeholders non-empty
+# and thus immune; these dumps are the telemetry proving that.
 # shellcheck disable=SC2012 # ls is deliberate: the probe must exercise readdir on the live directory
 dump_nvml_state() {
     echo "[nvml] === NVML jail state: ${1} ==="
+
+    # No libnvidia-ml entries at all -> the placeholders were deleted (a neighbor's CLI failed and its cleanup ran);
+    # 0-size or marker-size versioned libs -> this worker's bind mounts are gone
     echo "[nvml] --- path view:"
     ls -lai "usr/lib/${ALT_ARCH}-linux-gnu/" 2>&1 | awk '/libnvidia-ml|libcuda\.so/ {print "[nvml]   " $0; n++} END {if (!n) print "[nvml]   no libnvidia-ml/libcuda entries visible by path"}'
+
+    # A difference from the path view means ldconfig resolves the jail through another view
     echo "[nvml] --- chroot view (what ldconfig sees):"
     chroot "${jaildir}" ls -lai "/usr/lib/${ALT_ARCH}-linux-gnu/" 2>&1 | awk '/libnvidia-ml|libcuda\.so/ {print "[nvml]   " $0; n++} END {if (!n) print "[nvml]   no libnvidia-ml/libcuda entries visible via chroot"}'
+
+    # 0 mounts while the libs above look healthy -> the CLI's mounts landed in another namespace
     echo "[nvml] --- nvidia lib bind mounts in this namespace: $(grep -cF "${jaildir}/usr/lib/${ALT_ARCH}-linux-gnu/libnvidia" /proc/self/mountinfo) (libnvidia-ml: $(grep -cF "${jaildir}/usr/lib/${ALT_ARCH}-linux-gnu/libnvidia-ml" /proc/self/mountinfo))"
+
+    # Differing values -> the CLI targeted a wrong namespace: it derives the target
+    # from ${jaildir}/proc/<its parent pid>/ns/mnt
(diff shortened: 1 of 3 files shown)


images/common/scripts/complement_jail.sh
```diff
@@ -30,69 +30,249 @@ echo "🔧 Using ALT_ARCH = ${ALT_ARCH}"
...
+jail_mount_target() {
+    local name
+    name=$(basename "$1")
+    case "${name}" in
+        lib*.so*) realpath -m "./usr/lib/${ALT_ARCH}-linux-gnu/${name}" ;;
+        *)        realpath -m ".$1" ;;
+    esac
+}
+
+precreate_gpu_placeholders() {
```
rdjjke: I don't understand why it is so complicated.
What's the difference with just this change to `complement_jail.sh`?
```
get_missing_cuda_driver_files() {
  local missing_files=()
  for path in $(nvidia-container-cli list); do
    if [[ ! -e "/mnt/jail${path}" ]]; then
      missing_files+=("$path")
    fi
  done
  printf "%s\n" "${missing_files[@]}"
}

mount_cuda_driver() {
  nvidia-container-cli \
    --user \
    --debug=/dev/stderr \
    --no-pivot \
    configure \
    --no-cgroups \
    --cuda-compat-mode=disabled \
    --ldconfig=$FAKE_LDCONFIG \
    --device=all \
    "${cap_args[@]}" \
    "${jaildir}"
}


# ... in the main code path ...
readarray -t missing_files < <(get_missing_cuda_driver_files)

if [[ ${#missing_files[@]} -eq 0 ]]; then
  mount_cuda_driver
else
  (
    flock --exclusive 200
    readarray -t missing_files < <(get_missing_cuda_driver_files)
    for path in "${missing_files[@]}"; do
      host_path="/mnt/jail${path}"
      mkdir -p "$(dirname "$host_path")"
      echo "soperator cuda driver mount placeholder" > "$host_path"
    done
    mount_cuda_driver
  ) 200>"/mnt/jail/etc/complement_jail_cuda_driver_mount.lock"
fi
```

images/jail/populate_jail_entrypoint.sh
```diff
@@ -33,13 +33,15 @@ remove_empty_lib_mount_targets() {
...
+    echo "Removing placeholder library files that were used as bind-mount targets on the previous cluster"
+    # Placeholders are small marker files created by complement_jail.sh (or 0-byte on jails
+    # populated before that change) — match both by size; no real library is this small
     ARCH_LIST="x86_64 aarch64"
     for arch in $ARCH_LIST; do
         dir="/mnt/jail/lib/${arch}-linux-gnu"
         [ -d "$dir" ] || continue
         find "$dir" \
-            -maxdepth 1 -type f -empty -name '*.so*' -print |
+            -maxdepth 1 -type f -size -64c -name '*.so*' -print |
```
rdjjke: I think we should remove both empty files and files with our placeholder contents, to make it usable during upgrades from old versions to this one

theyoprst: It is exactly what is done: all files less than 64 bytes are removed. Do you also mean to check the contents? We can do it, but there will be a risk: if the constant with the file contents changes in the future, there will be problems. Plus, we need to copy-paste the constant into several scripts.

images/common/scripts/complement_jail.sh
theyoprst: It's pretty much like current PR changes, but with several misses:

1. [Critical] nvidia-container-cli doesn't mount paths 1:1. It puts all libraries in one directory ("flattens" them); there is one example of such a library now: see `jail_mount_target` and its long comment, which describes this example.

2. [Critical] No validation that it was really mounted; the worker can start with no libraries as a result; I would keep it: it is an extra protection against at least creation-time races

3. [Critical] `nvidia-container-cli` call is under the flock; is there a reason for that?

4. [Normal] Your version checks and can write files outside of jail, e.g. to `/run` or `/dev`, I'm not sure how bad it is, but there is protection of that in my version around `jail_dev` variable, 8 lines of code including comments, which give a guarantee that we don't write anything outside of jail

5. [Normal] files don't match permissions => nvidia-container-cli will think that they are absent (because st_mode doesn't match), will try to create them: most likely harmless, but better to prevent these syscalls at all

theyoprst: > [Normal] Your version checks and can write files outside of jail, e.g. to /run or /dev, I'm not sure how bad it is, but there is protection of that in my version around jail_dev variable, 8 lines of code including comments, which give a guarantee that we don't write anything outside of jail

Decided to get rid of this check. I'll also re-run e2e to ensure that it works without any issues

images/common/scripts/complement_jail.sh
```diff
@@ -30,69 +30,237 @@ echo "🔧 Using ALT_ARCH = ${ALT_ARCH}"
...
+    # Cold jail bring-up: create the missing placeholders under a cross-node lock so exactly one worker does it;
+    # the others re-check under the lock and find everything present.
+    echo "[nvml] ${#missing[@]} GPU placeholders missing, creating under lock"
+    exec {lockfd}>etc/complement_jail_gpu_placeholders.lock
+    if ! flock --wait 60 "${lockfd}"; then
+        echo "[nvml] timed out waiting for the GPU placeholder lock, exiting so the container restarts"
+        exit 1
+    fi
+    created=()
+    for host_path in "${missing[@]}"; do
```
rdjjke, line 240: Shouldn't we collect a new list of missing files after flock?
Otherwise, there could be race conditions - several workers will be creating such placeholders concurrently. Maybe it won't lead to any issues, but I think it will be safer to re-check missing libraries after flock. Then there's 100% guarantee that these files actually don't exist, and the lock holder can create them

theyoprst: We already do this check in this loop, see 2 lines below:

```bash
        [ -e "${jail_path}" ] && continue  # created by a previous lock holder
```

rdjjke (approved):

PR #2715, opened 2026-07-08
Mount slurm-scripts to nodesets worker

andriishestakov:
Mount %s-slurm-scripts configmap to worker if cluster name is set

### DIFF
--- a/helm/nodesets/templates/_helpers.tpl
+++ b/helm/nodesets/templates/_helpers.tpl
@@ -114,6 +114,22 @@ customInitContainers:
 {{- end }}
 {{- end -}}
 
+{{/* Name of the slurm-scripts ConfigMap. Matches slurm-cluster chart's naming convention. */}}
+{{- define "nodesets.slurmScriptsCMName" -}}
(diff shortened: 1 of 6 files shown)


Uburro:
@andriishestakov I understand the problem, but ...


The injected entries are unconditional and unguarded. If a NodeSet's values already contain a mount named slurm-scripts at /opt/slurm_scripts, the rendered pod gets duplicate volume names and duplicate mountPaths. Both rejected by API validation (must be unique), so the StatefulSet never rolls. 

CM naming duplicated across charts, and not identically. slurm-cluster computes it as default .Chart.Name .Values.clusterName | trunc 63 | trimSuffix "-" + -slurm-scripts. The new helper skips both the .Chart.Name fallback and the trunc 63, so an empty or >63-char clusterName produces divergent names (silent no-mount, or a reference to a CM that doesn't exist → pods stuck in ContainerCreating). An explicit slurmScriptsConfigMapName value that defaults to the convention would decouple the two charts.

andriishestakov:
Added checks for slurm-cluster and nodesets.

PR #2717, opened 2026-07-08
ActiveChecks: Fix chroot for create-user check

andriishestakov:
SYS_CHROOT is required in ActiveChecks to allow chroot to jail and create user

### DIFF
--- a/internal/consts/container.go
+++ b/internal/consts/container.go
@@ -18,8 +18,9 @@ const (
 	ContainerNameCustom            = "custom-container"
 	ContainerNameSConfigController = SConfigControllerName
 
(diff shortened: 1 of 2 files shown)


Uburro:
Can you show us a case where it doesn't work?

andriishestakov:
create-user-soperatorchecks job fails with this:

```
Creating soperatorchecks user...
+ echo 'Creating soperatorchecks user...'
+ chroot /mnt/jail /bin/bash -s
chroot: cannot change root directory to '/mnt/jail': Operation not permitted
```

SYS_CHROOT is not default in crio

asteny (approved):

asteny:
@andriishestakov  Thanks for your contribution!

PR #2719, opened 2026-07-09
SCHED-1897: Add guard to avoid calling get_node_info out of hc_program for undrain…

fabrizio2210:
## Problem

Passive checks can generate unnecessary Slurm RPCs from job-coupled contexts.

In particular, `boot_disk_full` is configured with `on_ok: undrain` and runs in both `prolog` and `hc_program`. Although the check itself only performs a local disk-usage check, `check_runner.py` evaluates the `on_ok: undrain` path after a successful `prolog` run. This calls `get_node_info()`, which executes `scontrol show node ... --json`.

As a result, every successful `boot_disk_full` execution in `prolog` can produce an unnecessary `scontrol show node` call. This adds per-job/per-node RPC load to Slurm and was identified as one of the main contributors in SCHED-1897.

## Solution

Restrict `on_ok` recovery actions to the `hc_program` context.

`undrain` and `uncomment` are intended to be issued only by periodic health checks, not by `prolog` or `epilog`. With this change, successful checks in `prolog`/`epilog` no longer evaluate the `undrain`/`uncomment` path and therefore no longer call `get_node_info()` just to decide whether a node should be resumed or uncommented.

Failure handling remains unchanged: checks can still drain or comment nodes on failure according to their configuration.

## Testing

Tested with local `check_runner_test.py`.
 
## Release Notes

Fix: Reduced Slurm RPC load from passive checks by preventing `prolog`/`epilog` success paths from evaluating `undrain`/`uncomment` actions.

The intended auto-recovery behavior remains available from `hc_program`; no user-facing configuration changes or migrations are required.

### DIFF
--- a/helm/slurm-cluster/slurm_scripts/check_runner_test.py
+++ b/helm/slurm-cluster/slurm_scripts/check_runner_test.py
@@ -1,0 +1,120 @@
+import importlib.util
+import os
+import tempfile
+import unittest
+from pathlib import Path
+from unittest import mock
+
+
+CHECK_RUNNER_PATH = Path(__file__).with_name("check_runner.py")
+
+
+class CheckRunnerOnOkContextTest(unittest.TestCase):
+    def load_runner(self, context: str, tmpdir: str):
+        env = {
+            "SLURMD_NODENAME": "worker-1",
+            "CHECKS_OUTPUTS_BASE_DIR": tmpdir,
+            "CHECKS_CONTEXT": context,
+            "CHECKS_CONFIG": str(Path(tmpdir) / "checks.json"),
+            "CHECKS_RUNNER_OUTPUT": "/dev/null",
+        }
+        patcher = mock.patch.dict(os.environ, env, clear=False)
+        patcher.start()
+        self.addCleanup(patcher.stop)
+
+        module_name = f"check_runner_under_test_{context}_{id(self)}"
+        spec = importlib.util.spec_from_file_location(module_name, CHECK_RUNNER_PATH)
+        runner = importlib.util.module_from_spec(spec)
+        spec.loader.exec_module(runner)
+        return runner
+
+    def test_undrain_on_ok_is_ignored_outside_hc_program(self):
+        with tempfile.TemporaryDirectory() as tmpdir:
+            runner = self.load_runner("prolog", tmpdir)
+            check = runner.Check(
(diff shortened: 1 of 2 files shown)


helm/slurm-cluster/slurm_scripts/check_runner_test.py
```diff
@@ -0,0 +1,120 @@
+import importlib.util
```
asteny, line 1: Where are u running this test?

fabrizio2210: It can be run manually with `python3 helm/slurm-cluster/slurm_scripts/check_runner_test.py`. I found it usefult to test the change.
But you're right, it should be more structured. 

In the repo I see many integration tests, where can I place a unit test like this?

asteny: Lets run it in our ci

fabrizio2210: Do you mean a file in https://github.com/nebius/soperator/tree/main/.github/workflows?

fabrizio2210: I created an extra PR (https://github.com/nebius/soperator/pull/2730) based on this branch for an easier review. Unfortunately, my understanding is that we cannot see any change until we commit to main because the Github Actions are taken from main.

fabrizio2210:
This is the stats of running 100 jobs on 2 nodes (all the jobs were scheduled on one node).

| Count | Context | Check | Command class |
| ---: | --- | --- | --- |
| 100 | `prolog` | `boot_disk_full` | `show node` |
| 100 | `prolog` | `alloc_mem_used` | `show job` |
| 3 | `hc_program` | `boot_disk_full` | `show node` |
| 3 | `hc_program` |  | `show node` |
| 1 | `hc_program` | `job_tmpfs_delete_leftover` | `listjobs` |

After the fix, same amount of jobs:

| Count | Context | Check | Command class |
| ---: | --- | --- | --- |
| 100 | `prolog` | `alloc_mem_used` | `show job` |
| 7 | `hc_program` |  | `show node` |
| 1 | `hc_program` | `job_tmpfs_delete_leftover` | `listjobs` |

It means 48% less RPC calls.

rdjjke (approved):

PR #2720, opened 2026-07-09
SCHED-1290 Run more checks in one slurm job for active checks

asteny:


helm/soperator-activechecks/values.yaml
```diff
@@ -54,60 +54,17 @@ slurmJob:
...
-        drainReasonPrefix: "[node_problem]"
-  all-reduce-perf-nccl-without-ib:
+  gpu-checks:
     enabled: true
     checkType: "slurmJob"
     dependsOn:
       - "manage-jail-state"
       - "wait-for-soperatorchecks-srun-ready"
       - "wait-for-topology"
-    schedule: "25 5-23/6 * * *"
```
theyoprst, line 107: No schedule anymore? As I understand it, by default it will be run once a year:

```go
	// +kubebuilder:default="0 0 1 1 *"
	Schedule string `json:"schedule,omitempty"`
```

Was it on purpose?

helm/soperator-activechecks/values.yaml
```diff
@@ -54,60 +54,17 @@ slurmJob:
...
-      - "manage-jail-state"
-      - "wait-for-soperatorchecks-srun-ready"
-      - "wait-for-topology"
-    schedule: "45 0-18/6 * * *"
-    suspend: false
-    runAfterCreation: true
-    slurmJobSpec:
-      sbatchScriptFile: "scripts/all-reduce-perf-nccl-with-ib.sh"
-      eachWorkerJobs: true
-      maxNumberOfJobs: 200
```
theyoprst, line 92: `maxNumberOfJobs: 200` was not added to gpu-checks, is it OK?

helm/soperator-activechecks/scripts/gpu-checks.sh
```diff
@@ -1,5 +1,5 @@
 #!/bin/bash
-#SBATCH --deadline="now+4hours"
+#SBATCH --deadline="now+12hours"
 #SBATCH --time=30:00
```
theyoprst: Shouldn't we also increase the time together with the deadline?

asteny: I decrease it. Dont think that this tests should work more then 1h

theyoprst (dismissed):

helm/soperator-activechecks/scripts/ib-gpu-perf.sh
```diff
@@ -25,8 +25,8 @@ echo "Platform found: $platform"
 echo "Running ib_gpu_perf check on $(hostname)..."
 HC_OUTPUT_DIR="/opt/soperator-outputs/health_checker_cmd_stdout"
 HC_OUTPUT=$(srun --container-image={{ include "activecheck.image.pyxis" . }} \
-  --container-mounts=$(which health-checker):/usr/local/bin/health-checker,$HC_OUTPUT_DIR:$HC_OUTPUT_DIR --cpu-bind=cores \
-  bash -l -c "health-checker run -e soperator -p $platform -n '^ib_write_bw_gpu.*$,^ib_send_lat_gpu.*$,^ib_read_lat_gpu.*$' -f json-partial --tests-stdout-path /opt/soperator-outputs/health_checker_cmd_stdout")
+  --container-mounts=$(which health-checker):/usr/local/bin/health-checker,$HC_OUTPUT_DIR:$HC_OUTPUT_DIR \
+  bash -l -c "health-checker run -e soperator_active_checks -p $platform -i '^ib_write_bw_gpu.*$,^ib_send_lat_gpu.*$,^ib_read_lat_gpu.*$' -f json-partial --tests-stdout-path /opt/soperator-outputs/health_checker_cmd_stdout --log-level info")
```
ChessProfessor, line 29: I'm thinking how can we test other active checks like ib_* with comment reaction later, because current design would require to add them to gpu-checks.sh exclusions + create separate active check.
Maybe we can decide on that in the health-checker? Having two namespaces soperator-stable-active-checks and soperator-unstable-active-checks and create two scripts here, one stable and one unstable. Not sure how good is that, just an idea. wdyt?

asteny: GPU team dont like idea to create more and more env for checks. I already created 4 new and started to use them. Also  ib_* in my next PR will be a stable check inside gpu_checks. In this PR I just exclude it from main scope and also I implemented -x flag in hell-checker to have an option to exclude checks. I cant say that making changes in a the hell checker scopes(envs) easier than changing our helms. It require PR, approval and release build and then we need to change release version in our checks

helm/soperator-activechecks/scripts/gpu-checks.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
-#SBATCH --deadline="now+4hours"
-#SBATCH --time=30:00
+#SBATCH --deadline="now+1hours"
```
rdjjke: Time limit 50min + deadline 1h means that if the job isn't able to start within 10 minutes after submission, it will be cancelled.
Shouldn't the deadline be higher?

helm/soperator-activechecks/scripts/gpu-checks.sh
```diff
@@ -22,8 +22,8 @@ else
 fi
 
 echo "Platform found: $platform"
-echo "Running all_reduce_perf_nccl check on $(hostname)..."
-HC_OUTPUT=$(srun --cpu-bind=cores sudo bash -l -c "health-checker run -e soperator -p $platform -n all_reduce_with_ib -f json-partial --tests-stdout-path /opt/soperator-outputs/health_checker_cmd_stdout --log-level info")
+echo "Running GPU checks on $(hostname)..."
+HC_OUTPUT=$(srun sudo bash -l -c "health-checker run -e soperator_active_checks -p $platform -x '^ib_write_bw_gpu.*$,^ib_send_lat_gpu.*$,^ib_read_lat_gpu.*$' -f json-partial --tests-stdout-path /opt/soperator-outputs/health_checker_cmd_stdout --log-level info")
```
rdjjke, line 26: AFAIK, we don't need to exclude `ib_*_gpu` checks from the main active check scope anymore. And there's no need to have the comment reaction for them, it's fine to drain nodes if they fail

asteny: I will do that in next PR (I have a task for that). Here I did exclusion because we definitely need such exclude in a feature. And I implemented this feature in hell-checker (exclusion by regexp, previously it works only with full names)

helm/soperator-activechecks/scripts/gpu-checks.sh
asteny: Oh, yes. Lets have time 50 and deadline 4hours than...

helm/soperator-activechecks/values.yaml
rdjjke: What about `0 0,12 * * *`?

```
US Eastern     | 19:00–20:00 | 07:00–08:00
US Pacific     | 16:00–17:00 | 04:00–05:00
Central Europe | 01:00–02:00 | 13:00–14:00
```

rdjjke (approved):

PR #2722, opened 2026-07-09
SCHED-1898: Speed up Exporter's Slurm metrics collection

dstaroff:
## Problem

Slurm clusters can make exporter collection lag when one slow sub-collector blocks the whole collection cycle. There was also no exporter metric showing which collector was slow or how long each collection took.

## Solution

- Added per-collector latency and in-flight metrics
- Changed collection to run sub-collectors independently on the configured interval
- Collector overlap is now configurable with `maxCollectorInflight` / `SLURM_EXPORTER_MAX_COLLECTOR_INFLIGHT`, defaulting to `1` to avoid extra Slurm API load unless explicitly tuned

## Testing

- Added and updated unit tests for collector latency metrics, async collection behavior, in-flight limits, stale result handling, and CLI/env parsing
- Added Helm tests for rendering `maxCollectorInflight` into the SlurmCluster CR
- Tested the behaviour on a dev cluster

## Release Notes

Feature: Slurm exporter now exposes per-collector latency/in-flight metrics and can run bounded asynchronous collection to keep faster metrics fresher when one collector is slow.
Note: `maxCollectorInflight` defaults to `1`; raising it can increase load on `slurmrestd` and Slurm controllers.


internal/exporter/collector.go
```diff
@@ -333,6 +335,8 @@ func (c *MetricsCollector) Describe(ch chan<- *prometheus.Desc) {
 func (c *MetricsCollector) updateState(ctx context.Context) (err error) {
```
theyoprst: This is dead code now in production, and the new path is not tested; maybe that's why the dead code still exists here. There is a plan from Fable how to fix that, please consider it:

> # Remove legacy sequential collection path from the Slurm exporter (PR #2722 follow-up commit)
> 
> ## Context
> 
> PR #2722 (branch `o11y/SCHED-1898`, checked out here) made exporter sub-collectors run
> asynchronously via `refreshNodes/refreshJobs/refreshDiag`, but left the old sequential path —
> `MetricsCollector.updateState` + `collectNodes/collectJobs/collectDiag` — in place. That path is
> now dead production code kept alive only by ~12 test call sites, so the well-tested path is the
> retired one while the shipping path is barely tested. Goal: delete the dead path, re-point tests
> at the production path, and close the related coverage/doc gaps. Lands as an additional commit on
> this PR branch. Design decision (user-approved, "Option A"): per-collector monitoring accounting
> moves *into* `refresh*` so existing collector tests exercise production accounting.
> 
> Facts this plan relies on (verified):
> - `collect{Nodes,Jobs,Diag}` are called only by `updateState`; `updateState` only by tests.
> - Only 2 test bodies depend on monitoring metrics `updateState` records itself:
>   `TestMetricsCollector_CollectorIsolation` (collector_test.go:376, `collector_errors_total`) and
>   `TestMetricsCollector_WithMonitoringMetrics` (:1159-1160, `collection_attempts/failures_total`).
>   All these metrics already have unit coverage in `monitoring_test.go`.
> - No test calls `collect*`/`list*`/`applyNodes`/`cloneMetricsCollectorState` directly.
> - The PR silently changed failure semantics from "metrics gap" to "last-good data preserved
>   (stale)"; `docs/slurm-exporter.md:139` still documents the old "gap" behavior.
> 
> ## Changes
> 
> ### 1. `internal/exporter/collector.go` — production
> 
> - Delete `updateState` (~:334-391), `collectNodes` (~:393-401), `collectJobs` (~:463-471),
>   `collectDiag` (~:509-517). Keep `listNodes`/`listJobs`/`getDiag`/`applyNodes`/
>   `cloneMetricsCollectorState` — they are the shared leaves `refresh*` uses.
> - Move per-collector accounting into `refresh*` via one small shared helper (named-return `err`
>   in each `refresh*`, `defer`red):
> 
>   ```go
>   // recordCollectorRun records duration always and an error only when the run
>   // was not cancelled, matching the loop's previous shutdown suppression.
>   func (c *MetricsCollector) recordCollectorRun(ctx context.Context, name string, start time.Time, err error) {
>       if err != nil && ctx.Err() == nil {
>           c.Monitoring.RecordCollectorError(name)
>       }
>       c.Monitoring.RecordCollectorDuration(name, time.Since(start).Seconds())
>   }
>   ```
> 
>   Collector names stay the string literals `"nodes"`/`"jobs"`/`"diag"` already used in
>   exporter.go's collector list.
> - Unify merge shape across all three `refresh*` to the `refreshNodes` order (check sequence
>   *before* cloning, avoiding a wasted clone on dropped runs): lock → load state (nil → fresh) →
>   stale check → clone → apply → store. Rewrite `refreshJobs`/`refreshDiag` accordingly.
> - Sequence-0 hardening: assign `newState.<x>CollectionSequence = sequence` only when
>   `sequence != 0`, so the synchronous test helper can never reset a real stored sequence.
> - The deleted isolation comment (collector.go:365-367) is replaced by an updated one above the
>   `refresh*` group (or the collectors list in exporter.go) stating the *new* semantics: a failed
>   collector keeps exporting its last successful data (stale) until it recovers; errors counted
>   via `slurm_exporter_collector_errors_total`.
> 
> ### 2. `internal/exporter/exporter.go` — loop simplification
> 
> - In `startCollectors`'s per-collector goroutine: drop `RecordCollectorError` and
>   `RecordCollectorDuration` calls (now inside `refresh*`); keep the `loopCtx.Err() == nil`-guarded
>   error log and `failed.Store(true)`, keep inflight/finish accounting. This also removes the inner
>   `start` variable, fixing the shadowing of the cycle-level `start` (:148 vs :170).
> - All-skipped fix: replace `if !started { e.monitoringMetrics.RecordCollection(0, nil); return }`
>   with a bare `return` — a cycle that ran nothing must not register as a successful zero-duration
>   collection; `slurm_exporter_collector_skipped_total` already accounts for it.
> 
> ### 3. `internal/exporter/collector_test.go` — migration
> 
> - Add a test-only helper and re-point all 12 `collector.updateState(ctx)` call sites to it:
> 
>   ```go
>   // collectOnce synchronously runs all sub-collectors once, bypassing sequence ordering.
>   func collectOnce(ctx context.Context, c *MetricsCollector) error {
>       return errors.Join(c.refreshNodes(ctx, 0), c.refreshJobs(ctx, 0), c.refreshDiag(ctx, 0))
>   }
>   ```
> 
>   Same nodes→jobs→diag order preserves `.Once()` mock sequencing; no goroutines, so the
>   `synctest.Test` (fake-clock) tests remain valid; `errors.Join` preserves `assert.Error` sites.
> - `TestMetricsCollector_CollectorIsolation`: survives as-is (error counter now incremented inside
>   `refresh*`). Its "failed collector's metrics absent" assertion is only true on a cold collector —
>   add a clarifying comment.
> - `TestMetricsCollector_WithMonitoringMetrics`: remove the cycle-level
>   `collection_attempts_total`/`collection_failures_total` assertions (:1159-1160) — cycle
>   accounting now happens only in the loop (asserted in exporter_test.go, see below). Repurpose to
>   assert per-collector production accounting instead: `collector_errors_total{nodes} == 1` after
>   the failing phase; keep the `metrics_exported`/`metricsCount` assertions.
> - `TestMetricsCollector_NodeFails` second phase (ListNodes fails, :500-513): review its
>   assertions under the new preserve-not-gap semantics; node data from phase 1 is now retained
>   after the failed phase. Adjust expectations if they encoded the gap behavior.
> - New tests:
>   - `refreshNodes`/`refreshDiag` stale-drop tests mirroring
>     `TestMetricsCollector_RefreshJobsDropsStaleSequence` (:32) — real sequences (2 then 1),
>     assert seq-2 data wins. The nodes one matters most: its stale guard protects the time-integral
>     `updateGPUSecondsMetrics` and the `updateNodeStateMetrics` state machine from negative-delta /
>     double-count corruption under overlap.
>   - Preserve-on-failure test: warm collector via `collectOnce`, then fail one collector (e.g.
>     jobs), assert its previous data is still in state/exported and `collector_errors_total`
>     incremented — pins the new stale semantics.
> 
> ### 4. `internal/exporter/exporter_test.go` — cycle-level accounting
> 
> - In an existing loop test (after `close(releaseJobs)`), add a loose `require.Eventually`
>   assertion that `slurm_exporter_collection_attempts_total >= 1` — the loop is now the only place
>   `RecordCollection` fires. Use `>=`, never exact counts, to avoid timing flakes.
> 
> ### 5. `docs/slurm-exporter.md`
> 
> - Rewrite the `slurm_exporter_collector_errors_total` row (:139): replace "the failed collector's
>   metrics simply gap until it recovers" with the new semantics — the failed collector keeps
>   exporting its last successfully collected data (stale) until it recovers.
> 
> ## Out of scope
> 
> - The two `exporter_test.go` timing tests' exact-count assertions (`jobsCalls == 2`) — noted as a
>   potential flake source in review, but not touched here.
> - Pushing the branch — commit locally on `o11y/SCHED-1898`; the user decides when/how to push to
>   the PR (branch is authored by dstaroff).
> 
> ## Verification
> 
> 1. `go build ./...` and `go vet ./internal/exporter/... ./cmd/exporter/...`
> 2. `grep -rn "updateState\|collectNodes\|collectJobs\|collectDiag" internal/ cmd/` → zero matches.
> 3. `go test ./internal/exporter/... ./cmd/exporter/... -race -count=5` — race detector plus
>    repetition to shake out timing flakes in the loop tests.
> 4. Sanity-check the monitoring surface didn't regress: run
>    `go test ./internal/exporter/ -run 'Monitoring|CollectorIsolation|RefreshNodes|RefreshJobs|RefreshDiag' -race -v`.
> 5. Commit on this branch in repo style, e.g. `[REFACTOR] Remove legacy sequential exporter
>    collection path` (Jira ID stays out of code, commit message may reference SCHED-1898).
>

dstaroff: done

theyoprst (approved):

PR #2723, opened 2026-07-09
Update NCCL to 2.30.7-1

ali-sattari:

### DIFF
--- a/helm/soperator-activechecks/values.yaml
+++ b/helm/soperator-activechecks/values.yaml
@@ -47,7 +47,7 @@ images:
   activeCheckImageTags:
     # Image training_diag from PR https://github.com/nebius/ml-containers/pull/88
     "12.9.0": "12.9.0-ubuntu24.04-20260605082528"
-    "13.0.2": "13.0.2-ubuntu24.04-20260605082528"
+    "13.0.2": "13.0.2-ubuntu24.04-20260709140028"
 # Optional override in Pyxis ("reg#repo:tag") format.
 activeCheckImage: ""
 slurmJob:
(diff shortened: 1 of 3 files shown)


helm/soperator-activechecks/values.yaml
```diff
@@ -47,7 +47,7 @@ images:
   activeCheckImageTags:
     # Image training_diag from PR https://github.com/nebius/ml-containers/pull/88
```
asteny, line 48: Can you put here PR number?

images/jail/jail.dockerfile
```diff
@@ -3,7 +3,7 @@
 ARG CUDA_VERSION
 ARG SLURM_VERSION
 # https://github.com/nebius/ml-containers/pull/88
```
asteny: same PR number

images/sansible/sansible.dockerfile
```diff
@@ -1,7 +1,7 @@
 # syntax=docker.io/docker/dockerfile-upstream:1.20.0
 
 # https://github.com/nebius/ml-containers/pull/88
```
asteny: same PR number

asteny (dismissed):

helm/soperator-activechecks/values.yaml
```diff
@@ -47,7 +47,8 @@ images:
...
```
asteny, line 48: Remove this?)

ali-sattari: That one is relevant for CUDA 12.9.0 image (which I haven't changed), right?

asteny: We need to change it also or remove CUDA12 complitely

asteny (approved):

PR #2735, opened 2026-07-13
SCHED-1893: Fix exporterContainer resources.

theyoprst:
Considered together with nebius/nebius-solutions-library#1073 (sizing tiers): that PR's terraform sends `exporterContainer.resources`, which this chart dropped until now.

## Problem

`exporterContainer` resources are not used in the Helm chart.

## Solution

Use them! The value is rendered into the CR verbatim, so keys are k8s resource names (`ephemeral-storage`, not `ephemeralStorage`).

## Testing

Add a Helm test.

## Release Notes

Fix: Exporter container resources were ignored

### DIFF
--- a/helm/slurm-cluster/templates/slurm-cluster-cr.yaml
+++ b/helm/slurm-cluster/templates/slurm-cluster-cr.yaml
@@ -473,6 +473,10 @@ spec:
         imagePullPolicy: {{ default "IfNotPresent" .Values.slurmNodes.exporter.imagePullPolicy | quote }}
         appArmorProfile: {{ default "unconfined" .Values.slurmNodes.exporter.appArmorProfile | quote }}
         {{- with (default dict .Values.slurmNodes.exporter.exporterContainer) }}
+        {{- with .resources }}
+        resources:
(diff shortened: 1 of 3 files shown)


helm/slurm-cluster/templates/slurm-cluster-cr.yaml
```diff
@@ -473,6 +473,12 @@ spec:
         imagePullPolicy: {{ default "IfNotPresent" .Values.slurmNodes.exporter.imagePullPolicy | quote }}
         appArmorProfile: {{ default "unconfined" .Values.slurmNodes.exporter.appArmorProfile | quote }}
         {{- with (default dict .Values.slurmNodes.exporter.exporterContainer) }}
+        {{- with .resources }}
```
Uburro, line 476:
```
{{- toYaml $resources | nindent 4 }}
{{- end }}
```

?

theyoprst: You are absolutely right! Fixed

Uburro (approved):

PR #2757, opened 2026-07-17
SCHED-1967: Add ActiveCheck initialization ownership switch

fabrizio2210:
## Context

[SCHED-1967](https://nebius.atlassian.net/browse/SCHED-1967) moves mandatory cluster initialization into the Slurm-cluster HelmRelease before ActiveChecks diagnostics are decoupled from Terraform apply. This compatibility-first PR prepares the ActiveChecks chart while preserving existing behavior by default.

## Changes

- add `clusterInitialization.managedBySlurmCluster`, disabled by default
- suppress legacy initialization ActiveChecks and their dependencies when enabled
- retain the force and dry-run jail maintenance checks
- add Helm rendering coverage for both ownership modes

## Validation

- `helm unittest helm/*/` (356 tests passed)
- default rendered manifests are unchanged from `HEAD`
- `git diff --check`

[SCHED-1967]: https://nebius.atlassian.net/browse/SCHED-1967?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


helm/soperator-activechecks/templates/active-checks.yaml
```diff
@@ -1,6 +1,9 @@
 {{- $root := . -}}
+{{- $managedBySlurmCluster := .Values.clusterInitialization.managedBySlurmCluster -}}
+{{- $legacyInitializationChecks := list "create-user-soperatorchecks" "create-user-nebius" "manage-jail-state" -}}
```
theyoprst: An LLM review:

> The list of "initialization checks" lives in the wrong file.
> The three check names are hardcoded as strings inside the template, while the checks themselves are defined in values.yaml. If someone renames one of those checks, or adds a new initialization check, they must remember to also update the hidden list in the template — nothing will remind them. The alternative: put a flag like initializationCheck: true on each of the three checks in values.yaml and have the template filter on that flag. Then the classification sits next to the check it describes.

If it will be removed in the next PR, maybe it's OK though.
But if there is a chance that this list will stay long, I would obey the LLM.
WDYT?

fabrizio2210: Good point. This compatibility logic will not be removed in the next PR; according to the migration plan it needs to remain until consumers have switched to Slurm-cluster-owned initialization. I agree that keeping the classification as a separate hard-coded list would be too easy to get out of sync. I’ll move the classification next to each check in `values.yaml` using an `initializationCheck` flag and use that metadata both for rendering and dependency filtering.

fabrizio2210:
I retire this change for lack of time to progress the whole project

PR #2771, opened 2026-07-22
SCHED-2184 Redesign enroot and docker acceptance tests

ChessProfessor:
## Problem

Docker and Enroot acceptance scenarios were based on long NCCL workloads and manual cache/runtime inspection flows. They were slow, fragile, hard to debug, and mixed several concerns together: container runtime lifecycle, image cache behavior, Slurm cancellation behavior, and GPU visibility.

## Solution

Simplified the container acceptance coverage into focused scenarios:

- Docker lifecycle now uses a small mirrored BusyBox image to verify Docker containers run under Slurm, populate local Docker storage, and are explicitly stopped after cancellation.
- 
- Enroot/Pyxis lifecycle now uses a mirrored Ubuntu image because Pyxis runs the Slurm task prolog inside the container and requires /bin/bash; the scenario verifies image artifact persistence, runtime cleanup, and cache reuse.

- GPU visibility is covered by separate Docker and Enroot smoke scenarios using the cluster-supported training_diag image and nvidia-smi -L.
- Removed the old shared hardcoded NCCL image constants and the obsolete NCCL-based scenario flow.
- Kept explicit Docker cleanup because dev-cluster testing confirmed Docker containers can remain running after scancel.

## Testing

E2E run - https://github.com/nebius/soperator/actions/runs/29919869911

### DIFF
--- a/internal/e2e/acceptance/steps/container_smoke.go
+++ b/internal/e2e/acceptance/steps/container_smoke.go
@@ -9,7 +9,14 @@ import (
 	"nebius.ai/slurm-operator/internal/e2e/acceptance/framework"
 )
 
-const containerSmokeSacctTimeout = 2 * time.Minute
+const (
+	containerSmokeSacctTimeout = 2 * time.Minute
+
+	// Cluster-supported GPU diagnostic image; includes nvidia-smi for container GPU visibility.
+	gpuSmokeImageRef    = "ml-containers/training_diag:13.0.2-ubuntu24.04-20260709140028"
+	gpuSmokeDockerImage = "cr.eu-north1.nebius.cloud/" + gpuSmokeImageRef
+	gpuSmokeEnrootImage = "docker://cr.eu-north1.nebius.cloud#" + gpuSmokeImageRef
+)
 
 func waitForJobSucceeded(ctx context.Context, exec framework.Exec, slurm *framework.SlurmClient, job framework.SbatchJob, timeout time.Duration) error {
 	if err := framework.AnnotateWithJobLog(ctx, exec, slurm, job, slurm.WaitForJobGone(ctx, job.ID, timeout)); err != nil {
--- a/internal/e2e/acceptance/steps/container_smoke.go
+++ b/internal/e2e/acceptance/steps/container_smoke.go
@@ -1,0 +1,88 @@
+package steps
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	"nebius.ai/slurm-operator/internal/e2e/acceptance/framework"
+)
+
+const containerSmokeSacctTimeout = 2 * time.Minute
+
+func waitForJobSucceeded(ctx context.Context, exec framework.Exec, slurm *framework.SlurmClient, job framework.SbatchJob, timeout time.Duration) error {
+	if err := framework.AnnotateWithJobLog(ctx, exec, slurm, job, slurm.WaitForJobGone(ctx, job.ID, timeout)); err != nil {
+		return err
+	}
+
+	waitCtx, cancel := context.WithTimeout(ctx, containerSmokeSacctTimeout)
+	defer cancel()
+
+	for {
+		if err := waitCtx.Err(); err != nil {
+			return framework.AnnotateWithJobLog(ctx, exec, slurm, job,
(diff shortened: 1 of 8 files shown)


internal/e2e/acceptance/steps/container_smoke.go
```diff
@@ -0,0 +1,95 @@
...
+	"time"
+
+	"nebius.ai/slurm-operator/internal/e2e/acceptance/framework"
+)
+
+const (
+	containerSmokeSacctTimeout = 2 * time.Minute
+
+	// Cluster-supported GPU diagnostic image; includes nvidia-smi for container GPU visibility.
+	gpuSmokeImageRef    = "ml-containers/training_diag:13.0.2-ubuntu24.04-20260709140028"
```
asteny, line 16: Lets maybe think how we can customize this image? For example we will support in a next month cuda 13, 14 and 15 and we need to test all this version

internal/e2e/acceptance/steps/docker_containers.go
```diff
@@ -92,33 +102,60 @@ func (s *DockerContainers) aLongRunningDockerNCCLJobIsSubmittedOnTwoGPUWorkers(c
...
+func (s *DockerContainers) dockerImageAndRuntimeStorageIsPopulatedOnAWorker(ctx context.Context) error {
+	if s.connectionWorker == "" {
+		return fmt.Errorf("Docker connection worker is not selected")
 	}
-	return framework.AnnotateWithJobLog(ctx, s.exec, s.slurm, s.job,
-		s.slurm.AssertJobRunning(ctx, s.job.ID))
-}
 
-func (s *DockerContainers) dockerOverlayfsStorageIsPopulatedOnAWorker(ctx context.Context) error {
-	return framework.WaitForTreeEntriesOnWorker(ctx, s.exec, s.connectionWorker, "/mnt/image-storage/docker/rootfs/overlayfs/", "Docker overlayfs storage", dockerProbeTimeout)
```
theyoprst, line 112: My LLM says that WaitForTreeEntriesOnWorker and TreeOutputHasEntries is not deadcode, worth to be deleted

internal/e2e/acceptance/steps/container_smoke.go
rdjjke: But probably not via this scenario

ChessProfessor: The best way I see is to somehow figure out the version used for jail image for e2e cluster and use the same here. Then we can run e2e on different clusters and different cudas and everything will be tested. But I'm not sure where to take this version on the cluster, it looks like it's hardcoded into the image too. Do you have any idea?

ChessProfessor: Before it was taken from prepull-container-image Active Check

theyoprst: An LLM suggestion:

```
1. Read the name off the worker:
sudo docker images --format '{{.Repository}}:{{.Tag}} {{.CreatedAt}}' | grep training_diag
1. Pick the newest line. That gives you cr.eu-north1.nebius.cloud/ml-containers/training_diag:<whatever-the-cluster-has>.
2. Use it for the Docker smoke directly.
3. Use it for the Enroot smoke by rewriting the same name into enroot's format — swapping the first / for # and adding docker:// — which is exactly the transformation the code already does today between gpuSmokeDockerImage and gpuSmokeEnrootImage.
4. Fall back to the current hardcoded constant if the grep finds nothing, so the test still runs on a cluster that hasn't cached it yet.
```

As I understand, this image can be large, so it's worth having generous timeouts for this Slurm job / whatever, so it will not fail often in e2e

ChessProfessor: But why would this training_diag image be in the docker images on the cluster? It was previously used in the prepull-container-image and active checks, but now they use binaries from the jail directly. training_diag is used only as base jail image, I don't think it is presented on the fresh cluster...

ChessProfessor: I believe hardcoding the image should work for now, and if we want to test different CUDAs we can have separate scenario for that later, let's keep current approach for now

theyoprst (approved):

PR #2773, opened 2026-07-23
SCHED-1897: Add idle node memory health check

fabrizio2210:
## Summary

- add the `idle_mem_used` periodic health check for `hc_program`
- determine whether the node has local jobs with RPC-free `scontrol listjobs`
- consume the authoritative Slurm `RealMemory` value exposed by #2774
- derive the maximum idle usage as `MemTotal - RealMemory`, without a separate Helm threshold
- drain idle nodes whose `MemAvailable` is below `RealMemory`, with an actionable diagnostic reason
- add shell behavior and Helm rendering coverage

## Why

This PR is related to the SCHED-1897 Jira ticket. Idle nodes retaining unexpectedly high memory usage may have leaked or spurious processes left behind by previous jobs. Detecting that condition periodically prevents affected nodes from silently returning to service.

Using local `scontrol listjobs` avoids a controller RPC for idle detection. Deriving the threshold from Slurm `RealMemory` keeps the check aligned with the worker memory configured by Soperator.

## Dependency

This PR is stacked on #2774, which publishes node `RealMemory` locally and makes `CHECKS_NODE_REAL_MEM_BYTES` RPC-free in the normal path. Merge #2774 first, then retarget this PR to `main`.

This replaces the feature diff from #2742. That PR was accidentally merged into #2743; #2774 replaces the detached #2743 pull ref and this PR carries only the dependent idle-memory check.

## Impact

The check is enabled by default and runs only through Slurm's `HealthCheckProgram`. Nodes with local Slurm jobs are left unchanged. Unexpected `listjobs` failures, unavailable memory data, missing `RealMemory`, and inconsistent values fail open.

On failure, the drain reason reports available memory, Slurm `RealMemory`, the deficit, current used memory, and the derived maximum idle usage.

## Validation

- `python3 -m unittest discover -s helm/slurm-cluster/slurm_scripts -p '*_test.py' -v` (15 tests passed)
- `bash -n helm/slurm-cluster/slurm_scripts/idle_mem_used.sh`
- parsed the check specification as JSON
- parsed `helm/slurm-cluster/tests/idle_mem_used_test.yaml` as YAML
- `git diff --check agent/rpc-free-node-real-memory...HEAD`
- confirmed the stable patch ID exactly matches the original #2742 feature diff


helm/slurm-cluster/slurm_scripts/idle_mem_used.sh
```diff
@@ -0,0 +1,91 @@
...
+    listjobs_rc=$?
+fi
+
+echo "scontrol listjobs exit code: ${listjobs_rc}"
+echo "scontrol listjobs output: ${listjobs_output:-<empty>}"
+
+if (( listjobs_rc == 0 )); then
+    node_is_idle=false
+    echo "Local Slurm jobs are present; treating the node as non-idle"
+elif (( listjobs_rc == 1 )) && [[ "${listjobs_output}" == *"No slurmstepd's found on this node"* ]]; then
```
fabrizio2210: When I manually tested, I got this RC and message. I don't know if it is brittle to check this, I was hoping to see no lines or something with RC=0.

rdjjke: What about using `scontrol listjobs --json` + `jq` to avoid parsing strings like "No slurmstepd's found on this node"?

helm/slurm-cluster/slurm_scripts/idle_mem_used.drain.sh
```diff
@@ -0,0 +1,91 @@
...
+mem_used_gb="$(awk -v bytes="${mem_used_bytes}" 'BEGIN { printf "%.2f", bytes / 1000000000 }')"
+node_real_memory_gb="$(awk -v bytes="${node_real_memory_bytes}" 'BEGIN { printf "%.2f", bytes / 1000000000 }')"
+max_idle_used_gb="$(awk -v bytes="${max_idle_used_bytes}" 'BEGIN { printf "%.2f", bytes / 1000000000 }')"
+
+echo "Memory source: ${meminfo_path}"
+echo "Memory measurements: total=${mem_total_gb} GB (${mem_total_bytes} bytes), available=${mem_available_gb} GB (${mem_available_bytes} bytes), used=total-available=${mem_used_gb} GB (${mem_used_bytes} bytes)"
+echo "Slurm RealMemory: ${node_real_memory_gb} GB (${node_real_memory_bytes} bytes)"
+echo "Derived maximum idle used memory: MemTotal-RealMemory=${max_idle_used_gb} GB (${max_idle_used_bytes} bytes)"
+
+if (( mem_available_bytes < node_real_memory_bytes )); then
```
rdjjke: `mem_available_bytes` is taken as it is from `/proc/meminfo` (except for multiplying by 1024).
`node_real_memory_bytes` is available in the env var.

Why do we perform the rest of the calculations (total, used, etc), only to print them?

If so, then what about using `free -hw` and printing its output directly?

helm/slurm-cluster/slurm_scripts/idle_mem_used.drain.sh
```diff
@@ -0,0 +1,91 @@
+#!/bin/bash
```
theyoprst: Some of our checks are already written in Python. Given that this check has Python tests, maybe write it in Python as well, and the testing part would be even easier?

helm/slurm-cluster/slurm_scripts/idle_mem_used.sh
```diff
@@ -0,0 +1,91 @@
...
+
+echo "Memory source: ${meminfo_path}"
+echo "Memory measurements: total=${mem_total_gb} GB (${mem_total_bytes} bytes), available=${mem_available_gb} GB (${mem_available_bytes} bytes), used=total-available=${mem_used_gb} GB (${mem_used_bytes} bytes)"
+echo "Slurm RealMemory: ${node_real_memory_gb} GB (${node_real_memory_bytes} bytes)"
+echo "Derived maximum idle used memory: MemTotal-RealMemory=${max_idle_used_gb} GB (${max_idle_used_bytes} bytes)"
+
+if (( mem_available_bytes < node_real_memory_bytes )); then
+    memory_deficit_bytes=$((node_real_memory_bytes - mem_available_bytes))
+    memory_deficit_gb="$(awk -v bytes="${memory_deficit_bytes}" 'BEGIN { printf "%.2f", bytes / 1000000000 }')"
+    echo "Node ${node_name} is IDLE but has only ${mem_available_gb} GB of memory available, below Slurm RealMemory ${node_real_memory_gb} GB by ${memory_deficit_gb} GB (used: ${mem_used_gb} GB; derived maximum idle usage: ${max_idle_used_gb} GB; MemTotal: ${mem_total_gb} GB). This may indicate leftover or spurious processes consuming memory." >&3
```
rdjjke: This message will be added to the drain reason visible to the end user.
I suggest making it shorter or printing the important information at the beginning, because in commands like `sinfo`, users will see only the first 20 characters.

For example, there is no need to have the node name here, since this message is already an attribute of a specific node. Also, it's better to start the message with lowercase, since it's appended after ": ".
And probably it doesn't make much sense to print used memory here, since in practice it will change, but the drain reason will have the value collected at the moment this check ran.

What about something like this?
"available memory < configured memory. Stop leftover processes or reboot"

helm/slurm-cluster/slurm_scripts/idle_mem_used.drain.py.json
```diff
@@ -0,0 +1,16 @@
+{
```
rdjjke, line 1: We need to make this check undrain nodes, or have a separate check that does undraining.
It will require additional RPCs, though (if it will be a single check that does both draining and undraining, it will do `scontrol update ...` on each pass, and if there will be 2 checks, the undraining one will need to get the node state to only start on drained nodes).
But at least this will be in hc_program, not prolog/epilog. And the problem with getting node states without RPCs can be solved later.

helm/slurm-cluster/slurm_scripts/idle_mem_used.drain.py.json
rdjjke, line 1: Let's also delete `alloc_mem_used` checks? This one is the RPC-less replacement for them.

helm/slurm-cluster/slurm_scripts/idle_mem_used.sh.json
```diff
@@ -0,0 +1,16 @@
...
+  "name": "idle_mem_used",
+  "command": "./idle_mem_used.sh",
+  "platforms": ["any"],
+  "skip_for_cpu_jobs": false,
+  "skip_for_partial_gpu_jobs": false,
+  "contexts": ["hc_program"],
+  "node_states": ["any"],
+  "on_fail": "drain",
+  "on_ok": "none",
+  "reason_base": "[node_problem] $name",
```
rdjjke: The reason the prefix should be `[user_problem]`. `[node_problem]` means that there is a hardware or software issue with the node, and the user can do nothing about it.
While `[user_problem]` means incorrect usage by the user, and it's their responsibility to fix the issue.

helm/slurm-cluster/slurm_scripts/idle_mem_used.sh
fabrizio2210: More elegant, thanks. I just changed the code

helm/slurm-cluster/slurm_scripts/idle_mem_used.drain.sh
fabrizio2210: Good point. `MemTotal` is also used to validate that `MemAvailable` and Slurm `RealMemory` are consistent, but the used-memory and maximum-idle-usage calculations are only diagnostic and are not needed for the decision. I’ll simplify the comparison to `MemAvailable < RealMemory`, keep the `MemTotal` sanity checks, print `free -hw` in the log, and shorten the drain reason to the two relevant values and the remediation.

helm/slurm-cluster/slurm_scripts/idle_mem_used.drain.py.json
fabrizio2210: I want to test this before deleting it.

helm/slurm-cluster/slurm_scripts/idle_mem_used.drain.py.json
fabrizio2210: Agreed that this check should recover nodes automatically. I propose a separate `hc_program` recovery check rather than setting `on_ok: undrain` on the current check. The current command deliberately exits successfully when the node is busy or memory information cannot be obtained, so using that success directly could undrain a node without confirming recovery.
The recovery check will run for drained nodes, use the same reason prefix, and return success only when there are no local jobs and available memory is enough

fabrizio2210: I uploaded the undrain script and renamed the current one.

helm/slurm-cluster/slurm_scripts/idle_mem_used.drain.sh
fabrizio2210: Python would simplify JSON parsing and enable more direct unit tests. However, most of the checks in this folder currently invoke shell scripts, and these Python tests deliberately exercise the shell command as a black box.
The scripts are mostly set of system commands.
I suggest keeping this check in shell for this PR and reconsidering Python if its logic grows.

helm/slurm-cluster/slurm_scripts/idle_mem_used.drain.py.json
```diff
@@ -0,0 +1,16 @@
...
+  "name": "idle_mem_used",
+  "command": "./idle_mem_used.drain.sh",
+  "platforms": ["any"],
+  "skip_for_cpu_jobs": false,
+  "skip_for_partial_gpu_jobs": false,
+  "contexts": ["hc_program"],
+  "node_states": ["any"],
+  "on_fail": "drain",
+  "on_ok": "none",
+  "reason_base": "[user_problem] $name",
```
fabrizio2210, line 11: @rdjjke We're introducing a new drain reason and it is fine. But what about the old `alloc_mem_used` drain reason? When we remove the old check we might have drained nodes which will never automatically be undrained.
So, we can avoid this by defining `alloc_mem_used` as a drain reason here, so we carry over the drain reason. Wht do you think? Do we have a lot of those nodes drained for this reason?

helm/slurm-cluster/slurm_scripts/idle_mem_used.drain.sh
theyoprst: I think it's a historical mistake that we wrote these checks in Bash. Bash scripts are getting more and more complex, and it's very hard for humans to deal with this, even though they are written by AI agents now.
E.g., in this script, there is JSON parsing and several other parsers and validators; to work with them, you have to be a bash+jq+awk+whatever tools are used. With Python, you only need to be a Python expert.

But it's up to you, we have a lot of bash already :-)

helm/slurm-cluster/slurm_scripts/idle_mem_used.drain.sh
```diff
@@ -0,0 +1,96 @@
...
+    echo "Could not determine whether the node is idle because 'scontrol listjobs --json' failed; skipping memory validation" >&2
+    exit 0
+fi
+
+if ! local_job_count="$(
+    jq -er '
+        if (.jobs | type) == "array" then
+            .jobs | length
+        else
+            error(".jobs must be an array")
```
theyoprst: What if the next version of Slurm returns not `[]`, but `null`? Or even skip the `job` key (like in Go by default)? Looks like this check will stop working, always exiting 0

rdjjke (approved):

PR #2782, opened 2026-07-24
SCHED-1308 Add health check acceptance tests + Get rid of cleanup_enroot and ssh checks

ChessProfessor:
## Problem

cleanup_enroot passive and active checks are not used anymore
The same for ssh active check
Also there are no acceptance health-check

## Solution

Removing cleanup_enroot passive and active checks and their scripts.
Also remove ssh active check with the script
Adding acceptance tests for active, passive and system checks


internal/e2e/acceptance/runner.go
```diff
@@ -219,6 +231,76 @@ func discoverNodeSets(ctx context.Context, w *world, clusterName string) ([]fram
...
+}
+
+func discoverGPUCounts(ctx context.Context, w *world, state *framework.ClusterState) error {
+	if len(state.GPUWorkers) == 0 {
+		return nil
+	}
+
+	counts := make(map[string]int, len(state.GPUWorkers))
+	for _, worker := range state.GPUWorkers {
+		out, err := w.Worker(worker).RunWithDefaultRetry(ctx, "nvidia-smi -L | wc -l")
```
theyoprst: > GPU-count discovery makes cluster discovery fragile (medium)
> 
> discoverGPUCounts (runner.go) now runs nvidia-smi -L | wc -l on every GPU worker during discovery, and parseDiscoveredGPUCount hard-fails when the count is ≤ 0, aborting the entire acceptance suite before any scenario runs — even CPU-only runs.
> 
> Two edges make this worse than it looks:
> - The pipe has no pipefail, so a failing/hanging-then-erroring nvidia-smi still exits 0 (from wc) and yields "0". RunWithDefaultRetry sees success and does not retry; discovery then hard-errors on "invalid GPU count". A transient GPU hiccup shouldn't fail the whole suite.
> - Previously GPU classification was purely nodeset-name based and required no per-node exec at all.
> 
> Suggest bash -lc 'set -o pipefail; nvidia-smi -L | wc -l' so a real nvidia-smi failure surfaces as an error and gets retried, and consider whether one bad GPU worker should abort discovery globally.

internal/e2e/acceptance/steps/check_helpers_test.go
```diff
@@ -0,0 +1,176 @@
...
+	assert.False(t, activeCheckSlurmJobRecordsTerminal(records, []string{"101", "102"}))
+
+	records["102"] = activeCheckSlurmJobRecord{ID: "102", State: "FAILED"}
+	assert.True(t, activeCheckSlurmJobRecordsTerminal(records, []string{"101", "102"}))
+}
+
+func TestAssertActiveGPUCheckOutputPassing(t *testing.T) {
+	require.NoError(t, assertActiveGPUCheckOutputPassing("Health checker status: PASS\n"))
+	require.Error(t, assertActiveGPUCheckOutputPassing("Health checker status: FAIL\n"))
+	require.Error(t, assertActiveGPUCheckOutputPassing("no health status here\n"))
```
theyoprst, line 128: > assertActiveGPUCheckOutputPassing is dead outside tests (minor)
> 
> active_checks.go:409 is referenced only from check_helpers_test.go; the production path (waitForGPUActiveCheckOutputsPassingOnAllGPUWorkers) inlines the equivalent activeGPUCheckOutputStatus logic instead. Either call the helper from the production path (DRY) or drop it.

ChessProfessor: Now it's used

internal/e2e/acceptance/steps/passive_checks.go
```diff
@@ -0,0 +1,766 @@
...
+	targets := []string{
+		"/mnt/memory/job_" + s.tmpfsJobID,
+		"/mnt/jail/mnt/memory/job_" + s.tmpfsJobID,
+	}
+	if err := s.exec.WaitFor(ctx, "job tmpfs directory removed", passiveTmpfsCleanupTimeout, framework.DefaultPollInterval, func(waitCtx context.Context) (bool, error) {
+		var checks []string
+		for _, target := range targets {
+			checks = append(checks, fmt.Sprintf("test ! -e %s", framework.ShellQuote(target)))
+		}
+		_, err := s.exec.Worker(s.worker).RunWithDefaultRetry(waitCtx, strings.Join(checks, " && "))
```
theyoprst: > passive_checks.go:433 — the "wait for the tmpfs dir to be gone" loop still uses the auto-retrying runner on a command that's meant to fail at first (test ! -e ...), then folds any error into ready = err == nil. So each poll burns extra retry time and a genuine failure surfaces as a vague timeout. Switching that probe to the plain (no-retry) runner would make it snappier and clearer. (Same shape in the system-checks "worker present again" loop.)

internal/e2e/acceptance/features/cluster_creation.feature
```diff
@@ -6,7 +6,10 @@ Feature: Cluster creation
     And all NodeSet CRs are ready
     And main and hidden partitions are present and sane
     And all Slurm nodes are healthy
+    And HealthCheckProgram outputs are healthy
     And all ActiveChecks completed successfully
```
rdjjke, line 10: What about removing this step? We now have a separate scenario for testing active checks

ChessProfessor: But that's more about cluster creation active checks, the ones with RunAfterCreation: true
Active check scenarios doesn't cover that, and this step is very cheap. I would keep it.

theyoprst (dismissed):

theyoprst (approved):

PR #2783, opened 2026-07-24
SCHED-2200: Bump golangci-lint to v2.12.2

theyoprst:
## Problem

The Go 1.26 migration (#2784) cannot pass lint while main pins golangci-lint v2.5.0: the lint job runs on `pull_request_target`, so the workflow definition comes from the PR base branch, and v2.5.0 is built with go1.25 and refuses go 1.26 modules.

## Solution

- bump golangci-lint to v2.12.2 in the workflow and Makefile while the module still targets Go 1.25; both linter versions pass on this branch
- fix govet inline-analyzer findings v2.12.2 reports: `reflect.Ptr` -> `reflect.Pointer`, `constraints.Ordered` -> `cmp.Ordered`, `x/exp/slices` -> stdlib `slices`

#2784 is stacked on this branch, so its lint runs with v2.12.2.

## Testing

Linted with v2.5.0 (what this PR's own CI runs) and v2.12.2 (what main runs after merge): 0 issues both. Unit suite green.

## Release Notes

None


theyoprst:
Replaced by a re-stacked PR: GitHub's stack feature does not allow changing the base branch, and this PR needs to sit on top of #2788 (golangci-lint bump) rather than main. Same branch, same content; see the new PR.

PR #2784, opened 2026-07-24
SCHED-2200: Update k8s libraries to v0.36 and controller-runtime to v0.24

theyoprst:
## Problem

Kubernetes libraries are at v0.34 and controller-runtime at v0.22, two releases behind. controller-runtime v0.23 changed the webhook builder API and moved event recording to the `events.k8s.io` API group, so the bump needs code and RBAC changes. Both target versions require Go 1.26 (previous PR in the stack, #2789).

## Solution

- bump `k8s.io/api`, `apimachinery`, `client-go` to v0.36.3 and `controller-runtime` to v0.24.1
- migrate the three admission webhooks to the generic typed builder and `admission.Validator[T]`, dropping the `runtime.Object` type assertions
- add `events.k8s.io` to events RBAC (kubebuilder markers, leader election role, hand-maintained helm leader-election template); the manager records events via `events/v1` now
- controller-tools v0.19.0 -> v0.21.0, setup-envtest release-0.24, wire the previously unused `ENVTEST_VERSION` variable into the install rule
- regenerate CRDs, RBAC, webhook manifests and helm output; CRD churn is the k8s 1.34 -> 1.36 schema refresh, the webhook manifest diff is reordering only

`Result{Requeue: true}` is deprecated in v0.24 but still honored; migrating it to `RequeueAfter` changes backoff behavior and is left for a follow-up.

## Testing

- Full unit suite (fake client SSA changes did not break tests), `make lint`, `make sync-version-from-scratch UNSTABLE=false` idempotent.
- E2E: https://github.com/nebius/soperator/actions/runs/30257091861

## Release Notes

None


helm/soperator/templates/manager-rbac.yaml
```diff
@@ -38,6 +31,14 @@ rules:
   - get
   - list
   - watch
+- apiGroups:
+  - ""
+  - events.k8s.io
+  resources:
+  - events
```
Uburro, line 38: does the soperator create any events?

Uburro (approved):

helm/soperator/templates/manager-rbac.yaml
theyoprst: Yes, LeaderElection ones:

```
❯ kubectl --context e2e-test get events -A  | grep 'nebius.ai\s'
soperator-system                    35m         Normal    LeaderElection            lease/a22156dp.nebius.ai                                                               soperator-checks-checks-5d9d9c457-jh7qk_a2a905f9-6515-4577-9987-396e9561a6b6 became leader
soperator-system                    36m         Normal    LeaderElection            lease/e21479ae.nebius.ai                                                               soperator-controller-manager-595bd7b969-bd4hv_732a5929-f6ac-4f45-acd0-43e4edd2bb43 became leader
soperator                           29m         Normal    LeaderElection            lease/soperator.vqeyz6ae.nebius.ai                                                     soperator-sconfigcontroller-7666b7c99d-fltth_14e3ca6d-e97b-4b9c-bad9-6161bbc17662 became leader
```

They are published by controller-runtime via the recorder we provide.

The RBAC change is required because of the controller-runtime v0.23+ bump: the manager's event recorder now posts to /apis/events.k8s.io/v1/.../events instead of /api/v1/.../events. Before the bump, calling it would have just worked under the existing core-group rule. After the bump, it would fail RBAC.

PR #2791, opened 2026-07-27
Allow set extra labels/annotations

andriishestakov:

### DIFF
--- a/api/v1/slurmcluster_types.go
+++ b/api/v1/slurmcluster_types.go
@@ -58,6 +58,16 @@ type SlurmClusterSpec struct {
 	// +kubebuilder:validation:Required
 	SlurmNodes SlurmNodes `json:"slurmNodes"`
 
+	// ExtraLabels are custom K8s labels added to every Pod and the spool PVC of this cluster.
+	//
+	// +kubebuilder:validation:Optional
+	ExtraLabels map[string]string `json:"extraLabels,omitempty"`
+
+	// ExtraAnnotations are custom K8s annotations added to every Pod and the spool PVC of this cluster.
+	//
+	// +kubebuilder:validation:Optional
+	ExtraAnnotations map[string]string `json:"extraAnnotations,omitempty"`
+
 	// PartitionConfiguration define partition configuration of slurm worker nodes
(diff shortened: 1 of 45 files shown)


internal/render/exporter/pod.go
```diff
@@ -23,9 +25,13 @@ func renderPodTemplateSpec(
 		_ = err // Ignore not found error, use "empty" node filter.
 		nodeFilter = slurmv1.K8sNodeFilter{}
 	}
+	labels := maps.Clone(matchLabels)
+	maps.Copy(labels, clusterValues.SlurmExporter.Labels)
```
Uburro: User labels can hijack selector labels and permanently wedge the StatefulSet. Every render does maps.Copy(labels, X.Labels) after building the operator's labels, so user keys win - but the selector is built separately from matchLabels, which is not updated. Setting extraLabels: {app.kubernetes.io/component: foo} (or name/instance, or slurm.nebius.ai/nodeset) makes the pod template diverge from the immutable selector.

api/v1/slurmcluster_types.go
```diff
@@ -58,6 +58,16 @@ type SlurmClusterSpec struct {
 	// +kubebuilder:validation:Required
 	SlurmNodes SlurmNodes `json:"slurmNodes"`
 
+	// ExtraLabels are custom K8s labels added to every Pod and the spool PVC of this cluster.
```
Uburro: The spool PVC doc comment doesn't match the behavior. Both CRD fields say "added to every Pod and the spool PVC of this cluster", but nothing in the operator labels a PVC: [common/volume.go:46-60](nternal/render/common/volume.go#L46-L60) (the VolumeClaimTemplates, including the worker spool) is untouched. The only PVC labeling is chart-side in [pvc.yaml](helm/slurm-cluster/templates/pvc.yaml), which iterates .Values.volumeSources with createPVC — jail and friends, not spool, and driven by chart values rather than the CR field. If the use case is billing attribution, unlabeled operator-created PVCs are probably the most valuable remaining gap; either way the comment should be corrected.

Uburro: ActiveCheck can use spool PVC. It needs to be rewritten.

PR #2792, opened 2026-07-27
SCHED-2106 mount nvidia-imex socket to the jail from the node

asteny:

### DIFF
--- a/internal/render/worker/statefulset_test.go
+++ b/internal/render/worker/statefulset_test.go
@@ -320,10 +320,10 @@ func TestRenderNodeSetStatefulSet_NvidiaIMEXCLIMount(t *testing.T) {
 	assertHostPathVolume(t, result.Spec.Template.Spec.Volumes, consts.VolumeNameNvidiaIMEXConfigJail)
 
 	slurmd := result.Spec.Template.Spec.Containers[0]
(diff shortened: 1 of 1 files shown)


theyoprst (dismissed):

internal/consts/nvidia.go
```diff
@@ -10,4 +10,6 @@ const (
 	NvidiaIMEXConfigHostPath  = "/etc/nvidia-imex"
 	NvidiaIMEXConfigMountPath = NvidiaIMEXConfigHostPath
 	NvidiaIMEXConfigJailPath  = VolumeMountPathJailUpper + NvidiaIMEXConfigHostPath
+
+	NvidiaIMEXSocketJailDirPath = VolumeMountPathJailUpper + "/var/run/nvidia"
```
rdjjke: What about replacing it with a clearer `/run/nvidia`? (`/var/run` is a symlink to `/run`)

rdjjke: We already mount something to `/run/nvidia/driver` directory (a subdirectory inside your new mount `/run/nvidia`). And in driverless, it's done automatically by the GPU operator, so even if we get rid of our mount, it won't help with driverless.
Can you make sure that `/run/nvidia/driver` is mounted after `/run/nvidia` so the second mount doesn't hide the first?

PR #2807, opened 2026-07-29
Don't undrain Slurm nodes with non-existing K8s nodes

rdjjke:
## Problem
During node replacement, Soperator deletes K8s nodes under Slurm nodes drained with the `[node_problem]` prefix.
Then, it undrains the Slurm node if the K8s node doesn't exist.
If the Slurm node gets undrained before the Slurm timeout passes and the node becomes NOT_RESPONDING (and then down), then there is a period of time when the node is IDLE in Slurm and can still accept new jobs, which fail.

## Solution
Wait for Slurm nodes to become DOWN before undraining them.

Also, bump the Soperator version to make a patch release.

## Testing
Tested on a dev cluster

## Release Notes
Fix the bug when Slurm nodes were undrained before becoming DOWN during node replacement.


Uburro (dismissed):

Uburro (dismissed):

internal/controller/soperatorchecks/slurm_nodes_controller.go
```diff
@@ -178,7 +178,7 @@ func (c *SlurmNodesController) processDegradedNode(
 	k8sNode, err := getK8SNode(ctx, c.Client, node.InstanceID)
 	if err != nil {
 		if client.IgnoreNotFound(err) == nil {
-			return c.undrainSlurmNode(ctx, slurmClusterName, node.Name)
+			return nil
 		}
 		return fmt.Errorf("get k8s node: %w", err)
```
Uburro:
```
if err != nil {
    return client.IgnoreNotFound(fmt.Errorf("get k8s node: %w", err))
}
```

asteny (approved):

PR #2825, opened 2026-08-03
SCHED-2206 Make E2E branch agnostic

ChessProfessor:
## Problem

Soperator acceptance tests were tied to the checked-out branch and could not reliably select scenarios based on the actual target cluster version. This blocks reuse of the same open-source acceptance suite against existing/dev clusters and future external Soperator runners.

## Solution

Refactored acceptance into reusable E2E packages with a standalone acceptance binary, explicit `--soperator-version` support, and scenario filtering by `@soperator_version_>=x.y.z` tags. Updated the E2E workflow to pass the discovered target Soperator version into acceptance and removed the old cmd/e2e acceptance wrapper.

## Testing

E2e passes - https://github.com/nebius/soperator/actions/runs/30807350235


theyoprst:
## Code review

### Overview

Decouples the acceptance suite from the checked-out branch so it can run against arbitrary target clusters. Two logically distinct changes are bundled:

1. **Feature:** new `e2e/versionfilter` package that pre-filters Gherkin scenarios by `@soperator_version_>=x.y.z` tags, a standalone `bin/acceptance` binary with an explicit `--soperator-version` flag (falling back to Flux `HelmRelease` discovery), and a package reshuffle (`cmd/e2e acceptance` wrapper removed; CLI moved to `e2e/cmd/acceptance/cli`; runner now requires a target version).
2. **Mechanical:** repo-wide `4.1.7 → 4.2.0` bump (VERSION, `internal/consts`, all Helm charts/values, config kustomizations, FluxCD envs).

Well-structured, well-tested work. Findings below are mostly minor.

### Strengths

- **Clean layering.** `versionfilter` is provider-agnostic (`Axis{TagPrefix, TargetVersion}`), `framework.NormalizeSoperatorVersion` is the Soperator-specific wrapper, CLI concerns live in `cmd/acceptance/cli`. No import cycles.
- **Prerelease handling is correct by design.** Stripping the suffix (`4.1.5-reb85d0e5 → 4.1.5`) before constraint checking sidesteps Masterminds semver's prerelease-exclusion behavior, since the normalized target is always a clean release. Right call.
- **Dependency hygiene.** `semver`, `gherkin`, `messages` correctly moved from `// indirect` to direct requires.
- **Good test coverage.** `fstest.MapFS` fixtures cover AND/OR (`>=4.0.0,<4.1.0||>=4.2.0`), exact-line selection, tag-line/body-line rejection, missing/duplicate/short tags, and multi-axis. `TestSharedFeaturesHaveValidVersionTags` runs the filter over all 11 real feature files, so a malformed or missing version tag on any scenario fails CI — nice guardrail.
- **Actionable errors,** e.g. `discover target Soperator version from Flux HelmRelease: …; pass --soperator-version to override`.

### Potential issues

**1. `framework.ParseSoperatorBaseVersion` is dead code (low).**
Defined in `e2e/acceptance/framework/version.go` but referenced only by its own test. No production path calls it — normalization goes through `NormalizeSoperatorVersion`. Recommend removing it (and its test) until there's a caller.

**2. Flux discovery reads `status.lastAttemptedRevision`, not `lastAppliedRevision` (low).**
`lastAttemptedRevision` is the version Flux last *tried* to reconcile, which can diverge from what's actually running if an upgrade is mid-flight or failed. For "what version is this cluster actually running," `lastAppliedRevision` is more accurate. Low severity: this is only the fallback path (CI always passes `--soperator-version` explicitly) and on a settled cluster the two agree. Worth a comment explaining the deliberate choice, or switching field.

**3. Baseline-tagging every scenario `>=4.2.0` partially undercuts the stated goal (design note).**
The motivation is reuse "against existing/dev clusters," but every current scenario is tagged `@soperator_version_>=4.2.0`. Running against any pre-4.2.0 cluster filters out *all* scenarios and the runner errors (`no acceptance scenarios compatible with Soperator version <X>`). Internally consistent and documented, but the branch-agnostic suite only works forward from 4.2.0. Note the filter is effectively a **no-op today** (all tags equal the deployed version); its value is future infrastructure. Both fine — worth stating explicitly that this is baseline scaffolding.

**4. Multi-axis generality is speculative (minor / YAGNI).**
`SelectScenarios(source, axes ...Axis)` supports multiple version families but only one exists. Clean and tested, so low cost — but the design is all-or-nothing: adding a second axis later would require the new tag on *every* existing scenario or the whole run errors. Acceptable for a controlled internal suite; flagging so it's a conscious choice.

**5. Minor redundancy in `versionfilter.go`.**
- `validateVersionConstraintExpression` runs at parse time *and* again inside `versionConstraintMatches`. Intentional fail-fast (validates tags on scenarios you aren't running) — keep it, but a short comment saying so would help.
- `semverConstraintExpression` rewrites `,` → ` ` before Masterminds, but Masterminds accepts comma-separated AND directly. Harmless but unnecessary.

### Conventions & style

- Error-message casing follows the project rule: lowercase starts except where beginning with the proper nouns `Soperator`/`Flux`/`FS`. ✅
- `log.Printf("acceptance: …")` lowercase prefix matches the pre-existing logging style in this file — not a regression.
- `var`-style slice declarations used throughout. ✅

### Test coverage

Strong for the filter and CLI parsing. Two gaps worth considering:
- No test asserts the `=`/bare-version constraint forms that `constraintPattern` permits (e.g. `@soperator_version_4.2.0`). If bare/`=` isn't a supported author form, consider rejecting it in the pattern; if it is, add a case.
- `discoverFluxSoperatorVersion`'s kubectl exec path is untested (understandable); the parse helper is well covered.

### Security

Nothing concerning. `kubectl` is invoked via `exec.CommandContext` with an argv slice (no shell), a 1-minute timeout, and a required operator-supplied `--context`. No injection surface, no secrets handled.

### Bottom line

Solid, cleanly separated refactor with good tests. No blocking correctness issues. Before merge: **(1)** drop the unused `ParseSoperatorBaseVersion`, and **(2)** decide `lastAttemptedRevision` vs `lastAppliedRevision` and leave a comment either way. Consider a sentence in the PR body noting the filter is forward-looking scaffolding and that the `>=4.2.0` tags require the deployed build to be 4.2.0 for any scenario to run — that dependency is why the version bump is bundled here.

ChessProfessor:
Done:
- removed unused ParseSoperatorBaseVersion
- simplified versionfilter constraint handling
- documented explicit @soperator_version_>=5.0.0 tags in README
- changed Soperator baseline from 4.2.0 to 5.0.0
- updated all E2E scenario tags to @soperator_version_>=5.0.0
- changed Flux discovery to use lastAppliedRevision, falling back to lastAttemptedRevision. This path is used for manual acceptance on a dev cluster, and for my dev cluster lastAppliedRevision is empty, so decided to have fallback.

Decided not to do:
- no 4.0/4.1 scenario compatibility adjustments for now, if we need new scenarios to run in 4.0 or 4.1 we can adjust it later
- no changes to multi-axis filtering behavior, that's intended
- no changes to bare / exact version tag support, probably won't be used but we can keep it for simplicity
- no extra kubectl exec-path tests now, overcomplicated to add

theyoprst (approved):

PR #2827, opened 2026-08-03
Decouple webhooks and certManager in helm

andriishestakov:
## Problem
when cert-manager disabled, webhooks are created in k8s but not handled in soperator.

## Solution
Decouple settings for enable webhooks and cert-manager to allow use webhooks without cert-manager


Uburro:
@andriishestakov can u rebase it ?

Uburro (approved):

PR #2833, opened 2026-08-04
SCHED-2207 refactor public acceptance API

ChessProfessor:
## Problem

The acceptance E2E runtime was still shaped around the open-source repository layout, with reusable step implementations and helper objects exposed as regular public packages.
This made the public API unclear for external runners: it was not obvious what should be imported directly, what should stay framework-internal, or how custom/private runners should combine shared Soperator scenarios with their own feature files, tag/version filtering, and discovery logic.

## Solution

Refactored the acceptance runner API around explicit suite configuration, shared step registration, version axes, and discovery hooks.

Main changes:
- Added SuiteConfig-based scenario registration. Each suite now maps to an independent Godog suite with its own feature source, tag filter, version axes, step registrars, and report files.
- Added SoperatorSuite(...) for embedded shared Soperator scenarios.
- Added DiscoveryHook support for external runners to attach custom cluster discovery after base Soperator discovery.
- Kept concrete shared step implementations internal and exposed reuse through SharedStepRegistrar().
- Moved reusable framework helpers such as Exec, ClusterState, Slurm job/node helpers, and WaitForWithJobAlive in the public acceptance/framework package.
- Moved concrete steps, Kubernetes object structs, discovery, and report formatting under acceptance/internal.
- Removed feature FS composition; independent Godog suites now avoid path remapping and report overwrites.
- Updated the standalone acceptance CLI and GitHub Actions acceptance source checkout to use the new public API.
- Moved and updated E2E documentation to e2e/README.md.
- Added tests for suite configuration, version filtering, discovery hooks, internal discovery, public API importability, and report formatting.
- Centralized shared step cleanup/reset so every shared step family runs cheap CleanupAndReset before and after each scenario, preventing scenario-local state such as selected workers, Slurm job IDs, ActiveCheck job IDs, and container cleanup handles from leaking between public or external feature files.

## Testing

E2e passes - https://github.com/nebius/soperator/actions/runs/30942698112


.github/workflows/e2e_test.yml
```diff
@@ -13,6 +13,11 @@ on:
         required: false
         default: ""
         type: string
+      soperator_e2e_repo_branch:
+        description: "Soperator E2E repo branch used to build acceptance (defaults to main)"
```
theyoprst: Better to use short text here so it fits on 1 line in the form. I suggest `E2E repo branch`

theyoprst (approved):

PR #2837, opened 2026-08-05
SCHED-662: Per-user resource isolation for login node SSH sessions

aydinmirac:
## Problem

All SSH sessions on a login node share one container cgroup. A single user's runaway process can consume the entire container memory budget — on cgroup v2 clusters the resulting container-scope OOM kills every user's session at once (k8s sets `memory.oom.group=1` on the container) — and CPU is shared per process, so a many-threaded job starves other users' shells. This is the bare-metal problem systemd-logind solves with `user-<uid>.slice`; our login containers run raw sshd with no systemd, so this PR provides that missing piece.

## How it works

- **Entrypoint** (`sshd_entrypoint.sh`): before `exec sshd`, resolves the container's own cgroup from `/proc/self/cgroup` (required on host-cgroup-namespace runtimes, where `/sys/fs/cgroup` is the node's root tree — verified on a real cluster), parks sshd in an `init/` child (cgroup v2 no-internal-processes rule), enables `+memory +cpu` for a `users/` subtree, and writes the resolved path into a ready-sentinel. Fails soft: on cgroup v1 or unwritable cgroupfs the feature disables itself with a log line.
- **PAM session hook** (`pam_exec`, session/optional): on each login, creates `users/user-<uid>`, applies `memory.high` / `memory.max` / `cpu.weight`, and moves the session process in — everything the user spawns inherits membership. Root and system users (uid < 1000) are exempt. Every failure path exits 0, so a broken hook can never block SSH logins.
- **Operator**: renders the limits into a ConfigMap (quantities converted to bytes at render time) mounted at `/etc/soperator/user-isolation.conf`, included in the login StatefulSet dependency hash so limit changes roll the pods. New optional CRD field `login.userIsolation` + chart values.

Cgroup files stay root-owned: the tree is visible inside the jail (users can read their own usage) but limits cannot be lifted without root.

## Defaults

Enabled by default. Unset memory limits are **derived from the sshd container memory limit** — `memory.high` = 80%, `memory.max` = 90% — so they are right-sized on any login node size (7.2/8.1Gi on the default 9Gi container, ~100/112Gi on a 128GB node). The derived max keeps the container-wide OOM unreachable by any single user without restricting how much a lone user may use. Explicit values override the derivation; `enabled: false` turns the feature off entirely.

## Validation (live CPU test cluster, 64GB login node / 54Gi sshd container)

Derived limits under test: `memory.high` = 43.2Gi, `memory.max` = 48.6Gi, `cpu.weight` = 100.

| Scenario | Without this PR | With this PR |
|---|---|---|
| User A allocates memory without bound | Allocation proceeds to the full 54Gi container limit; OOM kill at container scope (`oom_memcg=<container scope>`) | Allocation throttled at 43.2Gi (`memory.high`), then killed at 48.6Gi scoped to the user's cgroup (`oom_memcg=.../users/user-<user_id>`) |
| User B activity while user A allocates | Allocation stalls due to container-level reclaim; all sessions affected | User B's CPU benchmark unchanged vs idle (0.54s) for the entire duration |
| User B holds 30Gi resident; user A allocates until the container limit is reached | OOM victim selection is RSS-based across the whole container: user B's 30Gi process is killed while user A's process continues (verified in dmesg) | Per-user limits are enforced per user and do not cap the sum; combined usage can still reach the container limit (see "Limitations") |
| User A runs 32 CPU-bound processes; user B runs 1 | User B: 3.4× slowdown (real 1.85s vs 0.54s); degrades further with process count | User B: 0.88s real vs 0.82s user — scheduling delay eliminated; residual difference vs idle is hardware-level (SMT/turbo) sharing, independent of user A's process count |
| Idle pod, feature enabled | — | No measurable overhead (0.54s vs 0.54s baseline) |
| One user opens multiple SSH sessions | n/a | All sessions of a uid share one cgroup; limits apply to the user's total, not per process or per session |

## Limitations

Per-user limits bound each user individually, not the sum: N users can jointly exceed the container limit, in which case the container-scope OOM still applies (on cgroup v2 with `memory.oom.group=1`, this terminates all session processes; the sshd master and pod survive). Deployments expecting many concurrent memory-heavy users should set explicit budgets (`memoryMax ≈ container limit / N`) instead of relying on the derived defaults.


Uburro:
1. kubectl exec into the login container may break — needs manual verification (blocking)
This is the one thing I'd want tested before merge. [sshd_entrypoint.sh:466](vscode-webview://0qvd0lih4q8002u6kf8i8d4n51ugoeeo5sh2nijhhe28ntk44nvk/images/login/sshd_entrypoint.sh#L466) writes +memory +cpu to the container's own cgroup.subtree_control. Under cgroup v2's "no internal processes" rule, a non-root cgroup with non-empty subtree_control rejects writes to its cgroup.procs with EBUSY — and runc exec (behind kubectl exec, kubectl cp, ephemeral debug containers, and any exec probe on the sshd container) does exactly that write against the container cgroup path.

Nothing in the repo execs into login pods programmatically (no remotecommand usage anywhere), so CI will not catch this. Please verify on a real cluster with isolation enabled:


kubectl exec -it <cluster>-login-0 -c sshd -- true
If it does fail, there's no way to keep the controllers and avoid it — the fallback is documenting it loudly and defaulting enabled: false in the chart.

2. Failed controller enablement is silently reported as success
setup_user_isolation is invoked as setup_user_isolation || echo ..., which disables errexit for the whole function body. So if echo "+memory +cpu" > .../cgroup.subtree_control fails, execution continues and the sentinel is still written. The PAM hook then finds users/ present, creates users/user-N (which succeeds — no controller needed), and every limit write fails into 2>/dev/null. Net result: sessions are placed in an uncontrolled cgroup and users believe isolation is active.

Two contributing failure modes are realistic:

The proc-move loop reads cgroup.procs while mutating it (while read ... < "${cgroup_base}/cgroup.procs"), which can miss PIDs; any leftover process makes the subtree_control write fail with EBUSY. Loop until the file reads empty (with a bounded retry) instead of a single pass.
+memory +cpu is a single atomic write — if cpu isn't in cgroup.controllers (not delegated by the parent), memory isn't enabled either. Write them separately.
Suggested fix: check the write, and only write the sentinel after confirming the controllers landed.


for c in memory cpu; do
    echo "+${c}" > "${cgroup_base}/cgroup.subtree_control" 2>/dev/null || \
        echo "User isolation: controller ${c} unavailable"
done
case "$(cat "${cgroup_base}/cgroup.subtree_control")" in
    *memory*) ;;
    *) echo "User isolation: memory controller not enabled, feature disabled"; return 0 ;;
esac
3. Helm default (enabled: true) contradicts the CRD default (false)
[values.yaml:276](vscode-webview://0qvd0lih4q8002u6kf8i8d4n51ugoeeo5sh2nijhhe28ntk44nvk/helm/slurm-cluster/values.yaml#L276) ships enabled: true while the CRD defaults to false. Every existing cluster that re-renders the chart silently gains privileged cgroup restructuring on its login nodes — combined with issue 1, that's a meaningful upgrade risk for a brand-new feature. I'd default the chart to false for at least one release, or state the rollout intent explicitly in the PR description.

Separately: adding the ConfigMap to getLoginStatefulSetDependencies introduces a new pod-template annotation key, so all login pods roll once on operator upgrade regardless of whether the feature is used — that disconnects every active SSH session. Worth calling out in release notes.

4. Documented constraints aren't enforced
The CRD says "Must be lower than the sshd container memory limit" for both memoryHigh and memoryMax, and nothing checks it — not the ordering (memoryHigh <= memoryMax) either. Setting memoryHigh > memoryMax is accepted and produces a cgroup where memory.high is effectively ignored. The repo already uses CEL for this kind of thing ([slurmcluster_types.go:400](vscode-webview://0qvd0lih4q8002u6kf8i8d4n51ugoeeo5sh2nijhhe28ntk44nvk/api/v1/slurmcluster_types.go#L400)), so:


// +kubebuilder:validation:XValidation:rule="!has(self.memoryHigh) || !has(self.memoryMax) || quantity(self.memoryHigh).compareTo(quantity(self.memoryMax)) <= 0",message="memoryHigh must not exceed memoryMax"
The cross-field check against sshd.resources.memory is better placed in the existing login webhook.

5. All failure paths are invisible
The PAM hook logs exclusively via logger, which writes to /dev/log — no syslog daemon runs in the sshd container, so every log call is a no-op swallowed by || true. When isolation silently doesn't apply, there is zero signal anywhere. Since sshd runs as -D -e, its stderr goes to the container log and pam_exec propagates the child's stderr — so echo "..." >&2 actually surfaces in kubectl logs. Either that or pam_exec.so log=/var/log/user-isolation.log would make this debuggable.

aydinmirac:
@Uburro Thanks for the comprehensive review, all five points addressed.

**1. kubectl exec (blocking) — verified working on a live cluster with isolation enabled:**

    $ kubectl exec -it -n soperator login-0 -c sshd -- true && echo EXEC_OK
    EXEC_OK

    $ kubectl exec -n soperator login-0 -c sshd -- cat /proc/self/cgroup
    0::/kubepods.slice/kubepods-burstable.slice/kubepods-burstable-pod<...>.slice/cri-containerd-<...>.scope/init

The runtime attaches exec processes to the cgroup of the container's init process, which the entrypoint moves into `init/` — a leaf cgroup, so the "no internal processes" rule does not apply to the attach. `kubectl cp`, debug containers, and exec probes take the same path (the sshd container's own probes are tcpSocket-based). Caveat: verified on containerd/runc; a runtime that attaches exec to the container root cgroup would fail, and `enabled: false` (the default now, see point 3) leaves the cgroup tree completely untouched.

**2. Silent success — fixed:**
- The proc-move loop re-reads `cgroup.procs` until empty (bounded to 5 attempts).
- Controllers are enabled with separate writes, so an unavailable cpu controller no longer prevents memory from being enabled.
- The sentinel is written only after verifying `memory` is present in `users/cgroup.subtree_control` (one level deeper than suggested — that is the level the per-user cgroups inherit controllers from). On verification failure the feature logs and disables itself; the PAM hook stays inert.
- The `|| echo` call-site guard remains, but only as crash protection for the entrypoint's `set -e`: correctness no longer depends on errexit — every failure path is an explicit check, and none of them reaches the sentinel.

**3. Chart default — flipped to `enabled: false`** for this release as you suggested; we can plan that later in a following release (the derived limits make that a one-line change). Also correct about the pod roll: the new volume/mount rolls login pods once on operator upgrade regardless of the setting — this will be called out in the release notes.

**4. Constraint enforcement — added:** CEL rule for `memoryHigh <= memoryMax` on the CRD, plus a webhook check (create + update) that explicit per-user limits stay below `sshd.resources.memory`, with unit tests covering both rejections and the disabled-config case.

**5. Logging — fixed, with one deviation from the suggestion:** the hook now logs to stderr, but live testing showed pam_exec does **not** propagate the child's stderr to sshd — nothing reached the container log. The hook is therefore registered with `log=/proc/1/fd/1`, routing its output directly to the container's stdout. Verified end to end by corrupting a limit value and logging in three times:

    $ kubectl logs -n soperator login-0 -c sshd --since=2m | grep soperator-user-isolation
    soperator-user-isolation: failed to set memory.max for uid 1003
    soperator-user-isolation: failed to set memory.max for uid 1003
    soperator-user-isolation: failed to set memory.max for uid 1003

After restoring the value, subsequent logins log nothing — the hook only reports failures.

Please let me know if you want more improvements.

api/v1/slurmcluster_types.go
```diff
@@ -1044,6 +1045,51 @@ type SlurmNodeLogin struct {
...
+// Each user's session processes are placed into a dedicated child cgroup of the sshd container,
+// so one user cannot exhaust the memory or monopolize the CPU of the whole login node.
+// Requires cgroup v2 on the Kubernetes node; silently disabled otherwise.
+// +kubebuilder:validation:XValidation:rule="!has(self.memoryHigh) || !has(self.memoryMax) || quantity(self.memoryHigh).compareTo(quantity(self.memoryMax)) <= 0",message="memoryHigh must not exceed memoryMax"
+type LoginUserIsolation struct {
+	// Enabled turns on placement of each SSH session into a per-user cgroup
+	//
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=false
+	Enabled bool `json:"enabled,omitempty"`
```
Uburro: let make it pointer. we had couple of issue hen bool is not pointer in CR.

images/login/sshd_entrypoint.sh
```diff
@@ -28,6 +28,84 @@ mkdir -p /var/run/sshd
...
+    if ! grep -qw memory "${cgroup_base}/users/cgroup.subtree_control"; then
+        echo "User isolation: memory controller not enabled, feature disabled"
+        return 0
+    fi
+
+    # The sentinel activates the PAM hook; its content is the cgroup base path.
+    echo "${cgroup_base}" > /run/soperator-user-isolation.ready
+    echo "User isolation: per-user cgroup delegation is ready at ${cgroup_base}"
+}
+setup_user_isolation || echo "User isolation: setup failed, continuing without isolation"
```
Uburro: why we do not failed here, just skipping?

Uburro:
@aydinmirac Did you check what happens if multiple users reach the 90% limit at the same time? It looks like the entire container's `sshd` process would be killed. If that's the case, maybe we should configure more appropriate limits for that. What do you think?

api/v1/slurmcluster_types.go
aydinmirac: Agreed. I changed `LoginUserIsolation.Enabled` from `bool` to `*bool`, following the existing optional boolean pattern in the API.
- All consumers now handle `nil` safely as false using `ptr.Deref`, test fixtures use `ptr.To(true/false)`, and the generated deepcopy code was regenerated.
- I also added coverage for objects constructed directly in Go where API-server defaulting has not populated the pointer. `go test ./...` passes.

images/login/sshd_entrypoint.sh
aydinmirac: My original intention was to preserve the existing SSH behavior when cgroup isolation could not be established. I had conflated that availability fallback with the requested rollout behavior:
_"isolation is disabled by default for this release and can be enabled by default in a later release."_

But later I realized that my approach was wrong. The old code: `setup_user_isolation || echo "User isolation: setup failed, continuing without isolation"`

only protects against failures inside `setup_user_isolation`; it does not protect against the rest of the entrypoint crashing.

 I changed the behavior accordingly: 
- Missing/disabled configuration remains a successful no-op.
- Once isolation is enabled, missing cgroup v2 support, an unwritable cgroup, failed process migration, or unavailable controllers causes container startup to fail before sshd starts.
- Both memory and cpu controllers are verified before writing the readiness sentinel.
- The `setup_user_isolation || echo ...` wrapper was removed.

aydinmirac:
> @aydinmirac Did you check what happens if multiple users reach the 90% limit at the same time? It looks like the entire container's `sshd` process would be killed. If that's the case, maybe we should configure more appropriate limits for that. What do you think?

@Uburro let me share my tests and findings;

I tested this exact scenario on the live single GPU test cluster. The SSHD container and derived per-user configuration were:

```
container memory.max:       57982058496 (54 GiB)
container memory.oom.group: 1
per-user memory.high:       43.2 GiB
per-user memory.max:        48.6 GiB
```

I connected as two different users and confirmed they were in separate per-user cgroups:

```
john: .../users/user-1004
bob:  .../users/user-1003
```

User A held 25 GiB, and user B then allocated memory until reaching 28.5 GiB. Both users were below their individual 43.2 GiB memory.high, but their aggregate usage plus container overhead reached the 54 GiB container limit.

Before the test, the container OOM counters were:

```
low 0
high 0
max 0
oom 0
oom_kill 0
oom_group_kill 0
```

After the sessions were terminated:

```
low 0
high 0
max 138
oom 1
oom_kill 0
oom_group_kill 1
```

All active SSH sessions were disconnected, but the SSHD master survived. Its elapsed time continued increasing and it remained protected:

Before:
```
 PID COMMAND ELAPSED
   1 sshd      78724
 oom_score_adj: -1000
```

After:
```
 PID COMMAND ELAPSED
   1 sshd      79567
 oom_score_adj: -1000
```

The pod status also remained unchanged:

```
 containerID=containerd://b1cab25944fc1ab1020c6d1b55802ad9566e30d15413ef5aea22f1aae7257ae7
 restarts=0
 ready=true
```

I was able to establish a new SSH connection immediately after the OOM event.

Therefore, the container’s SSHD master is not killed or restarted. Because `memory.oom.group=1` is set at the container level, however, all killable SSH session processes are terminated together. PID 1 is exempt because it has `oom_score_adj=-1000`

I agree that the current 90% derived per-user maximum does not protect against aggregate usage by multiple users. Lowering it based on:

(container limit - system reserve) / expected concurrent heavy users

works when concurrency is known. For this 54 GiB container and two heavy users, approximately `memoryMax: 24Gi` and `memoryHigh: 21Gi `would prevent this particular scenario. However, no fixed per-user percentage handles an arbitrary number of users without unnecessarily restricting a lone user.

For a single abusive user, the isolation works as intended: allocation is throttled after `memory.high=43.2Gi` and the allocating process is OOM-killed at `memory.max=48.6Gi` within that user’s cgroup. Other users remain responsive, and the container limit is not reached as mentioned above in the PR description.

internal/render/login/container.go
```diff
@@ -29,6 +29,7 @@ func renderContainerSshd(
 		common.RenderVolumeMountInMemory(),
 		common.RenderVolumeMountTmpDisk(),
 		renderVolumeMountSshdConfigs(),
+		renderVolumeMountUserIsolation(),
```
ali-sattari: Can we avoid mounting and hashing the ConfigMap (and the volume mount) while the feature is disabled?

aydinmirac: Fixed. When user isolation is omitted or disabled, its ConfigMap is no longer added as a pod volume, mounted into the SSHD container, or included in the StatefulSet dependency hash.

Let me know if something is missing.

ali-sattari (approved):

PR #2840, opened 2026-08-06
SCHED-2014: Support setting hostNetwork and dnsPolicty for worker pods

ali-sattari:

### DIFF
--- a/api/v1alpha1/nodeset_types.go
+++ b/api/v1alpha1/nodeset_types.go
@@ -189,6 +189,21 @@ type NodeSetSpec struct {
 	// +kubebuilder:default=false
 	EnableHostUserNamespace bool `json:"enableHostUserNamespace,omitempty"`
 
+	// HostNetwork controls whether worker pods use the host network namespace.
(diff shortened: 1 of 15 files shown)


api/v1alpha1/nodeset_types.go
```diff
@@ -189,6 +189,21 @@ type NodeSetSpec struct {
 	// +kubebuilder:default=false
 	EnableHostUserNamespace bool `json:"enableHostUserNamespace,omitempty"`
 
+	// HostNetwork controls whether worker pods use the host network namespace.
+	// Defaults to false.
+	//
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=false
+	HostNetwork bool `json:"hostNetwork,omitempty"`
```
Uburro, line 197: would be cool if we can merge whole pod.spec here in a future

ali-sattari: I looked into it, it was not easy, I explained in more detail in the ticket.,

PR #2864, opened 2026-08-13
SCHED-2156: New metrics and labels for NVLink Instance Group

ali-sattari:
## Problem

- We need to add NVLink Instance Group as a label for node related metrics
- Grafana dashboards have incorrect and varying values for `datasource`

## Solution

- Added `nvlink_instance_group` as label to `slurm_node_jobs`
- Added new metric `slurm_node_nvlink_instance_group` to join with other metrics for NVLink centric panels and alerts
- Changes all Grafana dashboards to have a `$datasource` template variable that defaults to `VictoriaMetrics` 

## Testing

- Tested in a dev cluster, k8s nodes labels correctly get picked up by exporter

### DIFF
--- a/internal/exporter/kubernetes_topology.go
+++ b/internal/exporter/kubernetes_topology.go
@@ -1,0 +1,182 @@
+package exporter
+
+import (
+	"context"
+	"fmt"
+
+	corev1 "k8s.io/api/core/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/apimachinery/pkg/selection"
+	"k8s.io/client-go/rest"
+	ctrlcache "sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	"nebius.ai/slurm-operator/internal/consts"
+)
+
+type kubernetesNodeTopologySource struct {
+	reader      client.Reader
+	namespace   string
+	clusterName string
+}
+
+func topologyNodeSelector() (labels.Selector, error) {
+	nodeSetRequirement, err := labels.NewRequirement(slurmNodeSetNameLabel, selection.Exists, nil)
+	if err != nil {
+		return nil, fmt.Errorf("create Slurm NodeSet label selector: %w", err)
+	}
+	return labels.NewSelector().Add(*nodeSetRequirement), nil
+}
+
+// NewKubernetesNodeTopologyCache creates a list/watch cache scoped to the Kubernetes objects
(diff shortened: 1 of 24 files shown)


internal/exporter/kubernetes_topology.go
```diff
@@ -0,0 +1,78 @@
...
+		client.MatchingLabels{
+			consts.LabelInstanceKey: s.clusterName,
+			consts.LabelWorkerKey:   consts.LabelWorkerValue,
+		},
+	); err != nil {
+		return nil, fmt.Errorf("list Slurm worker pods: %w", err)
+	}
+
+	var nodeList corev1.NodeList
+	if err := s.client.List(ctx, &nodeList); err != nil {
```
Uburro: This is an uncached list request. It could become an issue in large clusters

internal/exporter/kubernetes_topology.go
```diff
@@ -0,0 +1,78 @@
...
+		return nil, fmt.Errorf("list Kubernetes nodes: %w", err)
+	}
+	nodes := make(map[string]corev1.Node, len(nodeList.Items))
```
Uburro: do we realy need to store whole  corev1.Node spec

Uburro:
1. main.go:237 — the 10 s warm-up budget is one-shot and permanent. The real sync is the ListNodeTopologies at main.go:261, which triggers the lazy initial LIST of every Node in the cluster. On the large clusters this feature targets that can exceed 10 s; then main.go:392 cancels the cache and topology metrics are off for the whole process lifetime — no retry, only a pod restart recovers. Make the timeout configurable, or retry in the background instead of disabling permanently.

2. main.go:252 — WaitForCacheSync there is still a no-op in production. No informers exist yet at that point (ByObject creates them lazily), so it returns true immediately and the "topology cache stopped before synchronization" branch is unreachable. Harmless now, but the code reads as if it waits, and the tests at main_test.go:30 use a fake syncer so they don't cover it. Either call GetInformer for Pod and Node first, or drop the wait and rely on the List.

3. kubernetes_topology.go:43 — Node informer is still cluster-wide with no selector (unchanged in the force-push). Pod gets namespace + label scoping, Node gets only Transform, which runs after decode — saves indexer memory, not wire traffic or decode CPU, per exporter per SlurmCluster. This is also what makes the initial LIST slow enough to hit the 10 s budget in point 1. Add an Exists selector on slurm.nebius.ai/nodeset-name, and/or use a PartialObjectMetadata informer for Node.

4. kubernetes_topology.go:84-96 — whole-cluster node map rebuilt every cycle (unchanged). The cached List deep-copies every item and builds a map over all nodes, to serve lookups for only the worker pods' NodeNames. Collect the distinct pod.Spec.NodeName values and reader.Get each instead — O(1) on the indexer, copies only what's used, and skips node work entirely when no pods matched.

Uburro (approved):

PR #2865, opened 2026-08-13
SCHED-1519: add Infrastructure control-plane Deployments schedule onto worker nodes due to missing node affinity

Uburro:
## Problem

Non-soperator control-plane Deployments installed via the soperator-fluxcd umbrella chart carried no
nodeSelector or affinity, so the scheduler was free to bin-pack them onto worker nodes — and did,
preferring the large worker-64vcpu nodes. A worker node has only ~900m of CPU left for non-worker
pods, most of it already taken by the base DaemonSets. On bfl-me-west1-cpu-dataprocessing this left
3 of 5 worker-64vcpu nodes unable to schedule the node-configurator (rebooter) DaemonSet.

## Solution

Pin every control-plane workload the chart installs to the system nodeset via
`nodeSelector: slurm.nebius.ai/nodeset=system`, plus a soft pod anti-affinity so a Deployment's
replicas spread across nodes. All of it lives in `values.yaml`, so it stays visible and overridable.

Give the rebooter `priorityClassName: system-node-critical` so it can preempt whatever else lands on
a packed worker node.

Add a CPU request floor where there was none (100m; 50m for k8up and the tsa-token-writer) —
without a request these pods score as free and bin-pack anywhere.

Add `affinity`/`nodeSelector`/`tolerations` to the `soperatorchecks` chart, which had no scheduling
fields at all.

## Testing

`make helmtest` (388 tests, 17 charts) and `make test` pass. New suites cover the placement, the CPU
floors and the rebooter priority class; verified the rendered HelmReleases against each sub-chart's
actual value schema (`helm show values`) and pod labels (`helm template`). `make sync-version-from-scratch`
produces no drift.

Not verified on a live cluster. `make lint` fails on a pre-existing toolchain mismatch (`bin/golangci-lint`
built with go1.25 vs the repo's 1.26.4 target) — reproduces on a clean `main`; no Go code changed here.

## Release Notes

Fix: control-plane components installed by soperator-fluxcd now run on system nodes instead of
competing for the worker nodes' reserved CPU budget, and the rebooter DaemonSet is guaranteed a slot
on every node.

Risk: the pin is a hard `nodeSelector`. A cluster with no `slurm.nebius.ai/nodeset=system` nodes will
leave these pods Pending — clear the component's `values` block or set `overrideValues` to opt out.
The SPO webhook, kruise manager anti-affinity, nfs-server and notifier are out of scope.

### DIFF
--- a/helm/soperator-fluxcd/values.yaml
+++ b/helm/soperator-fluxcd/values.yaml
@@ -50,7 +50,67 @@ certManager:
   version: v1.20.*
   namespace: cert-manager-system
   releaseName: cert-manager
-  values: null
+  values:
+    # kubernetes.io/os is the chart's own default and has to be repeated here: setting
+    # nodeSelector replaces it wholesale.
+    nodeSelector:
+      kubernetes.io/os: linux
+      slurm.nebius.ai/nodeset: system
+    affinity:
+      podAntiAffinity:
+        preferredDuringSchedulingIgnoredDuringExecution:
+          - weight: 100
+            podAffinityTerm:
+              topologyKey: kubernetes.io/hostname
+              labelSelector:
+                matchLabels:
+                  app.kubernetes.io/name: cert-manager
+                  app.kubernetes.io/component: controller
+    resources:
+      requests:
+        cpu: 100m
+    webhook:
+      nodeSelector:
+        kubernetes.io/os: linux
+        slurm.nebius.ai/nodeset: system
+      affinity:
+        podAntiAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
+            - weight: 100
+              podAffinityTerm:
+                topologyKey: kubernetes.io/hostname
+                labelSelector:
+                  matchLabels:
+                    app.kubernetes.io/name: webhook
+                    app.kubernetes.io/component: webhook
+      resources:
+        requests:
+          cpu: 100m
+    cainjector:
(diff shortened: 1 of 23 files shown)


helm/soperator-fluxcd/values.yaml
```diff
@@ -694,6 +955,9 @@ soperator:
         remediateLastFailure: true
     values:
       rebooter:
+        # Cluster-critical DaemonSet: without a priority class it loses the worker nodes'
+        # remaining CPU budget to whatever Deployment lands there first and never preempts.
+        priorityClassName: system-node-critical
```
ali-sattari: This can preempt worker pods as well, right? ([ref](https://github.com/nebius/soperator/blob/063ec3eb780bf99d00e3ddc66c23d395ff38c64a/helm/nodesets/values.yaml#L11-L24))

Uburro: > This can preempt worker pods as well, right? ([ref](https://github.com/nebius/soperator/blob/063ec3eb780bf99d00e3ddc66c23d395ff38c64a/helm/nodesets/values.yaml#L11-L24))

yup. it's good point. thank you. I will fix it

Uburro: fixed

ali-sattari (approved):

PR #2883, opened 2026-08-18
SCHED-0000: Publish E2E module from Soperator

ChessProfessor:
## Problem

The E2E module inside the Soperator repository still uses the standalone nebius.ai/soperator-e2e module path. External consumers therefore cannot import the framework directly from the public Soperator repository without a replace directive.

## Solution

Rename the nested module and its internal imports to github.com/nebius/soperator/e2e. This allows external runners to depend directly on the framework version published from the Soperator repository.

### DIFF
--- a/e2e/README.md
+++ b/e2e/README.md
@@ -81,7 +81,7 @@ instance removal.
 
 ## Reusable Runner API
 
-Use `nebius.ai/soperator-e2e/acceptance` from another Go runner when you
+Use `github.com/nebius/soperator/e2e/acceptance` from another Go runner when you
 want to reuse the shared Soperator scenarios for a different Soperator cluster
 type.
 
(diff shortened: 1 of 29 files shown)


theyoprst (approved):
checked the rename - complete. `git grep nebius.ai/soperator-e2e` finds nothing at head, no replace directives left in `e2e/go.mod`, and the `soperator-e2e` string in `e2e_test.yml` is just a checkout dir name, unaffected. the path now matches the repo layout, so external `go get github.com/nebius/soperator/e2e@<commit>` resolves - which is exactly why removing the replace couldn't work with the old `nebius.ai` path (no go-import metadata behind it).

one thing for later: if consumers want to pin by version instead of commit, nested modules need tags in `e2e/vX.Y.Z` format.

LGTM

PR #2884, opened 2026-08-18
SCHED-2127: Derive MariaDB buffer pool size from memory allocation

theyoprst:
### Problem

The operator hardcodes `innodb_buffer_pool_size=32768M` in the MariaDB `myCnf` for every cluster, with no override. The 32Gi pool exceeds the acct-db pod memory limit (12Gi in prod today), so once accounting data outgrows the limit, InnoDB caches past the cgroup limit and the DB gets OOM-killed. Memory set in Terraform has no effect on the actual pool size.

### Solution

- Render `innodb_buffer_pool_size` as 50% of the acct-db container memory allocation (floored to whole MiB), falling back to the legacy 32768M when no memory allocation is known.
- Propagate `MyCnf` in the reconciler patch so existing clusters pick up the derived value.

### Testing

Unit tests for the size derivation, rendering with and without resources, and `MyCnf` propagation in the reconciler patch.

### Release Notes

Fix: MariaDB `innodb_buffer_pool_size` is now derived from the accounting DB memory allocation (50% of it) instead of the hardcoded 32Gi.

### DIFF
--- a/internal/consts/mariadb.go
+++ b/internal/consts/mariadb.go
@@ -9,11 +9,12 @@ const (
 	MariaDbSecretName     = "mariadb-password"
 	MariaDbSecretRootName = "mariadb-root"
 	MariaDbPort           = 3306
-	MariaDbDefaultMyCnf   = `[mariadb]
+	// MariaDbMyCnfTemplate is a fmt template; the verb is the innodb_buffer_pool_size value in MiB.
+	MariaDbMyCnfTemplate = `[mariadb]
 bind-address=*
 default_storage_engine=InnoDB
 innodb_default_row_format=DYNAMIC
-innodb_buffer_pool_size=32768M
+innodb_buffer_pool_size=%dM
 innodb_log_file_size=64M
 innodb_lock_wait_timeout=900
 max_allowed_packet=16M`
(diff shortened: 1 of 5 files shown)


Uburro (dismissed):

itechdima (dismissed):

nebius-devtools (commented):
minor findings from a deep pass, nothing blocking:

1. `internal/render/accounting/mariadb_test.go:56` - `ptrQuantity` duplicates `ptr.To` from `k8s.io/utils/ptr`, already used in this package
2. `internal/controller/reconciler/mariadb_test.go:33` - the test checks that `patch()` mutates the object but never inspects the marshalled patch payload, and spot-checks 2 of the 14 copied fields

also noted for a follow-up: the field-by-field copy list in `patch()` is the same bug class this PR fixes for MyCnf - every new rendered field silently doesn't propagate until someone remembers to add it there

nebius-devtools:
1. fixed in 5ae45b30
2. leaving as is - checking the mutated object matches the other patch tests in the package (`TestConfigMapReconciler_patch` does the same)
3. agreed, same bug class - `patch()` also misses `Metrics`, `Replication`, `ImagePullPolicy`, `PriorityClassName`, `Affinity`, `ImagePullSecrets`. will track in a follow-up ticket

nebius-devtools (approved):

PR #2887, opened 2026-08-18
SCHED-2333: add rolling update controller

itechdima:
## Problem

Default rolling update in k8s doesn't respect slurm workload inside slurm nodes.

## Solution

Implement custom rolling update controller that would perform pod recreation when needed.

## Testing

Dev cluster.

## Release Notes

Feature: add custom rolling update controller in soperator to respect slurm workload during NodeSet pods redeployment.


helm/nodesets/values.yaml
```diff
@@ -67,7 +67,7 @@ nodesets:
     # Optional, defaults to 20%
     maxUnavailable: 1
```
Uburro: maxConcurrentStartup is the same maxUnavailable

Uburro (approved):

PR #2889, opened 2026-08-19
SCHED-2288: Support Slurm multiple topologies (topology.yaml) in soperator

Uburro:
## Problem
A cluster with heterogeneous fabrics could describe only one flat topology, so InfiniBand GPU NodeSets and Ethernet-only CPU NodeSets were squeezed into the same switch tree, degrading scheduling decisions and pushing users into `customSlurmConfig` workarounds.

## Solution
Render `spec.topology.topologies` into Slurm 25.05+ `topology.yaml` as several named topologies, let partitions bind to one via `topologyRef` (`Topology=<name>` in `slurm.conf`), and have each worker join exactly the topologies the rendered file lists it in.

## Testing
`make test`, `make lint`, `make helmtest` and the worker Python suite pass; verified end to end on a live cluster — `scontrol show topo <name>` matches the rendered file for flat, tree and block topologies, partition bindings show up in `scontrol show partition`, and `--switches=1` demonstrably constrains placement in the tree topology.

## Release Notes
Feature: A SlurmCluster can now define several named network topologies and bind partitions to them individually, so different fabrics are optimized separately. Breaking: `slurmConfig.topologyPlugin` and `topology.blockSize` are removed — clusters using a topology must declare `spec.topology.topologies`, and workers roll on upgrade.


api/v1/slurmcluster_types.go
```diff
@@ -216,9 +216,77 @@ type Topology struct {
...
+	// Leaving it empty keeps the legacy single-topology behaviour driven by
+	// SlurmConfig.TopologyPlugin and BlockSize, which renders topology.conf instead.
+	//
+	// +kubebuilder:validation:Optional
+	Topologies []NamedTopology `json:"topologies,omitempty"`
+}
+
+// NamedTopology defines one entry of topology.yaml: a named topology, the plugin that backs it,
+// and the NodeSets whose Slurm nodes it covers.
+type NamedTopology struct {
```
rdjjke, line 229: Why not "Topology"?

api/v1/slurmcluster_types.go
```diff
@@ -216,9 +216,77 @@ type Topology struct {
...
+	// operator promotes the first one rather than leave the choice to file order. Setting this to
+	// false everywhere therefore does not produce a cluster without a default.
+	//
+	// +kubebuilder:validation:Optional
+	ClusterDefault *bool `json:"clusterDefault,omitempty"`
+
+	// Topo selects the topology plugin and carries its plugin-specific parameters.
+	//
+	// +kubebuilder:validation:Required
+	Topo TopologyPluginSpec `json:"topo"`
```
rdjjke: Why not "TopologyPlugin"?

api/v1/slurmcluster_types.go
Uburro: we have alreafe had Topology struct

rdjjke (approved):

PR #2893, opened 2026-08-21
SCHED-682: Enable Docker from login SSH and unify the Docker proxy

aydinmirac:
## Summary

This PR contains the Docker portion of SCHED-682.

The Enroot-over-SSH implementation was split into PR #2901 (https://github.com/nebius/soperator/pull/2901) and has already been merged. This PR adds optional Docker support to login nodes and replaces the worker Nginx/NJS Docker proxy with a shared Go proxy used by both login and worker nodes.

Login Docker remains disabled by default in the Soperator Helm chart. Deployment profiles such as Soperator Pro can enable it after provisioning suitable node-local image storage.

```yaml
  slurmNodes:
    login:
      userIsolation:
        enabled: true
      docker:
        enabled: true
      volumes:
        jailSubMounts:
          - name: image-storage
            mountPath: /mnt/image-storage
            volumeSourceName: image-storage
```

### Shared Docker proxy

Login and worker nodes now use the same Go proxy implementation, with different cgroup-resolution policies:

  - Login mode identifies the caller through Unix `SO_PEERCRED` and assigns containers to that user’s Docker cgroup.
  - Worker mode identifies the calling Slurm task through its peer PID and cgroup membership.
  - Worker resolution supports both the Slurm 25.x `job_<id>` hierarchy and the Slurm 26.05 `SLUID` hierarchy.
  - Client-provided `Cgroup-Parent` headers and `HostConfig.CgroupParent` values are removed and replaced where attribution is available.
  - Streaming and upgraded Docker connections, including attached docker run and docker exec, are supported.

The existing worker Nginx/NJS proxy is removed. Both login and worker nodes run dockerd and the shared proxy under supervisord in their main container.

### Login nodes

When login.docker.enabled=true:

  - dockerd stores data under `/mnt/image-storage/docker`.
  - The standard `/var/run/docker.sock` is the user-facing proxy socket.
  - The private daemon socket is `/run/soperator-dockerd.sock` and is inaccessible to ordinary SSH users.
  - Container-create requests are assigned to:

`  users/user-<uid>/docker/<container-id>`

  When Docker is enabled, SSH processes are moved into a sibling sessions leaf:

```
  users/user-<uid>
  ├── sessions
  └── docker
      └── <container-id>
```

  The aggregate user cgroup limits therefore cover both SSH sessions and Docker containers while satisfying the cgroup v2 no-internal-process rule.

  The validating webhook requires:

  - login user isolation to be enabled;
  - a writable jail sub-mount at `/mnt/image-storage`;
  - the Docker environment variable to remain managed by Soperator.

  ### Worker nodes

  Worker Docker behavior remains controlled by `NodeSet.spec.docker.enabled`.

  The Slurm TaskProlog exports:

`DOCKER_HOST=unix:///var/run/soperator-docker.sock`

  The worker proxy derives the correct Slurm job cgroup from the calling process rather than trusting a client-provided header. Containers launched through srun docker run are placed beneath the corresponding
  Slurm step’s user cgroup.

  Custom worker supervisor ConfigMaps must launch both the guarded dockerd and shared proxy entrypoints when Docker is enabled.

  ## Validation

  Automated validation included:
  - focused race tests for the proxy, login rendering, and E2E helpers
  - Helm unit tests
  - Linux AMD64 and ARM64 image builds
  - Helm integration tests using the built images

  All CI checks are currently passing.

  Live validation against revision `r1855a008` covered:

  - Docker build, image execution, bind mounts, attached runs, detached runs, and docker exec from login SSH.
  - Denial of access to the private login dockerd socket.
  - Enforcement of `users/user-<uid>/docker` despite a forged cgroup parent.
  - SSH session placement under `users/user-<uid>/sessions`.
  - Detached container survival across SSH reconnects.
  - Worker Docker placement under a Slurm 26.05 SLUID cgroup.
  - Worker Docker storage under `/mnt/image-storage/docker`.
  - Job cancellation and Docker container cleanup.
  - Docker GPU access from a Slurm job; nvidia-smi detected an NVIDIA B200.
  - Enroot-over-SSH regression checks on login and worker nodes.
  - Controller, login, worker, REST, and sconfigcontroller health with zero container restarts.

  ## Operational notes
  
  - Each login pod requires non-shared writable image storage suitable for running a Docker daemon.
  - This uses a shared rootful Docker daemon and assumes trusted cluster users, consistent with the existing worker-node model. The proxy provides cgroup attribution and protects the private daemon socket; it is not a complete Docker authorization boundary.

  - Docker containers are attributed to individual users. BuildKit workloads from the default Buildx path remain bounded by the login pod rather than the individual user cgroup.
  - Docker health is not currently included in the SSHD readiness probe.
  - This worker proxy replacement targets the Soperator 5.0 major release, whose upgrades currently use a maintenance window. Temporary compatibility with the removed Nginx/NJS proxy is therefore not included.


images/login/sshd.dockerfile
```diff
@@ -2,14 +2,32 @@
...
-# Install OpenSSH server
+# Install OpenSSH server, Docker, and process-supervision dependencies
 # Create root .ssh directory
 # Remove autogenerated sshd server keys
 RUN apt-get update && \
-    apt install -y openssh-server && \
+    apt install -y \
+        containerd.io \
+        docker-buildx-plugin \
+        docker-ce \
```
asteny: Here should be an ansible role for install docker and also we need to use it in worker image. Previously same installation was hadcoded in docker file https://github.com/nebius/soperator/blob/main/images/worker/slurmd.dockerfile#L112

images/login/sshd.dockerfile
```diff
@@ -92,11 +110,45 @@ RUN rm -rf /home
 # Delete SSH "message of the day" scripts because they will be linked from jail
 RUN rm -rf /etc/update-motd.d
 
+# Install the native PAM module that places each SSH session in its own mount
+# namespace and pivots it into /mnt/jail.
+COPY images/login/pam_soperator_jail.c /usr/src/pam-soperator-jail/
+RUN PAM_MODULE_DIR="/usr/lib/$(gcc -print-multiarch)/security" && \
```
asteny: lets move all compilation to the build stage and use only .so from it

api/v1/slurmcluster_types.go
```diff
@@ -1062,6 +1062,22 @@ type SlurmNodeLogin struct {
 	//
 	// +kubebuilder:validation:Optional
 	UserIsolation *LoginUserIsolation `json:"userIsolation,omitempty"`
+
+	// Docker defines Docker support for SSH sessions on login nodes.
+	// A writable jail sub-mount at /mnt/image-storage is required when enabled.
+	//
+	// +kubebuilder:validation:Optional
+	Docker *LoginDocker `json:"docker,omitempty"`
```
asteny, line 1070: Maybe here could be an options for enabling/disabling docker on login and on worker? And we need to disable docker at all (maybe in terraform) if there is no /mnt/image-storage

rdjjke: There is an option for disabling docker on worker nodes, but it's in NodeSet CR

internal/consts/docker.go
```diff
@@ -1,5 +1,10 @@
...
-// EnvDockerEnabled tells worker entrypoint scripts whether Docker components
-// (dockerd, docker CLI in jail) should be available on the node.
-const EnvDockerEnabled = "SOPERATOR_DOCKER_ENABLED"
+const (
+	// EnvDockerEnabled tells node entrypoint scripts whether Docker components
+	// should be available on the node.
+	EnvDockerEnabled = "SOPERATOR_DOCKER_ENABLED"
+
+	// DockerImageStorageMountPath is the fixed jail and daemon mount for OCI data.
+	DockerImageStorageMountPath = "/mnt/image-storage"
```
rdjjke, line 9: Shouldn't it be `/mnt/image-storage/docker`? I see that this path is used for docker running on worker nodes

aydinmirac: - `/mnt/image-storage` is the volume mount root.
- `/mnt/image-storage/docker` is Docker’s data-root inside that volume.

the volume is mounted at `/mnt/image-storage`, while Docker stores the data under `/mnt/image-storage/docker`.

Something like:

```bash
  /mnt/image-storage/
  ├── docker/
  └── enroot/
```

If you prefer, I can rename the constant to `DockerImageStorageVolumeMountPath` to make it more explicit.

api/v1/slurmcluster_types.go
aydinmirac: Right now login and worker Docker already have separate mechanisms at their levels:
  - Login: `SlurmCluster.spec.slurmNodes.login.docker.enabled`
  - Worker: `NodeSet.spec.docker.enabled`

When login Docker is enabled, it requires a jailSubMount at `/mnt/image-storage` and rejects the configuration when the mount is absent or read-only, here in the function:

https://github.com/nebius/soperator/blob/007fdf5d2d9e14097e07720134431c41266b8617/internal/webhook/v1/slurmcluster_webhook.go#L75

images/login/sshd.dockerfile
aydinmirac: Thanks. Done. I added a shared ansible/roles/docker role with an ansible/docker.yml playbook and now use it from both the login and worker images. The Docker package installation is no longer duplicated in the Dockerfiles.

images/login/sshd.dockerfile
aydinmirac: I moved the PAM module compilation into a dedicated login_pam_builder stage. The final login image only copies the compiled .so, it no longer installs libpam0g-dev or compiles the PAM source.

rdjjke:
Docker on login nodes is disabled by default in Helm values, but we want this to be enabled in Soperator Pro. So there should be a separate PR to nebius-solution-library that sets it to true. It's JFYI, doesn't block this PR.

Also, the TF should add the node local image storage disks for login nodes. And as for them, there's one more difference -- worker nodes currently have dedicated structure in nodeset Helm values, but login nodes don't have one. I suggest adding it just to keep things more uniform.

rdjjke:
Worker nodes already have their own implementation of Docker proxy, that does two things:
1. Prepends mount paths with /mnt/jail/ (since dockerd runs out of jail)
2. Sets container cgroup to the one coming from slurmstepd, so that `srun docker run ...` is limited to the Slurm job's resources.

Could you pls investigate whether it's possible to reuse the same proxy implementation on both login and worker nodes? Either yours or the one that already exists.

It also applies to the supervisord configuration -- worker nodes also have a custom configuration when docker is enabled, similary to how login nodes have it in your PR. However, these customizations are applied in different ways. It would be great to make this more uniform as well.

rdjjke:
I suspect that Enroot doesn't work in SSH to worker nodes, simply because the PAM module you added isn't installed there. I think we should make it work there as well, but then we should either split this PAM modules into two modules (one handling login user cgroups and another handling the pivot_root stuff), and install only one of them on worker nodes, or maybe just get rid of login cgroup inheritance for Docker. Because I'm not sure if it's really necessary - we don't have data on that.

aydinmirac:
> Docker on login nodes is disabled by default in Helm values, but we want this to be enabled in Soperator Pro. So there should be a separate PR to nebius-solution-library that sets it to true. It's JFYI, doesn't block this PR.
> 
> Also, the TF should add the node local image storage disks for login nodes. And as for them, there's one more difference -- worker nodes currently have dedicated structure in nodeset Helm values, but login nodes don't have one. I suggest adding it just to keep things more uniform.

@rdjjke I will create a non-blocking follow-up PR in nebius-solutions-library for Soperator Pro. It should add login node-local image storage analogous to the existing worker node_local_image_storage structure.

One clarification: when you mention adding a dedicated structure to keep it uniform with NodeSets, do you mean the equivalent node_local_image_storage structure in nebius-solutions-library, or do you also want a new high-level structure in the Soperator slurm-cluster Helm values?

aydinmirac:
> Worker nodes already have their own implementation of Docker proxy, that does two things:
> 
> 1. Prepends mount paths with /mnt/jail/ (since dockerd runs out of jail)
> 2. Sets container cgroup to the one coming from slurmstepd, so that `srun docker run ...` is limited to the Slurm job's resources.
> 
> Could you pls investigate whether it's possible to reuse the same proxy implementation on both login and worker nodes? Either yours or the one that already exists.
> 
> It also applies to the supervisord configuration -- worker nodes also have a custom configuration when docker is enabled, similary to how login nodes have it in your PR. However, these customizations are applied in different ways. It would be great to make this more uniform as well.

@rdjjke The forwarding and Docker API rewriting logic can be shared, but the cgroup source differs:
- Workers receive the Slurm cgroup from slurmstepd through TaskProlog.
- Login nodes do not have slurmstepd; the current implementation derives the calling SSH user from Unix socket peer credentials.

Therefore, a shared proxy would likely need different attribution modes rather than using the worker implementation unchanged. Reusing the worker header directly on login nodes would allow the SSH user to control the header and would not reliably preserve per-user isolation.

I will keep this design discussion in the Docker-specific PR together with mount-path handling and supervisord unification.

aydinmirac:
> I suspect that Enroot doesn't work in SSH to worker nodes, simply because the PAM module you added isn't installed there. I think we should make it work there as well, but then we should either split this PAM modules into two modules (one handling login user cgroups and another handling the pivot_root stuff), and install only one of them on worker nodes, or maybe just get rid of login cgroup inheritance for Docker. Because I'm not sure if it's really necessary - we don't have data on that.

@rdjjke Confirmed on the test cluster. The current worker image has Enroot 4.0.1 installed, but it does not contain pam_soperator_jail.so or a corresponding PAM session entry, and worker SSH still uses ChrootDirectory /mnt/jail.

I verified both direct SSH to a worker and the normal login-node-to-worker SSH path:

```
jail / virtiofs rw,relatime
unshare: unshare failed: Operation not permitted
unshare_exit=1
```

enroot create successfully extracted the SquashFS image, but enroot start failed:

```
create_exit=0
enroot-nsenter: failed to create user namespace: Operation not permitted
start_exit=1
```

So having Enroot installed is insufficient: direct worker SSH needs the same mount-namespace/pivot-root PAM mechanism as login SSH.

As you suggested, I will split this current work into two stacked PRs:

1. A focused PAM jail/Enroot PR for both login and worker SSH.
2. A Docker-specific PR covering login Docker, proxy reuse, supervisord, storage, and cgroup policy.

The PAM responsibilities are already separated:
- pam_soperator_jail.so handles the mount namespace and pivot_root.
- The optional pam_exec hook handles login-only per-user cgroup placement.

I will make the jail module generic, remove its Docker-specific environment handling, install it in both images, and install only the pivot-root module on workers. The login cgroup hook will remain login-only.

I would prefer not to remove Docker cgroup inheritance unconditionally. Without it, a user could bypass per-user limits merged in the SCHED-662 by moving the workload into docker run. We can discuss making that inheritance conditional on login.userIsolation.enabled in the Docker-specific PR.

images/worker/slurmd.dockerfile
```diff
@@ -2,6 +2,18 @@
...
+
+WORKDIR /build
+
+COPY go.mod go.sum ./
+RUN go mod download
+
+COPY cmd/soperator-docker-proxy cmd/soperator-docker-proxy
+RUN --mount=type=cache,target=/root/.cache/go-build \
+    CGO_ENABLED=0 GOOS=linux \
+    go build -v -o soperator-docker-proxy ./cmd/soperator-docker-proxy
```
rdjjke, line 15: Why is the build command different in sshd.dockerfile and slurmd.dockerfile?

helm/soperator-custom-configmaps/config-files/supervisord.conf
rdjjke, line 1: It seems like this "custom configmap" is now identical to the default supervisord configuration, rendered by Soperator.
https://github.com/nebius/soperator/pull/2893/changes#diff-7d981bdd8096c8bbdf86b9656bbd373645365ae39266a32f734c50af94138e73R96

So we probably can entirely get rid of this custom configmap.

Also, your new login pods also run supervisord and there the configuration is defined in login images: https://github.com/nebius/soperator/pull/2893/changes#diff-ad4c97d01ea2786f7394487d4f5bc0b6666054e380cbfbf5a5285c62caa441b4R1

So there is a drift between how it works on login and worker nodes.
What about storing this configuration in both login and worker images? I think this approach is cleaner.

e2e/acceptance/features/docker_containers.feature
```diff
@@ -1,4 +1,11 @@
 Feature: Docker containers
+  @soperator_version_>=5.0.0
+  Scenario: Docker containers can run over login SSH
```
rdjjke, line 3: Maybe it's worth to add a scenario for testing Docker containers over worker SSH to have the full coverage. Not a reason to block the PR though

e2e/acceptance/internal/sharedsteps/docker_containers.go
```diff
@@ -70,6 +114,178 @@ func (s *DockerContainers) CleanupAndReset(ctx context.Context) {
...
+		"-n",
+		framework.SoperatorNamespace,
+		"-o",
+		"json",
+	); err != nil {
+		return fmt.Errorf("get SlurmCluster for login Docker validation: %w", err)
+	}
+	if !cluster.Spec.SlurmNodes.Login.Docker.Enabled {
+		s.runtime.Logf("login Docker is disabled, skipping scenario")
+		return godog.ErrSkip
```
ChessProfessor, line 137: Do we have E2E profile that enables login Docker? It is disabled by default and I don't see E2E configuration enabling it. It seems like this scenario will always be skipped

e2e/acceptance/internal/sharedsteps/docker_containers.go
```diff
@@ -58,6 +82,26 @@ func (s *DockerContainers) RegisterSteps(sc *godog.ScenarioContext) {
 }
 
 func (s *DockerContainers) CleanupAndReset(ctx context.Context) {
+	if s.loginSSHContainerName != "" && s.loginSSHIdentitySet {
+		cleanupCommand := fmt.Sprintf(
+			"docker rm -f %s >/dev/null 2>&1 || true",
+			framework.ShellQuote(s.loginSSHContainerName),
+		)
+		if _, cleanupErr := s.runLoginDockerSSHCommand(ctx, cleanupCommand); cleanupErr != nil {
```
ChessProfessor: Can we execute `docker rm -f` via `s.runtime.Jail().Run(...)` instead? That would make the cleanup simpler

e2e/acceptance/internal/sharedsteps/docker_containers.go
```diff
@@ -70,6 +114,178 @@ func (s *DockerContainers) CleanupAndReset(ctx context.Context) {
...
 	s.containerNamePrefix = ""
 	s.connectionWorker = framework.WorkerInfo{}
+	s.loginSSHContainerName = ""
+	s.loginSSHIdentitySet = false
+	s.loginSSHOutput = ""
+}
+
+func (s *DockerContainers) loginDockerIsEnabled(ctx context.Context) error {
+	var cluster kubeobjects.SlurmCluster
+	if err := s.kubectl.GetJSON(
```
ChessProfessor: We have `KubectlClient.SlurmCluster`, which is used by the accounting and Slurm-config tests. Could we add `LoginDockerEnabled` to `framework.SlurmClusterInfo`, populate it in that helper, and use it here?

e2e/acceptance/internal/sharedsteps/docker_containers.go
```diff
@@ -70,6 +114,178 @@ func (s *DockerContainers) CleanupAndReset(ctx context.Context) {
...
+	return ensureSSHTestIdentity(
+		ctx,
+		s.runtime,
+		dockerSSHUserName,
+		dockerSSHKeyName,
+		dockerSSHKeyComment,
+	)
+}
+
+func (s *DockerContainers) theUserRunsDockerLifecycleCommandsOverSSHOnTheLoginNode(ctx context.Context) error {
```
ChessProfessor, line 158: Could we narrow this script to the integration contracts that need E2E coverage? A detached container is needed for cgroup inspection, and one `docker exec` is enough to exercise the upgraded streaming path.

Maybe something like that?

  ```
  remoteCommand := fmt.Sprintf(`
  set -euo pipefail
  name=%s
  image=%s
  cleanup() {
      docker rm -f "${name}" >/dev/null 2>&1 || true
  }
  trap cleanup EXIT

  echo "DOCKER_ROOT=$(docker info --format '{{.DockerRootDir}}')"

  if docker -H unix:///run/soperator-dockerd.sock version >/dev/null 2>&1; then
      echo "private Docker socket is accessible" >&2
      exit 1
  fi

  uid=$(id -u)
  echo "EXPECTED_SESSION_CGROUP=/users/user-${uid}/sessions"
  echo "SESSION_CGROUP=$(sed -n 's/^0:://p' /proc/self/cgroup)"

  docker run -d \
      --name "${name}" \
      --cgroup-parent=/escape \
      "${image}" sleep 300 >/dev/null

  container_pid=$(docker inspect --format '{{.State.Pid}}' "${name}")
  echo "EXPECTED_USER_CGROUP=/users/user-${uid}/docker/"
  echo "CONTAINER_CGROUP=$(sed -n 's/^0:://p' "/proc/${container_pid}/cgroup")"

  docker exec "${name}" true
  `, framework.ShellQuote(s.loginSSHContainerName), framework.ShellQuote(dockerLifecycleImage))
  ```

The following assertion step would then only need to parse `DOCKER_ROOT`, `SESSION_CGROUP`, and `CONTAINER_CGROUP`; private-socket access and `docker exec` are asserted directly through their exit status.

If we need to test `docker build` over login SSH we can have separate scenario for that?

images/worker/slurmd.dockerfile
aydinmirac: Thanks for the catch, just an overlooked mismatch. I aligned the worker proxy build command with the login image by adding -trimpath and -ldflags='-s -w'. Both amd64 and arm64 image builds now pass.

images/worker/supervisord.conf
aydinmirac: Agreed. The default worker Supervisor configuration is now stored in images/worker/supervisord.conf and copied into the worker image, matching the login-node approach. 

The corresponding solutions-library reference will be removed in a separate PR.

e2e/acceptance/features/docker_containers.feature
aydinmirac: Added a minimal worker SSH scenario. It connects through the existing login-to-worker SSH path and verifies that a disposable Docker container runs successfully.

e2e/acceptance/internal/sharedsteps/docker_containers.go
aydinmirac: You’re right: the current repository E2E profile does not enable login Docker, so this scenario is skipped in the standard run. Enabling it also requires per-login writable /mnt/image-storage provisioning and user isolation, which are owned by the solutions-library deployment profile. Is it ok to fix this problem in `nebius-solutions-library` ?

e2e/acceptance/internal/sharedsteps/docker_containers.go
aydinmirac: Done. Cleanup now executes `docker rm -f` directly through `s.runtime.Jail().Run(...)`

e2e/acceptance/internal/sharedsteps/docker_containers.go
aydinmirac: Done. I added `LoginDockerEnabled` to `framework.SlurmClusterInfo`, populate it in `KubectlClient.SlurmCluster`, and use that helper in the Docker step.

e2e/acceptance/internal/sharedsteps/docker_containers.go
aydinmirac: Thanks. implemented as you suggested.

e2e/acceptance/internal/sharedsteps/docker_containers.go
```diff
@@ -386,6 +626,20 @@ func graphDriverPathsUnder(output, root string) bool {
 	return foundPath
 }
 
+func dockerCgroupParentBelongsToJob(value, jobID string) bool {
```
ChessProfessor, line 633: My agent says that Slurm 26.05 cgroup validation will always fail here.

> `dockerCgroupParentBelongsToJob` requires a `job_<SLURM_JOB_ID>` component. Slurm 26.05 uses an opaque SLUID path such as `slurmstepd.scope/s8G5M22WGXB100/step_0/user`, which this PR explicitly supports and 5.0 uses by default. The lifecycle scenario will therefore fail when actually executed on 26.05.

aydinmirac: thanks for the catch. I updated the assertion to support both the legacy `job_<jobID>/step_*/user` hierarchy and Slurm 26.05’s strict `slurmstepd.scope/<SLUID>/step_*/user` hierarchy.

rdjjke (approved):

ChessProfessor (approved):

PR #2901, opened 2026-08-26
SCHED-682: Make Enroot work over SSH

aydinmirac:
## Summary
Make Enroot work from SSH sessions on both login and worker nodes.

OpenSSH previously used `ChrootDirectory /mnt/jail`. A process inside that chroot cannot create the user namespace Enroot requires, so `enroot start` failed with `Operation not permitted`.

This PR replaces the SSH session chroot with a required PAM module that:
 - Creates a private mount namespace for each SSH session.
 - Pivots the session into `/mnt/jail`.
 - Detaches the old root.
 - Recreates `/proc` inside the session namespace.
 - Fails the SSH session if jail setup fails.

The login-only per-user cgroup PAM hook remains separate and runs before the jail module. Worker SSH uses only the jail module.

## Rollout safety
The stored SSH ConfigMap retains `ChrootDirectory /mnt/jail` so an older image remains jailed during an upgrade.

A new image:
 1. Copies the SSH configuration into a private runtime directory.
 2. Removes `ChrootDirectory` from that copy.
 3. Validates it with `sshd -t`.
 4. Mounts it over `/mnt/ssh-configs` before starting SSHD.

This is necessary because image and ConfigMap updates are reconciled independently. A live upgrade reproduced this ordering: pods and configuration converged in separate steps.

## Tests
Validated using the final CI images through a normal Terraform/Flux upgrade:

 - Login and worker pods became Ready with zero restarts.
 - Stored configuration retained `ChrootDirectory /mnt/jail`.
 - Effective configuration reported `chrootdirectory none`.
 - Enroot created and started containers over direct SSH on login and worker nodes.
 - Enroot also worked through the normal login-to-worker SSH route.
 - SCP worked on both node types.
 - Enroot entered different user and mount namespaces from the SSH session while preserving the user UID.
 - An injected PAM failure immediately after `unshare()` prevented the requested command from running and returned SSH exit code 254.
 - An existing Slurm/Pyxis container job still used the requested SquashFS content successfully.
 - The added E2E scenario passed for both login and worker SSH.


helm/soperator-custom-configmaps/config-files/supervisord.conf
```diff
@@ -27,7 +27,7 @@ stdout_logfile_maxbytes=0
 stderr_logfile=/dev/fd/2
 stderr_logfile_maxbytes=0
 redirect_stderr=true
-command=/usr/sbin/sshd -D -e -f /mnt/ssh-configs/sshd_config
+command=/bin/sh -c 'if [ -x /opt/bin/slurm/sshd_pam_jail_entrypoint.sh ]; then exec /opt/bin/slurm/sshd_pam_jail_entrypoint.sh; else exec /usr/sbin/sshd -D -e -f /mnt/ssh-configs/sshd_config; fi'
```
rdjjke: I think we don't need this `if`. This change will go to the new major release of Soperator, where supervisord.conf will be updated together with login & worker container images.
It's not that important, but just for the sake of simplicity.

images/login/sshd.dockerfile
```diff
@@ -2,6 +2,20 @@
 
 ARG SLURM_VERSION
 
+# https://github.com/nebius/ml-containers/pull/99
+FROM cr.eu-north1.nebius.cloud/ml-containers/slurm:${SLURM_VERSION}-20260816173636 AS login_pam_builder
```
rdjjke: It uses the Slurm image just to build the pam module. I think you just copied the base image from the next `login_sshd` build stage.
But it's not actually needed to use the Slurm image here, and using it can create confusion like this build somehow depends on Slurm and some people will bump it on new Slurm releases, some will not, and so on.
I suggest using a more base image, "neubuntu", which is just a pure Ubuntu.

It's already used in this repo in the `munge.dockerfile` for example.
```
# https://github.com/nebius/ml-containers/pull/90
FROM cr.eu-north1.nebius.cloud/ml-containers/neubuntu:noble-20260624084749 AS munge
```

images/worker/slurmd.dockerfile
```diff
@@ -2,6 +2,20 @@
 
 ARG SLURM_VERSION
 
+# https://github.com/nebius/ml-containers/pull/99
+FROM cr.eu-north1.nebius.cloud/ml-containers/slurm:${SLURM_VERSION}-20260816173636 AS worker_pam_builder
```
rdjjke: Same here
https://github.com/nebius/soperator/pull/2901/changes#r3871315968

images/common/scripts/sshd_pam_jail_entrypoint.sh
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
+
+set -euo pipefail
+
+source_config="${1:-/mnt/ssh-configs/sshd_config}"
+effective_config="${2:-/run/soperator-sshd_config}"
+
+# Keep ChrootDirectory in the rendered config so older images remain jailed
```
rdjjke: It's great that you thought about backward compatibility. But IMO we don't need it here. This feature will go to a major release, and we don't have requirements for backward compatibility for them.

internal/render/worker/configmap.go
```diff
@@ -85,7 +85,7 @@ func generateDefaultSupervisordConfig() renderutils.ConfigFile {
 	res.AddLine("stderr_logfile=/dev/fd/2")
 	res.AddLine("stderr_logfile_maxbytes=0")
 	res.AddLine("redirect_stderr=true")
-	res.AddLine("command=/usr/sbin/sshd -D -e -f /mnt/ssh-configs/sshd_config")
+	res.AddLine("command=/bin/sh -c 'if [ -x /opt/bin/slurm/sshd_pam_jail_entrypoint.sh ]; then exec /opt/bin/slurm/sshd_pam_jail_entrypoint.sh; else exec /usr/sbin/sshd -D -e -f /mnt/ssh-configs/sshd_config; fi'")
```
rdjjke: Same as here -- https://github.com/nebius/soperator/pull/2901/changes#r3871255821

rdjjke:
What about adding a new scenario to our E2E tests that checks that Enroot works (pulling an image and starting a container) over SSH on login & worker nodes?

I think it can be added to the existing `enroot_containers.feature` file -- https://github.com/nebius/soperator/blob/main/e2e/acceptance/features/enroot_containers.feature
With `@soperator_version_>=5.0.0` tag.

images/common/pam_soperator_jail.c
```diff
@@ -0,0 +1,251 @@
...
+        return jail_errno(context, "detach inherited proc mount");
+    }
+    if (mount("proc", "/proc", "proc", 0, NULL) != 0) {
+        return jail_errno(context, "mount proc in jail");
+    }
+
+    return PAM_SUCCESS;
+}
+
+PAM_EXTERN int pam_sm_open_session(
```
rdjjke, line 205: My AI spotted one potential problem with it.
This hook is executed in the forked per-session sshd process. So, in the following example:
```
ps auxf
USER         PID %CPU %MEM    VSZ   RSS TTY      STAT START   TIME COMMAND
root           1  0.0  0.0  12152  8208 ?        Ss   Aug24   0:18 sshd: /usr/sbin/sshd -D -e -f /run/soperator-sshd_config [listener] 0 of 100-300 startups
root       35640  0.0  0.0  14792 10132 ?        Ss   12:52   0:00 sshd: root@pts/1
root       35689  0.0  0.0   8908  5332 pts/1    Ss+  12:52   0:00  \_ -bash
```
It will change the root for the process `35640`.

ChrootDirectory setting works differently -- it would only change the root directory of the process `35689`.

So any code that runs inside `sshd` after this hook (i.e. other modules) will also run inside the jail. Theoretically, it shouldn't be an issue, because SSSD PAM module is installed into both container and jail, and various dirs like `/proc`, `/dev`, `/sys`, `/run`, `/tmp` are bind-mounted from container to jail and should be identical.
But I'm not sure whether this implementation works correctly with the previous PAM module you added for per-user cgroups, so it's at least worth checking.

In any case, if it's possible to change the PAM hook so that the only user shell process changes it roots, it would be safer.

rdjjke: I asked the AI where we should put this logic instead, and it eventually said that probably the current approach is the best.
There's a hook `do_pam_setcred` that can be used for that, but it introduces another long set of problems.
So what it suggests now is to just make sure that the `pam_soperator_jail` module is always last in the list, and add a comment that the reason is that all modules located after it run jailed.

images/common/pam_soperator_jail.c
```diff
@@ -0,0 +1,251 @@
...
+        message
+    );
+}
+
+static int jail_errno(const struct jail_context *context, const char *operation)
+{
+    const int saved_errno = errno;
+
+    jail_log(context, "%s: %s", operation, strerror(saved_errno));
+    return PAM_SESSION_ERR;
```
rdjjke, line 53: All failures in the code are handled by returning `PAM_SESSION_ERR`.

AI says that `PAM_SESSION_ERR` leads to error logging but doesn't stop the session. So it can lead to SSH sessions being created out of jail or in a mixed environment (some operations are completed, but some aren't).

So it's worth investigating what actually happens if this module ends with errors after the `unshare` call is made (the first operation that modifies the environment and shouldn't lead to a successfully started SSH session).

helm/soperator-custom-configmaps/config-files/supervisord.conf
aydinmirac: Done. The fallback `if` was removed from both the shared Supervisor configuration and the worker renderer. Supervisor now starts SSHD directly.

internal/render/worker/configmap.go
aydinmirac: Done. Removed

images/login/sshd.dockerfile
aydinmirac: Done. The PAM build stage now uses the pinned Noble `neubuntu` image and installs only the compiler and PAM development headers required for the build.

The final image receives only the compiled PAM module from the builder stage.

images/worker/slurmd.dockerfile
aydinmirac: Done here as well. Login and worker now use the same pinned `neubuntu` builder and shared PAM build script.

images/common/scripts/sshd_pam_jail_entrypoint.sh
aydinmirac: I initially removed it, but live upgrade testing exposed an ordering problem.

The worker image started before the independently reconciled SSH ConfigMap had converged. SSHD therefore retained the old `ChrootDirectory` behavior, and Enroot over SSH failed until the pod was recreated manually.

The final implementation keeps a small rollout guard:

  - The stored ConfigMap retains `ChrootDirectory` for an older image.
  - A new image removes it only from a private runtime copy before starting SSHD.
  - The runtime configuration is validated with `sshd -t`.

A normal Terraform/Flux upgrade with the final image then completed without manual pod recreation. The stored configuration contained the fallback, while both running daemons reported `chrootdirectory none`.

images/common/pam_soperator_jail.c
aydinmirac: Agreed. I kept this in `pam_sm_open_session` and made `pam_soperator_jail` the last session module in both login and worker PAM stacks.

I also added a comment explaining that the per-session SSHD process pivots into the jail at this point, so any modules placed after it would run inside the jail.

On login nodes, the per-user cgroup hook runs before it. Worker nodes only add the jail module.

images/common/pam_soperator_jail.c
aydinmirac: I verified this with a test-only failure injected immediately after `unshare(CLONE_NEWNS)` succeeded.

Authentication completed, but `pam_open_session` returned an error. OpenSSH did not execute the requested command, returned exit code 254, and logged the PAM session failure. The marker file that the requested command would have created remained absent.

I also confirmed the same fail-closed behavior by configuring an invalid jail path.

Therefore, with this module configured as `session required`, `PAM_SESSION_ERR` prevents the SSH session from opening, including when the error occurs after the mount namespace has already been modified.

aydinmirac:
> What about adding a new scenario to our E2E tests that checks that Enroot works (pulling an image and starting a container) over SSH on login & worker nodes?
> 
> I think it can be added to the existing `enroot_containers.feature` file -- https://github.com/nebius/soperator/blob/main/e2e/acceptance/features/enroot_containers.feature With `@soperator_version_>=5.0.0` tag.

Done. I added a scenario outline to `enroot_containers.feature` with the `@soperator_version_>=5.0.0` tag and examples for both login and worker nodes.

The scenario uses a real SSH connection and verifies that:

  - The SSH session is inside the jail.
  - User namespace creation succeeds.
  - Enroot imports, creates, and starts an image.
  - The Enroot process preserves the SSH user UID.
  - Enroot creates separate user and mount namespaces.
  - Temporary SSH credentials and Enroot resources are cleaned up.

I also ran the scenario against the test cluster: both login and worker examples passed.

images/common/scripts/sshd_pam_jail_entrypoint.sh
rdjjke: OK. I created a task for the next major release to remove this backward compatibility logic -- SCHED-2385

ChessProfessor:
I started acceptance - https://github.com/nebius/soperator/actions/runs/33190461403

e2e/acceptance/internal/sharedsteps/enroot_containers.go
```diff
@@ -79,6 +90,165 @@ func (s *EnrootContainers) CleanupAndReset(ctx context.Context) {
...
+		framework.ShellQuote(host),
+		framework.ShellQuote(remoteCommand),
+	)
+	command := fmt.Sprintf(
+		"su - %s -c %s",
+		framework.ShellQuote(sshUserName),
+		framework.ShellQuote(sshCommand),
+	)
+
+	output, err := s.runtime.Jail().RunWithDefaultRetry(ctx, command)
```
ChessProfessor: We can use Run here without retry I believe

e2e/acceptance/internal/sharedsteps/enroot_containers.go
```diff
@@ -79,6 +90,165 @@ func (s *EnrootContainers) CleanupAndReset(ctx context.Context) {
...
+			"unexpected SSH Enroot output %q, expected %q",
+			strings.TrimSpace(s.sshEnrootOutput),
+			expected,
+		)
+	}
+
+	return nil
+}
+
+func enrootSSHSmokeScript(target string) string {
```
ChessProfessor: We have huge bash script that checks something inside (like test "${root_source}" = "jail"), but we better do these checks in go with minimal bash. Can we somehow simplify this part?

For example we can use simpler script and verify behaviour outside of it?
```
  set -euo pipefail

  workdir="$(mktemp -d /tmp/soperator-enroot-ssh.XXXXXX)"
  trap 'rm -rf -- "${workdir}"' EXIT

  enroot import --output "${workdir}/image.sqsh" IMAGE
  enroot start "${workdir}/image.sqsh" grep '^ID=ubuntu$' /etc/os-release
```

e2e/acceptance/internal/sharedsteps/enroot_containers.go
```diff
@@ -79,6 +90,165 @@ func (s *EnrootContainers) CleanupAndReset(ctx context.Context) {
...
+		if cleanupErr := s.removeEnrootSSHTestIdentity(ctx); cleanupErr != nil {
+			s.runtime.Logf("cleanup: remove Enroot SSH test identity: %v", cleanupErr)
+		}
+	}
+	s.sshEnrootOutput = ""
+	s.sshEnrootTarget = ""
+	s.sshIdentitySet = false
+}
+
+func (s *EnrootContainers) anEnrootSSHTestUserExists(ctx context.Context) error {
```
ChessProfessor, line 113: We can reuse the code from ssh test for user creation. Create a common helper and use in both places where we do `createuser`.

e2e/acceptance/features/enroot_containers.feature
```diff
@@ -1,4 +1,15 @@
 Feature: Enroot containers
+  @soperator_version_>=5.0.0
+  Scenario Outline: Enroot containers can run over SSH
+    Given an Enroot SSH test user exists
+    When the user imports and starts an Enroot container over SSH on a <node> node
+    Then the SSH Enroot smoke test succeeds
```
ChessProfessor: Maybe we can do this to simplify the code a little:

```
  Given an Enroot SSH test user exists
  When the user imports an Enroot image over SSH on a <node> node
  And the user starts the imported Enroot image over SSH
  Then the container reports Ubuntu
```

e2e/acceptance/internal/sharedsteps/enroot_containers.go
aydinmirac: Done. Replaced `RunWithDefaultRetry` with `Run`, so SSH import/start failures are reported immediately instead of retrying the complete operation.

e2e/acceptance/internal/sharedsteps/enroot_containers.go
aydinmirac: Done. Removed the large Bash smoke-test script. The test now uses small commands to import and start the image, while Go parses `/etc/os-release` and verifies `ID=ubuntu`.

e2e/acceptance/internal/sharedsteps/enroot_containers.go
aydinmirac: Done. Extracted the shared `createuser` logic into `ensureSSHTestUser` and reused it from both the internal SSH and Enroot SSH scenarios.

e2e/acceptance/features/enroot_containers.feature
aydinmirac: Done. Split the scenario into separate image import, image start, and Ubuntu validation steps for both login and worker nodes.

aydinmirac: I also ran the revised acceptance scenario against the test cluster. Both login and worker examples passed: 2 scenarios and 8 steps.

  - `ensureSSHTestUser`.
  - Temporary SSH key creation.
  - Login and worker selection.
  - Separate import and start Go steps.
  - State propagation of the selected host and image path.
  - Run without workflow retries.
  - Go parsing of `/etc/os-release`.
  - `CleanupAndReset`.

rdjjke (approved):

ChessProfessor (approved):

PR #2909, opened 2026-08-31
SCHED-2384: Add acceptance tests for multiple Slurm topologies

Uburro:


ChessProfessor:
In this PR we throw godog.ErrSkip late into the scenarios, and do unnecessary work sometimes. I suggest first checking for the required  configuration and skip immediately if we do not have it. After that only fail if something goes wrong.

Suggested split:

```
### topology.feature

Feature: Slurm named topologies

  Background:
    Given the cluster is configured with several named topologies

  @multi_topology @soperator_version_>=5.0.0
  Scenario: The operator publishes and Slurm loads the configured topologies
    Given the operator published the topology config
    Then the config declares several named topologies
    And exactly one of them is the cluster default
    And one of them is a flat topology
    And the topology config is delivered to the jail
    When Slurm is asked which topologies it loaded
    Then Slurm loaded exactly the topologies the operator rendered

  @multi_topology @soperator_version_>=5.0.0
  Scenario: Partitions and workers use their configured topologies
    Given the operator published the topology config
    When Slurm is asked which topologies it loaded
    Then every partition is bound to a topology Slurm loaded
    And every GPU worker is listed by a fabric topology
    And no CPU-only worker is listed by a fabric topology
    And every running worker is registered into the topologies that list it

  @multi_topology @soperator_version_>=5.0.0
  Scenario: Every configured topology can schedule a job
    Given the operator published the topology config
    Then a job runs in a partition of every topology

### topology_tree.feature

Feature: Slurm tree-topology scheduling

  @gpu @tree_topology @soperator_version_>=5.0.0
  Scenario: A switch limit rejects a cross-leaf allocation
    Given the cluster is configured with a tree topology spanning multiple leaf switches
    And the operator published the topology config
    And Slurm loaded the tree topology
    When two configured workers on different leaf switches are requested without a switch limit
    Then the cross-leaf allocation succeeds
    When the same workers are requested with at most one switch
    Then the cross-leaf allocation is rejected

### topology_block.feature

Feature: Slurm block-topology scheduling

  @gpu @block_topology @soperator_version_>=5.0.0
  Scenario: A block topology schedules jobs after its base size changes
    Given the cluster is configured with a block topology whose base size can be changed
    And the operator published the topology config
    Then Slurm reports the blocks the operator rendered
    And a job runs in a partition of the block topology
    When the block topology base size is halved
    Then the rendered config carries the new block sizes
    And Slurm reports the new base block size
    And the topology JailedConfig has no pending reconfigure request
    And a job runs in a partition of the block topology
```

The first step of each scenario inspect desired configuration:

- Required capability not configured → skip immediately.
- Capability configured but resources are missing, unhealthy, or unusable → fail.

The tests are currently failing for clusters without topology, so this split will also solve the problem.

e2e/acceptance/internal/sharedsteps/topology.go
```diff
@@ -0,0 +1,1163 @@
...
+}
+
+func (s *Topology) runJobInTopology(ctx context.Context, topology string) error {
+	partition, node, ok, err := s.pickReadyNodeOfTopology(ctx, topology)
+	if err != nil {
+		return err
+	}
+	if !ok {
+		s.runtime.Logf("acceptance: no partition of topology %s has a running worker, skipping its job", topology)
+		return nil
```
ChessProfessor: It's it a failure? Maybe we should throw here?

e2e/acceptance/internal/sharedsteps/topology.go
```diff
@@ -0,0 +1,1163 @@
...
+	sc.Step(`^the topology JailedConfig has no pending reconfigure request$`, s.noPendingReconfigureRequest)
+}
+
+func (s *Topology) CleanupAndReset(ctx context.Context) {
+	if s.blockSizesRestore != nil {
+		cleanupCtx, cancel := context.WithTimeout(ctx, topologyCleanupTimeout)
+		defer cancel()
+
+		sizes := s.blockSizesRestore
+		s.blockSizesRestore = nil
```
ChessProfessor, line 164: We can do that in the end after block sizes restored

e2e/acceptance/internal/sharedsteps/topology_parse.go
```diff
@@ -0,0 +1,431 @@
...
+}
+
+// halveBlockSizes returns the block sizes to test a reconfigure with. Halving the base keeps it a
+// power of two and never asks for blocks larger than the ones the cluster already has, which
+// slurmctld would reject.
+func halveBlockSizes(sizes []int) ([]int, bool) {
+	if len(sizes) == 0 || sizes[0] < 2 {
+		return nil, false
+	}
+	return []int{sizes[0] / 2}, true
```
ChessProfessor: Shouldn't we return the whole sizes with halved sizes[0] here?

e2e/acceptance/internal/sharedsteps/topology.go
```diff
@@ -0,0 +1,1163 @@
...
+func (s *Topology) readJailedConfig(ctx context.Context) (kubeobjects.JailedConfig, error) {
+	var config kubeobjects.JailedConfig
+	if err := s.kubectl.GetJSON(ctx, &config,
+		"get", "jailedconfig", s.configName, "-n", framework.SoperatorNamespace, "-o", "json"); err != nil {
+		return kubeobjects.JailedConfig{}, fmt.Errorf("get JailedConfig %s: %w", s.configName, err)
+	}
+	return config, nil
+}
+
+func (s *Topology) waitForRenderedBlockSizes(ctx context.Context, topology string, sizes []int) error {
```
ChessProfessor, line 1221: From my agent:
```
  > The operator intentionally updates the ConfigMap before the JailedConfig annotation. Waiting for the ConfigMap and
  > then reading the annotation once can observe this valid intermediate state. Please poll until both the rendered
  > configuration and the JailedConfig annotation have converged.
```

e2e/acceptance/internal/sharedsteps/topology.go
```diff
@@ -0,0 +1,1411 @@
...
+
+	if err := s.loadRenderedConfig(ctx); err != nil {
+		return err
+	}
+	for _, name := range configured {
+		entry, ok := topologyByName(s.rendered, name)
+		if !ok {
+			return fmt.Errorf("tree topology %s is configured but missing from the published config", name)
+		}
+		if len(entry.Switches) > 1 {
```
ChessProfessor: Question: Might there be only one root and one leaf nodes here? As I understand we are checking for two leaf nodes here? Or it's not valid configuration?

e2e/acceptance/internal/sharedsteps/topology.go
```diff
@@ -0,0 +1,1411 @@
...
+	if run.Err == nil {
+		return nil
+	}
+	pair := *s.crossLeaf
+	if srunRefused(run.Output, run.Err) {
+		// Nothing about the topology can be concluded while the pair cannot be allocated at all,
+		// and a cluster busy with someone else's job is not a defect of the topology.
+		s.runtime.Logf("acceptance: workers %s and %s cannot be allocated at all right now, skipping the switch limit check",
+			pair.Nodes[0], pair.Nodes[1])
+		return godog.ErrSkip
```
ChessProfessor: Maybe we should throw a real error here? We validated already in this test that suitable tree topology is configured, if we cannot submit a job it means something goes wrong? Maybe we should add small timeout for workers to become allocatable and then fail if not?

e2e/acceptance/features/topology.feature
```diff
@@ -0,0 +1,28 @@
...
+
+  @multi_topology @soperator_version_>=5.0.0
+  Scenario: The operator publishes and Slurm loads the configured topologies
+    Given the operator published the topology config
+    Then the config declares several named topologies
+    And exactly one of them is the cluster default
+    And one of them is a flat topology
+    And the topology config is delivered to the jail
+    When Slurm is asked which topologies it loaded
+    Then Slurm loaded exactly the topologies the operator rendered
```
ChessProfessor, line 14: nit: We kinda check only snapshots of the topologies here, so it's not "loaded exactly". We can rename it to "Then Slurm loaded the rendered topology definitions"

e2e/acceptance/internal/sharedsteps/topology.go
Uburro: This is possible. We cannot guarantee that the two nodes will be on different branches of the tree.

ChessProfessor (approved):

PR #2917, opened 2026-09-01
SCHED-1858: Add login node autoscaling

Ronsenval:
## Problem

[RFC 055: Login Node Resource Fairness and Autoscaling](https://nebius.atlassian.net/wiki/spaces/SCHED/pages/2212561681/RFC+055+Login+Node+Resource+Fairness+and+Autoscaling)

Login capacity is currently fixed by `slurmNodes.login.size`. Clusters must therefore provision enough login pods for peak SSH workload or risk CPU contention during load spikes.

## Solution

Add optional CPU-based autoscaling for login pods through `slurmNodes.login.autoscaling`.

### Pod autoscaling

Autoscaling is omitted and disabled by default. When configured and enabled, the Soperator controller creates an HPA that:

- Scales login pods between `minReplicas` and `maxReplicas`.
- Uses average SSHD container CPU utilization across eligible login pods.
- Uses a target CPU utilization of 70% by default.
- Disables automatic scale-down to avoid terminating active login sessions.

### Replica ownership and scale-down

While autoscaling is enabled, regular SlurmCluster reconciliation preserves the replica count managed by the HPA. Changes to `slurmNodes.login.size` are retained but do not override the live replica count.

To reduce the number of login pods, users can lower `autoscaling.maxReplicas` below the current replica count or disable autoscaling. Disabling autoscaling removes the HPA and restores `slurmNodes.login.size`. StatefulSet scale-down removes the highest ordinals first; deleting a pod directly does not scale down the workload because the StatefulSet creates a replacement.

### Kubernetes node capacity

The dedicated Managed Kubernetes login node group uses the Nebius cluster autoscaler with broad 1–100 node infrastructure guardrails. Effective login workload limits are defined on the Soperator side through `minReplicas` and `maxReplicas`, which the controller reconciles into the HPA; the node-group range is not the desired login capacity.

### Compatibility

Login pods remain managed by the existing OpenKruise Advanced StatefulSet and retain stable ordinal names such as `soperator-login-0`.

Terraform integration: [nebius-solutions-library#1156](https://github.com/nebius/nebius-solutions-library/pull/1156)

## Testing

The E2E autoscaling lifecycle verifies:

- No scaling without CPU pressure.
- Login pod and Kubernetes node-group scale-up under SSHD CPU pressure.
- No automatic scale-down after pressure is removed.
- Preservation of autoscaled pods when the fixed login size changes.
- Removal of the HPA and return to the fixed replica count when autoscaling is disabled.

The scenario is limited to Soperator 5.0.0 and later with the `@soperator_version_>=5.0.0` tag.

Full E2E validation passed:

- [Workflow run 33848508000](https://github.com/nebius/soperator/actions/runs/33848508000) / [MDC_B200 job](https://github.com/nebius/soperator/actions/runs/33848508000/job/100945948954)
- Soperator commit: [`b2c93638`](https://github.com/nebius/soperator/commit/b2c936380ee6b7761b142e68b10dac1e1b5d0736)
- Terraform integration commit: [`e4e0b7e1`](https://github.com/nebius/nebius-solutions-library/commit/e4e0b7e1)
- Terraform apply, the full acceptance suite, and Terraform destroy completed successfully.
- `features/login_autoscaling.feature` passed: the login workload scaled from 2 to 3 replicas under CPU pressure, Kubernetes added a node, replicas were preserved after pressure removal and a fixed-size change, and disabling autoscaling returned the workload to 2 replicas.

## Release Notes

Feature: Added optional CPU-based autoscaling for Slurm login pods. Autoscaling increases login capacity under average SSHD CPU load without automatically scaling down active login pods.


internal/webhook/v1/slurmcluster_webhook.go
```diff
@@ -54,14 +55,44 @@ var _ admission.Validator[*slurmv1.SlurmCluster] = &SlurmClusterCustomValidator{
...
 
 // ValidateUpdate implements admission.Validator so a webhook will be registered for the type SlurmCluster.
 func (v *SlurmClusterCustomValidator) ValidateUpdate(_ context.Context, _, newSlurmCluster *slurmv1.SlurmCluster) (admission.Warnings, error) {
 	slurmClusterLog.Info("Validation for SlurmCluster upon update", "name", newSlurmCluster.GetName())
 
-	return nil, validateLoginUserIsolation(newSlurmCluster)
+	return nil, validateSlurmCluster(newSlurmCluster)
+}
+
+func validateSlurmCluster(cluster *slurmv1.SlurmCluster) error {
```
Uburro, line 68: we can use ValidatingAdmissionPolicy without adding logic to webhook. it's old behaviour

internal/controller/clustercontroller/login.go
```diff
@@ -246,6 +247,27 @@ func (r SlurmClusterReconciler) ReconcileLogin(
...
+							return fmt.Errorf("cleaning up login HorizontalPodAutoscaler: %w", err)
+						}
+					}
+
+					desiredReplicas := ptr.To(clusterValues.NodeLogin.StatefulSet.Replicas)
+					if maintenanceActive {
+						desiredReplicas = ptr.To(consts.ZeroReplicas)
+					} else if autoscalingActive {
+						// A nil override tells the StatefulSet reconciler to preserve the live HPA-managed replica count.
+						desiredReplicas = nil
```
theyoprst: A bug found by Astra. Consider there autoscaling enabled, and a maintenance mode was turned on. Then desiredReplicas as set to zero (see 3 lines above). Then when maintenance mode is turned on, we set desiredReplicase to nil, which means "keep existing value", the replicase will still be 0. A fix suggested by an LLM: check the current STS replicas and if (and only if) it's zero, set it to min_replicas:

```go
  if maintenanceActive {
        desiredReplicas = ptr.To(consts.ZeroReplicas)
  } else if autoscalingActive {
        current := &kruisev1b1.StatefulSet{}
        err := r.Get(stepCtx, types.NamespacedName{
                Namespace: clusterValues.Namespace,
                Name:      clusterValues.NodeLogin.StatefulSet.Name,
        }, current)
        if err != nil && !apierrors.IsNotFound(err) {
                return fmt.Errorf("get login StatefulSet: %w", err)
        }

        if apierrors.IsNotFound(err) || ptr.Deref(current.Spec.Replicas, 0) == 0 {
                // CPU-based HPA cannot resume a workload scaled to zero.
                desiredReplicas = ptr.To(clusterValues.NodeLogin.Autoscaling.MinReplicas)
        } else {
                desiredReplicas = nil
        }
  }
```

e2e/acceptance/internal/sharedsteps/login_autoscaling.go
```diff
@@ -0,0 +1,654 @@
...
+	}
+	pods, err := s.loginPods(ctx)
+	if err != nil {
+		return err
+	}
+	nodes, err := s.k8sNodes(ctx)
+	if err != nil {
+		return err
+	}
+	newNode, err := loginPodOnNewNode(pods, nodes, s.initialK8sNodes)
```
theyoprst: This will not work on GB300: it's arm64 platform, and on this platform login pods are collocated on worker nodes. Meaning no new node will be created. I think the test should not check nodes at all, it's kind of implementation detail. Or maybe get all node ids for login nodes and check them instead.

internal/controller/clustercontroller/login.go
Ronsenval: Fixed https://github.com/nebius/soperator/pull/2917/commits/03684aeef181ae33bf742b0c83f7d9ca6e423f42

e2e/acceptance/internal/sharedsteps/login_autoscaling.go
Ronsenval: Removed assumptions on underling nodes in testing https://github.com/nebius/soperator/pull/2917/commits/92b98d67056f460640171851010e089560b5116d

Ronsenval: Also added doc as you suggested in Slack https://github.com/nebius/soperator/pull/2917/commits/39fbf19602521f83b1a8385a1b551a68bae97efe

theyoprst (approved):

internal/webhook/v1/slurmcluster_webhook.go
Ronsenval: Added validation as a part of CRD validation (x-kubernetes-validations) https://github.com/nebius/soperator/pull/2917/commits/7426494512e0204db95f50418d93a809c52d9b5a. Is it fine or ValidatingAdmissionPolicy (+ binding) is still a better choice here?

internal/controller/clustercontroller/login.go
theyoprst: Thanks

PR #2937, opened 2026-09-04
SCHED-2397: add per-command retry timeout

ChessProfessor:
## Problem

The reusable acceptance runtime imposes a fixed 10-minute timeout on every external command. Consumers cannot extend it for legitimately long operations, such as managed cluster topology updates, without retrying commands based on process error text.

## Solution

Add a per-attempt timeout argument to both RunWithRetry scope methods. RunWithDefaultRetry uses a fresh 10-minute timeout for every attempt. Run supplies a 10-minute timeout when its context has no deadline and otherwise respects the caller-provided deadline


ChessProfessor:
Closing in favor of polling managed operation status in managed-soperator-e2e. This avoids changing the shared command timeout API.

PR #2938, opened 2026-09-04
SCHED-2378: Login user isolation enabled

Ronsenval:
## Problem

Login user isolation is available, but it is disabled by default. Unless operators explicitly enable it, one user can monopolize login-node CPU or exhaust the memory available to the `sshd` container, affecting every other logged-in user.

The previous derived per-user memory defaults (`memory.high = 80%` and `memory.max = 90%` of the `sshd` container memory) also allow one user to consume nearly all login-node memory and do not leave useful capacity for a second user.

## Solution

- Enable `slurmNodes.login.userIsolation` by default.
- Derive `memory.high` at 40% and `memory.max` at 50% of the `sshd` container memory when explicit values are not provided.
- Add a black-box acceptance scenario that creates two regular login users and compares completed CPU work under two contention patterns:
  - both users run workloads at twice the login CPU capacity;
  - the first user runs at login CPU capacity while the second runs at four times capacity.
- Skip the acceptance scenario once at the top level when login user isolation is disabled.

## Testing

- `go test ./...`, `go vet ./...` — passed.
- [Build All in one job](https://github.com/nebius/soperator/actions/runs/33881965250) — passed, including lint, E2E build, images, Helm charts, and Helm integration tests.
- Manual validation on a personal dev cluster. Usage is each user's share of the total completed CPU work:

  | Isolation | First user workload | Second user workload | First user usage | Second user usage |
  | --- | ---: | ---: | ---: | ---: |
  | Enabled | `2x` login CPU capacity | `2x` login CPU capacity | `49.2%` | `50.8%` |
  | Enabled | `1x` login CPU capacity | `4x` login CPU capacity | `45.9%` | `54.1%` |
  | Disabled | `1x` login CPU capacity | `4x` login CPU capacity | `20.4%` | `79.6%` |

  The disabled-isolation row was a negative control run with the scenario skip bypassed only in the local runner; it failed as expected because the first user's share was outside the accepted `[35%, 65%]` range.
- [Full GitHub E2E](https://github.com/nebius/soperator/actions/runs/33886548071) — passed on the `MDC_B200` profile.

## Release Notes

Login-node user isolation is now enabled by default, with per-user memory thresholds derived at 40%/50% of the `sshd` container memory and proportional CPU sharing under contention. This changes the default behavior; clusters that need the previous behavior can explicitly set `slurmNodes.login.userIsolation.enabled: false`.

### DIFF
--- a/e2e/acceptance/features/login_user_isolation.feature
+++ b/e2e/acceptance/features/login_user_isolation.feature
@@ -1,0 +1,8 @@
+Feature: Login user isolation
+  @soperator_version_>=5.0.0
+  Scenario: Additional processes do not increase a login user's CPU share
+    Given two regular users can SSH to the login node for CPU isolation testing
+    When both users run workloads at twice the login CPU capacity
+    And the first user runs at login CPU capacity while the second runs at four times capacity
+    Then both users complete similar amounts of work in both cases
+    And the first user's amount of completed work remains stable
--- a/e2e/acceptance/internal/sharedsteps/login_user_isolation.go
+++ b/e2e/acceptance/internal/sharedsteps/login_user_isolation.go
@@ -1,0 +1,381 @@
+package sharedsteps
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/cucumber/godog"
+
+	"github.com/nebius/soperator/e2e/acceptance/framework"
+)
+
+const (
+	loginCPUFirstUser  = "soperatorcpua"
+	loginCPUSecondUser = "soperatorcpub"
+
+	loginCPUBurnMarker = "soperator-e2e-login-cpu-burn"
+
+	loginCPUWorkloadDuration = 8 * time.Second
+	loginCPUCommandTimeout   = 20 * time.Second
+
+	loginCPUMinShare            = 0.35
+	loginCPUMaxShare            = 0.65
+	loginCPUMaxWorkRelativeDiff = 0.20
+)
+
+type loginCPUWork struct {
+	first  uint64
+	second uint64
+}
+
+type loginCPUWorkloadResult struct {
+	userName string
+	work     uint64
+	err      error
+}
+
+type LoginUserIsolation struct {
+	info    *framework.ClusterInfo
+	runtime framework.Runtime
+
+	capacity       int
+	balancedWork   loginCPUWork
+	unbalancedWork loginCPUWork
+	workloadActive bool
+}
+
+func NewLoginUserIsolation(info *framework.ClusterInfo, runtime framework.Runtime) *LoginUserIsolation {
+	return &LoginUserIsolation{info: info, runtime: runtime}
+}
+
+func (s *LoginUserIsolation) RegisterSteps(sc *godog.ScenarioContext) {
+	sc.Step(`^two regular users can SSH to the login node for CPU isolation testing$`, s.twoRegularUsersCanSSHToTheLoginNode)
+	sc.Step(`^both users run workloads at twice the login CPU capacity$`, s.bothUsersRunAtTwiceCapacity)
+	sc.Step(`^the first user runs at login CPU capacity while the second runs at four times capacity$`, s.usersRunAtDifferentLoads)
+	sc.Step(`^both users complete similar amounts of work in both cases$`, s.bothUsersCompleteSimilarAmountsOfWork)
+	sc.Step(`^the first user's amount of completed work remains stable$`, s.firstUsersAmountOfWorkRemainsStable)
+}
+
+func (s *LoginUserIsolation) CleanupAndReset(ctx context.Context) {
+	if s.workloadActive {
+		if err := s.stopCPUWorkloads(ctx); err != nil {
+			s.runtime.Logf("cleanup: stop login CPU isolation workloads: %v", err)
+		}
+	}
+	s.capacity = 0
+	s.balancedWork = loginCPUWork{}
+	s.unbalancedWork = loginCPUWork{}
+	s.workloadActive = false
+}
+
+func (s *LoginUserIsolation) twoRegularUsersCanSSHToTheLoginNode(ctx context.Context) error {
+	enabled, err := s.loginUserIsolationIsEnabled(ctx)
+	if err != nil {
+		return err
+	}
+	if !enabled {
+		s.runtime.Logf("acceptance: login user isolation is disabled, skipping scenario")
+		return godog.ErrSkip
+	}
+
+	for _, userName := range []string{loginCPUFirstUser, loginCPUSecondUser} {
+		if err := ensureSSHTestUser(ctx, s.runtime, userName); err != nil {
+			return err
+		}
+		if err := s.verifyUserCanSSH(ctx, userName); err != nil {
+			return err
+		}
+	}
+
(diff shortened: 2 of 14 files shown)


e2e/acceptance/features/login_user_isolation.feature
```diff
@@ -0,0 +1,8 @@
+Feature: Login user isolation
+  @soperator_version_>=5.0.0
+  Scenario: Additional processes do not increase a login user's CPU share
+    Given two regular users can SSH to the login node for CPU isolation testing
+    When both users run workloads at twice the login CPU capacity
+    And the first user runs at login CPU capacity while the second runs at four times capacity
```
ChessProfessor: I don't understand this from description..
So both run at twice the capacity or the first runs at capacity and second 4 times capacity?

e2e/acceptance/internal/sharedsteps/login_user_isolation.go
```diff
@@ -0,0 +1,500 @@
...
+	capacity, err := s.loginCPUCapacity(ctx)
+	if err != nil {
+		return err
+	}
+	s.capacity = capacity
+	s.runtime.Logf("login CPU capacity: %d", capacity)
+	return nil
+}
+
+func (s *LoginUserIsolation) ensureSSHTestUser(ctx context.Context, userName string) error {
```
ChessProfessor: We already have ensureSSHTestUser in ssh_helpers.go. Could we reuse it and remove this duplicate?

e2e/acceptance/internal/sharedsteps/login_user_isolation.go
```diff
@@ -0,0 +1,500 @@
...
+func (s *LoginUserIsolation) loginCPUCapacity(ctx context.Context) (int, error) {
+	output, err := s.runtime.Kubectl().RunWithDefaultRetry(
+		ctx,
+		"get",
+		"slurmcluster",
+		s.info.SlurmClusterName,
+		"-n",
+		framework.SoperatorNamespace,
+		"-o",
+		"jsonpath={.spec.slurmNodes.login.sshd.resources.cpu}",
```
ChessProfessor: This reads the CPU request, but the login container has no CPU limit, so a 2-CPU request on an idle 16-CPU node can give a valid 20:80 split with 2 versus 8 processes. Could we use the effective capacity?

e2e/acceptance/internal/sharedsteps/login_user_isolation.go
```diff
@@ -0,0 +1,500 @@
...
+printf '%%s\n' "${count}"
+`, framework.ShellQuote(loginCPUBurnMarker), loginCPUWorkloadDuration.Seconds())
+	remoteScript := fmt.Sprintf(`
+set -euo pipefail
+workdir="$(mktemp -d /tmp/soperator-e2e-login-cpu.XXXXXX)"
+trap 'rm -r -- "${workdir}"' EXIT
+for ((i = 0; i < %d; i++)); do
+    bash -c %s >"${workdir}/${i}" &
+done
+wait
```
ChessProfessor: Bare wait returns success even when a background worker fails, and set -e does not catch this. Could we check each worker’s exit status?

e2e/acceptance/internal/sharedsteps/login_user_isolation.go
```diff
@@ -0,0 +1,500 @@
...
+		return fmt.Errorf(
+			"compare first user's login CPU work: relative difference %.3f exceeds %.2f",
+			difference,
+			loginCPUMaxWorkRelativeDiff,
+		)
+	}
+	return nil
+}
+
+func (s *LoginUserIsolation) prepareSSHIdentities(ctx context.Context) error {
```
ChessProfessor: createuser --without-external-ssh already generates ~/.ssh/id_ecdsa and adds its public key to authorized_keys. Could we reuse that identity and verify SSH connectivity?

e2e/acceptance/internal/sharedsteps/login_user_isolation.go
```diff
@@ -0,0 +1,500 @@
...
+			framework.ShellQuote(framework.BashLC(setupIdentity)),
+		)
+		if _, err := s.runInLoginJail(ctx, command); err != nil {
+			return fmt.Errorf("prepare login CPU isolation SSH identity for %s: %w", userName, err)
+		}
+	}
+	return nil
+}
+
+func (s *LoginUserIsolation) removeSSHIdentities(ctx context.Context) error {
```
ChessProfessor: Reusing the existing internal key would remove the need for this cleanup

e2e/acceptance/internal/sharedsteps/login_user_isolation.go
```diff
@@ -0,0 +1,500 @@
...
+
+	work, err := waitForLoginCPUWorkloads(ctx, results)
+	if err != nil {
+		return loginCPUWork{}, err
+	}
+	s.workloadActive = false
+	return work, nil
+}
+
+func (s *LoginUserIsolation) runInLoginContainer(ctx context.Context, script string) (string, error) {
```
ChessProfessor: Could we replace these three wrappers and pod discovery with runtime.Jail().Run() and remove the extra chroot from the workload command?

e2e/acceptance/internal/sharedsteps/login_user_isolation.go
```diff
@@ -0,0 +1,500 @@
...
+		}
+	}
+	if err := errors.Join(workloadErrors...); err != nil {
+		return loginCPUWork{}, err
+	}
+	return work, nil
+}
+
+func loginCPUWorkloadCommand(userName string, burners int) string {
+	burnScript := fmt.Sprintf(`
```
ChessProfessor, line 303: To be honest I don't like adding such bash scripts but I also don't know how to replace it with something simpler. If you have any ideas please tell me, otherwise just ignore this message :)

rdjjke (dismissed):

internal/values/login_user_isolation.go
```diff
@@ -7,8 +7,8 @@ import (
 )
 
 const (
-	loginUserIsolationMemoryHighPercent int64 = 80
-	loginUserIsolationMemoryMaxPercent  int64 = 90
+	loginUserIsolationMemoryHighPercent int64 = 40
+	loginUserIsolationMemoryMaxPercent  int64 = 50
```
aydinmirac, line 11: Non-blocking comment: I realized that partial overrides are sensitive to the derived defaults. I didn’t catch this in my original implementation because my test value happened to remain below the derived `memoryMax` of 90%, so the configuration was valid and the edge case didn’t draw my attention. 

With the new defaults, an sshd container with 10 GiB and `memoryHigh`: 6Gi would derive `memoryMax`: 5Gi, so the validating webhook would reject the configuration.

Would it be clearer to require `memoryHigh` and `memoryMax` to be specified together, or both omitted? We could enforce this in the Soperator API and optionally add matching validation in `nebius-solutions-library` for an earlier, friendlier Terraform error.

ChessProfessor:
E2e part looks good, thank you!

internal/values/login_user_isolation.go
Ronsenval: Added a CRD validation to enforce both settings to be presented or omitted in the same time https://github.com/nebius/soperator/pull/2938/commits/c52c91086975444ddd694f9232837c6fa880dfa1

rdjjke (approved):

PR #2940, opened 2026-09-07
SCHED-1332 Add system settings acceptance coverage

ChessProfessor:
## Problem

Soperator acceptance tests do not cover the expected system settings on login and worker nodes or inside native and Enroot Slurm jobs. Regressions in resource limits, nested Bash behavior, kernel settings, or NVIDIA driver capabilities can therefore go unnoticed.

## Solution
 
Add acceptance scenarios for:

- Login and worker resource limits, including nested Bash sessions
- Native and Enroot Slurm job resource limits
- Required kernel settings in SSH sessions and Slurm jobs
- NVIDIA driver capabilities on GPU workers

The resource-limit scenarios affected by the known SCHED-2442 behavior are marked @unstable until the product behavior is fixed. All scenarios are gated to Soperator 4.0.0 or newer.

### DIFF
--- a/e2e/acceptance/features/system_settings.feature
+++ b/e2e/acceptance/features/system_settings.feature
@@ -1,26 +1,26 @@
 Feature: System settings
 
   @soperator_version_>=4.0.0
-  Scenario: Login SSH sessions retain the default resource limits in nested Bash
+  Scenario: Login SSH sessions expose the default resource-limit profile
     When root collects direct and nested Bash resource limits over SSH to the login node
     Then direct and nested login SSH resource limits match the expected default profile
 
   # SCHED-2442: remove @unstable after root worker SSH and Slurm jobs receive the compute limit profile.
   @unstable @soperator_version_>=4.0.0
-  Scenario: Worker SSH sessions expose fully configured resource limits
+  Scenario: Worker SSH sessions expose the compute resource-limit profile
     Given a healthy worker is selected for system-settings checks
     When root collects direct and nested Bash resource limits over SSH to the worker
     Then direct and nested worker SSH resource limits match the compute profile
 
   @unstable @soperator_version_>=4.0.0
-  Scenario: Native Slurm jobs expose fully configured resource limits
+  Scenario: Native Slurm jobs expose the compute resource-limit profile
     Given a healthy worker is selected for system-settings checks
     When a native Slurm job collects direct and nested Bash resource limits
     Then the native system-settings job succeeds
     And its direct and nested resource limits match the compute profile
 
   @unstable @soperator_version_>=4.0.0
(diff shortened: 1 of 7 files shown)


e2e/acceptance/framework/kubectl_client_test.go
```diff
@@ -188,6 +188,54 @@ func TestKubectlClientWorkerPodsRejectsDuplicateHostname(t *testing.T) {
...
 }
 
+func TestKubectlClientWorkerContainerEnvironmentVariable(t *testing.T) {
+	exec := &kubectlClientTestExec{kubectl: map[string]string{
+		"get\x00pod\x00worker-0\x00-n\x00soperator\x00-o\x00json": `{
+			"metadata": {"name": "worker-0"},
+			"spec": {
+				"containers": [
+					{"name": "slurmd", "env": [
+						{"name": "NVIDIA_DRIVER_CAPABILITIES", "value": "compute,graphics,utility,video"}
```
asteny, line 198: I am not sure how that should work, but you are testing Nebius terraform overwrides from terrform repo. Because soperator have different defaults https://github.com/nebius/soperator/blob/main/helm/nodesets/values.yaml#L147 Is it ok?

ChessProfessor: We create clusters for e2e using terraform, we have no other ways now, so e2e will be coupled with terraform values
I can update defaults for these values in soperator now, but there is much more drift between these two repos, are we planning to do smth with it?
@rdjjke wdyt?

ChessProfessor: Misha says terraform repo might be close-sourced one day, then we will need to find a way to create clusters without using it (or close-source e2e too and then soperator too)
But until we decouple e2e and terraform repo this issue will persist

e2e/acceptance/features/system_settings.feature
```diff
@@ -0,0 +1,41 @@
+Feature: System settings
+
+  @soperator_version_>=4.0.0
+  Scenario: Login SSH sessions retain the default resource limits in nested Bash
```
rdjjke: Since both scenarios are groped into one, I would suggest to remove "in nested Bash" from the scenario title.

rdjjke: The next scenario "Worker SSH sessions expose fully configured resource limits" doesn't have it already, btw

e2e/acceptance/framework/kubectl_client_test.go
rdjjke: Well, these E2E are testing Nebius TF in fact, so I think it's fine.

rdjjke (approved):

PR #2955, opened 2026-09-09
check_runner: do not quote scontrol Reason= and Comment= values

### LINKED ISSUE #2954: check_runner.py stores scontrol Reason/Comment with literal quotes, so on_ok undrain/uncomment never fires
`drain_node`, `comment_node` and `uncomment_node` in `helm/slurm-cluster/slurm_scripts/check_runner.py` build the scontrol argv as

    f'Reason="{reason}"'
    f'Comment="{comment}"'
    'Comment=""'

`subprocess.run` gets a list, there is no shell to strip the quotes, so Slurm stores them as part of the value. On a node marked by a check, `scontrol show node --json` returns

    "comment": "\"[node_problem] ib_link_check: ... [hc_program]\""

On the OK path `run_check` checks `comment.startswith(reason_base)` (same for `reason` and undrain). The stored value starts with `"`, the base with `[`, so it never matches and `on_ok: uncomment` / `on_ok: undrain` never clear the mark. `Comment=""` does not clear the field either, it stores the two-character string `""`.

Seen in production: a node kept its `[node_problem] ib_link_check` comment after the IB link came back and the check passed. With the quotes removed, the next hc_program run logged `Uncomment Slurm node ...` and the comment was cleared.

Fix: drop the quotes — `Reason={reason}`, `Comment={comment}`, `Comment=`.

dnugmanov:
## Problem

The scontrol calls in `drain_node`, `comment_node` and `uncomment_node` pass a list to `subprocess.run`, so the quotes around the value are stored in Slurm as part of it. `on_ok: undrain` / `uncomment` compare with `startswith(reason_base)` and never match a value that starts with `"`, and `Comment=""` stores `""` instead of clearing the field. Nodes keep their `[node_problem] ...` reason and comment after the check starts passing.

## Solution

Drop the quotes from the three scontrol arguments: `Reason={reason}`, `Comment={comment}`, `Comment=`.

## Testing

On a production cluster a node kept its `[node_problem] ib_link_check` comment after the IB link came back and the check passed on every run. With the quotes removed, the next hc_program run logged `Check ib_link_check: OK` and `Uncomment Slurm node worker-gpu-29`, and the comment was cleared.

## Release Notes

Fix: `on_ok: undrain` and `on_ok: uncomment` now clear the drain reason and node comment set by a check, instead of leaving the node marked forever.

Fixes #2954.

### DIFF
--- a/helm/slurm-cluster/slurm_scripts/check_runner.py
+++ b/helm/slurm-cluster/slurm_scripts/check_runner.py
@@ -565,7 +565,7 @@ def drain_node(reason):
     try:
 
         subprocess.run(
-            ["scontrol", "update", f"NodeName={SLURMD_NODENAME}", "State=drain", f"Reason=\"{reason}\""],
+            ["scontrol", "update", f"NodeName={SLURMD_NODENAME}", "State=drain", f"Reason={reason}"],
             check=False, stderr=subprocess.DEVNULL
         )
         # Invalidate cache for the Slurm node info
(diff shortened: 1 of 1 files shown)


asteny (approved):

asteny:
@dnugmanov Thanks for your contribution!

PR #2968, opened 2026-09-10
SCHED-2492: add essential E2E scenarios

ChessProfessor:
## Problem

PR-triggered E2E runs always execute the complete acceptance suite, even when only fast validation of release-critical behavior is needed. In the measured run, acceptance tests took approximately 53 minutes, making feedback from /e2e unnecessarily slow for many changes.

## Solution

Added an @essential tag to a curated set of critical acceptance scenarios covering cluster readiness, workloads, accounting, health checks, system settings, and topology.

Added /e2e-essential and a corresponding E2E workflow flag to run only these scenarios. The existing /e2e command continues to run the full stable suite unchanged. The essential subset is estimated to take approximately 12 minutes based on current scenario timings.

### DIFF
--- a/e2e/acceptance/features/docker_containers.feature
+++ b/e2e/acceptance/features/docker_containers.feature
@@ -1,5 +1,5 @@
 Feature: Docker containers
-  @essential @soperator_version_>=5.0.0
+  @soperator_version_>=5.0.0
   Scenario: Docker container lifecycle uses local storage
     Given a long-running Docker container job is submitted on two workers
     When the Docker container job is running
--- a/e2e/acceptance/features/docker_containers.feature
+++ b/e2e/acceptance/features/docker_containers.feature
@@ -1,5 +1,5 @@
 Feature: Docker containers
-  @soperator_version_>=5.0.0
+  @essential @soperator_version_>=5.0.0
   Scenario: Docker container lifecycle uses local storage
     Given a long-running Docker container job is submitted on two workers
     When the Docker container job is running
--- a/e2e/acceptance/features/enroot_containers.feature
+++ b/e2e/acceptance/features/enroot_containers.feature
@@ -11,7 +11,7 @@ Feature: Enroot containers
       | login  |
       | worker |
 
-  @essential @soperator_version_>=4.0.0
+  @soperator_version_>=4.0.0
   Scenario: Enroot and Pyxis cache images and clean up runtime state
     Given a long-running Enroot container job is submitted on two workers
     When the Enroot container job is running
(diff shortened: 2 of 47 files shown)


rdjjke:
I would make some changes to which scenarios are considered essential.
But by "essential" I understand the scope with the "most important scenarios that must always work", not "the scenarios that take short".
See my other comments for details.

e2e/acceptance/features/docker_containers.feature
```diff
@@ -1,5 +1,5 @@
 Feature: Docker containers
-  @soperator_version_>=5.0.0
+  @essential @soperator_version_>=5.0.0
```
rdjjke: I don't think that Docker is an essential scenario, because it's much less popular than Enroot

e2e/acceptance/features/enroot_containers.feature
```diff
@@ -11,7 +11,7 @@ Feature: Enroot containers
       | login  |
       | worker |
 
-  @soperator_version_>=4.0.0
+  @essential @soperator_version_>=4.0.0
```
rdjjke: Maybe Enroot image caching is not essential as well, because even without it, the cluster is still usable.

e2e/acceptance/features/login_user_isolation.feature
```diff
@@ -1,5 +1,5 @@
 Feature: Login user isolation
-  @soperator_version_>=5.0.0
+  @essential @soperator_version_>=5.0.0
```
rdjjke: Not essential IMO

e2e/acceptance/features/nodeset_ephemeral_mode_transition.feature
```diff
@@ -1,5 +1,5 @@
 Feature: NodeSet ephemeral mode transitions
-  @soperator_version_>=5.0.0
+  @essential @soperator_version_>=5.0.0
```
rdjjke: Even if it's broken, it's not so critical, so it's not essential IMO

e2e/acceptance/features/observability.feature
```diff
@@ -1,4 +1,4 @@
 Feature: Observability stack
-  @soperator_version_>=5.0.0
+  @essential @soperator_version_>=5.0.0
```
rdjjke: Not essential IMO. But there could be other o11y scenarios in the future that are essential

e2e/acceptance/features/system_settings.feature
```diff
@@ -1,6 +1,6 @@
 Feature: System settings
 
-  @soperator_version_>=4.0.0
+  @essential @soperator_version_>=4.0.0
```
rdjjke: I think the login scenario is not essential, but the "Native Slurm jobs expose the compute resource-limit profile" is essential

rdjjke:
I think the scenario "A maintenance event replaces the selected worker node" is essential because if it's broken, then we are cooked.

rdjjke (approved):

PR #2983, opened 2026-09-15
SCHED-1907: Replace SPO with provisioned AppArmor profiles

theyoprst:
## Problem

Soperator depends on SPO to distribute a static AppArmor policy, adding a controller and per-node daemons.

## Solution

Remove SPO reconciliation, deployment configuration, RBAC, and dependencies. Remove `useDefaultAppArmorProfile` without a replacement cluster-wide setting. Containers use their existing per-container `appArmorProfile` fields, which default to `unconfined` in the CRDs and Helm charts.

The paired [Terraform change](https://github.com/nebius/nebius-solutions-library/pull/1181) loads the unchanged policy as `soperator-default` through cloud-init and sets that name directly for login sshd and worker slurmd containers. Other containers keep their own profiles.

## Testing

Provisioned the cluster, checked that the apparmor profile is still there.

## Release Notes

Breaking: `useDefaultAppArmorProfile` is removed. Configure `SlurmCluster.spec.slurmNodes.login.sshd.appArmorProfile` and `NodeSet.spec.slurmd.security.appArmorProfile` directly; both default to `unconfined`. Provisioning must load any selected named profile on the nodes before the containers start.


api/v1/slurmcluster_types.go
```diff
@@ -115,7 +115,8 @@ type SlurmClusterSpec struct {
 	// +kubebuilder:validation:Optional
 	SConfigController SConfigController `json:"sConfigController,omitempty"`
 
-	// Generate and set default AppArmor profile for the login nodes. The Security Profiles Operator must be installed.
+	// Use the soperator-default AppArmor profile for login and worker containers.
+	// Node provisioning must load this profile before the containers start.
 	//
 	// +kubebuilder:default=false
 	UseDefaultAppArmorProfile bool `json:"useDefaultAppArmorProfile,omitempty"`
```
Uburro: i would say better rename this field and propagate the apparmorprofile name as is for login and worker pods.

theyoprst: We still have this for login and worker pods:

```
	// +kubebuilder:default="unconfined"
	AppArmorProfile string `json:"appArmorProfile,omitempty"`
```

Should we keep it as overrides?
If so, I suggest changing their default to empty ("inherit") and replacing `UseDefaultAppArmorProfile bool` with `DefaultAppArmorProfile string`, defaulting to "unconfined". Terraform would set it to "soperator-default". An explicit per-container value would take precedence.

Does it make sense?

Uburro: yes, i am agree

Uburro: I mean that `AppArmorProfile: unconfined` would be a good default value. Later, we can change it to the required value in Terraform.

theyoprst: Done

rdjjke (approved):

Uburro (approved):

PR #3000, opened 2026-09-18
SCHED-2379 Remove "soperator-" prefix from login node names

asteny:


e2e/acceptance/framework/kubectl_client.go
```diff
@@ -66,6 +66,47 @@ func (c *KubectlClient) SlurmCluster(ctx context.Context, name string) (SlurmClu
...
+		return "", fmt.Errorf("no Ready %s pods found for SlurmCluster %s", component, clusterName)
+	}
+
+	sort.Strings(readyPods)
+	ordinalZeroSuffix := component + "-0"
+	for _, podName := range readyPods {
+		if podName == ordinalZeroSuffix || strings.HasSuffix(podName, "-"+ordinalZeroSuffix) {
+			return podName, nil
+		}
+	}
```
ChessProfessor: Why do we need that? Doesn't `return readyPods[0]` always return the same value after sorting?

e2e/acceptance/runner.go
```diff
@@ -350,11 +349,11 @@ func NewRuntime(kubectlContext, slurmClusterName, targetSoperatorVersion string)
 	if targetSoperatorVersion == "" {
 		return nil, fmt.Errorf("target Soperator version is required")
 	}
-	normalizedVersion, err := framework.NormalizeSoperatorVersion(targetSoperatorVersion)
+	_, err := framework.NormalizeSoperatorVersion(targetSoperatorVersion)
```
ChessProfessor: Why do we need to keep this check if we never use soperator version later?
Let's get rid of `targetSoperatorVersion` in `NewRuntime`

e2e/acceptance/world.go
```diff
@@ -51,13 +50,21 @@ func (w *world) Local() framework.ArgsScope {
 
 func (w *world) Controller() framework.CommandScope {
 	return framework.NewCommandScope(func(ctx context.Context, command string) (string, error) {
-		return w.Kubectl().Run(ctx, "exec", "-n", framework.SoperatorNamespace, framework.SoperatorPodName(w.slurmClusterName, w.soperatorVersion, "controller-0"), "--", "bash", "-lc", command)
+		podName, err := framework.NewKubectlClient(w).ReadyWorkloadPodName(ctx, w.slurmClusterName, "controller")
```
ChessProfessor: Why not `w.Kubectl().ReadyWorkloadPodName()`?
We can also calculate `ReadyWorkloadPodName()` only once outside the lambda and use it inside, it doesn't change for one cluster.

e2e/acceptance/world.go
```diff
@@ -51,13 +50,21 @@ func (w *world) Local() framework.ArgsScope {
...
+			return "", err
+		}
+		return w.Kubectl().Run(ctx, "exec", "-n", framework.SoperatorNamespace, podName, "--", "bash", "-lc", command)
 	})
 }
 
 func (w *world) Jail() framework.CommandScope {
 	return framework.NewCommandScope(func(ctx context.Context, command string) (string, error) {
-		return w.Kubectl().Run(ctx, "exec", "-n", framework.SoperatorNamespace, framework.SoperatorPodName(w.slurmClusterName, w.soperatorVersion, "login-0"), "--", "chroot", "/mnt/jail", "bash", "-lc", command)
+		podName, err := framework.NewKubectlClient(w).ReadyWorkloadPodName(ctx, w.slurmClusterName, "login")
```
ChessProfessor: The same here?

api/v1/slurmcluster_types.go
```diff
@@ -13,13 +13,32 @@ import (
 	prometheusv1 "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
 )
 
+// WorkloadNamePrefixMode controls whether workload names include the SlurmCluster name.
+type WorkloadNamePrefixMode string
```
rdjjke: Why do you call it "workload" and not "worker"?

rdjjke: Ah, because "It applies to controller, accounting, login, REST, exporter, and sconfigcontroller workloads."

Then, maybe we can call it "PodNamePrefixMode"? Because it's not clear what "workload" is - this term isn't used anywhere else.

e2e/acceptance/world.go
asteny: w.Kubectl() returns an ArgsScope, so ReadyWorkloadPodName is not available on it. I now create a single KubectlClient when the world is initialized and reuse it. Pod lookup remains lazy because the pod may not be ready during initialization, but a successful result is cached and reused. Lookup errors are not cached, so retries continue to work while the pod is starting.

ChessProfessor:
E2e part looks good, thank you!

e2e/acceptance/world.go
```diff
@@ -22,7 +23,10 @@ type world struct {
 
 	kubectlContext   string
 	slurmClusterName string
-	soperatorVersion string
+	kubectlClient    *framework.KubectlClient
```
ChessProfessor: nit: Maybe let's not store it here? Just resolve and store `loginPodName` and `controllerPodName` in `newWorld()`? And fail `.Controller()` and `.Login()` methods when these are undefined.
Otherwise `KubectlClient` depends on `world` and `world` depends on `KubectlClient`.

ChessProfessor (approved):

rdjjke (approved):

PR #3004, opened 2026-09-18
SCHED-2553: Use regional ingester for logs by default

yozel:
Use regional ingester for logs by default


helm/soperator-fluxcd/templates/_helpers.tpl
```diff
@@ -62,3 +62,8 @@ Validate observability.publicEndpointTokenKind is one of: secret, hostPath
   {{- fail (printf "observability.publicEndpointTokenKind must be one of: secret, hostPath (got %q)" $kind) -}}
 {{- end -}}
 {{- end -}}
+
+{{/* Resolve public logging from the region unless an endpoint is explicitly configured. */}}
+{{- define "soperator-fluxcd.loggingEndpoint" -}}
+{{- .Values.observability.opentelemetry.publicEndpoint | default (printf "dns:///write.logging.%s.nebius.cloud.:443" .Values.observability.region) -}}
```
Uburro, line 68: i do not think it's a good idea. We have a plan to move values ​​from the templates into `values.yaml`, and this is just make this movement harder

docs/logs-pipeline.md
```diff
@@ -150,8 +150,10 @@ The logging system automatically extracts metadata from filenames and creates th
 observability:
   # Cloud delivery
   publicEndpointEnabled: true  # Enable/disable cloud export
-  projectId: "your-nebius-project-id"
-  region: "eu-north1"
+  logsProjectId: "your-nebius-project-id"
+  region: "eu-west2"
```
Uburro: why not a "region-id" :) ?

helm/soperator-fluxcd/templates/_helpers.tpl
yozel: I can't think of another way to use region for log ingestion then. Any ideas?

docs/logs-pipeline.md
yozel: Fixing it :)

PR #3013, opened 2026-09-21
SCHED-886: Add optional pam_slurm_adopt support

aydinmirac:
## Problem
TBD


images/worker/slurmd.dockerfile
```diff
@@ -41,7 +43,8 @@ RUN apt-get update && \
         kmod \
         libncurses5-dev \
         supervisor \
-        openssh-server && \
+        openssh-server \
+        slurm-smd-libpam-slurm-adopt=${SLURM_DEB_VERSION} && \
```
rdjjke, line 47: Isn't this package already installed in the base image?

Here is where Slurm deb packages are built -- https://github.com/nebius/slurm-deb-packages/blob/main/slurm-packages/Dockerfile#L68
Here is how the packages are installed in the base "slurm" image -- https://github.com/nebius/ml-containers/blob/main/ansible/roles/slurm/tasks/main.yml#L27

@asteny mb you can comment on this -- what is the preferred way to install this slurm package?

api/v1/slurmcluster_types.go
```diff
@@ -140,6 +146,27 @@ type SlurmClusterSpec struct {
 	HealthCheckConfig *HealthCheckConfig `json:"healthCheckConfig,omitempty"`
 }
 
+// PAMSlurmAdopt defines access and adoption policy for SSH sessions on worker nodes.
+type PAMSlurmAdopt struct {
+	// Enabled requires non-root SSH users to own a running job on the target worker.
+	//
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:default=true
```
rdjjke, line 154: I think it should be false by default. Some users want it to be enabled, but I can't confidently say it's true for most users

rdjjke: But maybe it's worth setting it to true for clusters we create in E2E tests

e2e/acceptance/features/pam_slurm_adopt.feature
```diff
@@ -0,0 +1,13 @@
+Feature: PAM Slurm adopt
+
+  @gpu @unstable @soperator_version_>=5.0.0
```
ChessProfessor, line 3: Why do we add @unstable tag by default? Do we expect it to fail?

e2e/acceptance/features/internal_ssh.feature
```diff
@@ -1,6 +1,7 @@
 Feature: Internal SSH
   @essential @soperator_version_>=4.0.0
-  Scenario: A regular user can SSH to a worker without extra options
+  Scenario: A regular user with a job can SSH to a worker without extra options
```
ChessProfessor, line 3: I suggest splitting this onto two scenarios based on version. Keep the one that we had before (initial version) with `@soperator_version_>=4.0.0,<5.0.0`, and add a new one (current version) with `@soperator_version_>=5.0.0`, wdyt?

e2e/acceptance/internal/sharedsteps/docker_containers.go
```diff
@@ -251,6 +251,17 @@ func (s *DockerContainers) theUserRunsADockerContainerOverSSHOnAWorkerNode(ctx c
 	if err := waitForSSHTestUserOnWorker(ctx, s.runtime, dockerSSHUserName, worker); err != nil {
 		return err
 	}
+	job, err := startSSHAccessJob(
```
ChessProfessor, line 254: So everywhere where we need to ssh to worker we will need to submit ssh job before doing so? It seems like huge overcomplication for every test that we develop. Do we have some way to create separate acceptance user on a cluster and setup access from it to the workers without submitted job? Or at least we should have some kind of helper for ssh that does job submitting in background without the test itself submitting anything, but that's quite risky approach, it may interfere with other acceptance load in general.

e2e/acceptance/internal/sharedsteps/enroot_containers.go
```diff
@@ -82,22 +82,14 @@ func (s *EnrootContainers) RegisterSteps(sc *godog.ScenarioContext) {
...
-		s.runtime.Logf("cleanup: cancel enroot job: %v", cleanupErr)
-	}
-	s.workers = nil
-	s.connectionWorker = framework.WorkerInfo{}
-	s.job = framework.SbatchJob{}
-	s.squashPath = ""
-	s.expectedSquashPath = ""
-	s.squashStatBefore = ""
-	s.runtimeNamePrefix = ""
-	s.directSquashFS = nil
```
ChessProfessor, line 95: What has changed here, why do we need this?

ChessProfessor (changes requested):

images/worker/slurmd.dockerfile
asteny: Misha, you are right we need to adjust this list of packages, build the new image and use it here

aydinmirac: I checked the exact base image used by the worker:

`cr.nebius.cloud/ml-containers/slurm:26.05.4-nebius-1-20260918141513
`
The installed Slurm packages are:

```bash
$ apt list --installed | grep -i slurm

WARNING: apt does not have a stable CLI interface. Use with caution in scripts.

slurm-smd-client/noble,now 26.05.4-nebius-1 amd64 [installed]
slurm-smd-dev/noble,now 26.05.4-nebius-1 amd64 [installed]
slurm-smd-libnss-slurm/noble,now 26.05.4-nebius-1 amd64 [installed]
slurm-smd-libpam-slurm-adopt/noble,now 26.05.4-nebius-1 amd64 [installed]
slurm-smd-slurmctld/noble,now 26.05.4-nebius-1 amd64 [installed]
slurm-smd-slurmd/noble,now 26.05.4-nebius-1 amd64 [installed]
slurm-smd-slurmdbd/noble,now 26.05.4-nebius-1 amd64 [installed]
slurm-smd-slurmrestd/noble,now 26.05.4-nebius-1 amd64 [installed]
slurm-smd/noble,now 26.05.4-nebius-1 amd64 [installed]
```

I also checked the PAM package directly:

```bash
$ dpkg -l | grep -i slurm
ii  slurm-smd                        26.05.4-nebius-1                          amd64        Slurm Workload Manager
ii  slurm-smd-client                 26.05.4-nebius-1                          amd64        Slurm Workload Manager client commands.
ii  slurm-smd-dev                    26.05.4-nebius-1                          amd64        Development package for Slurm-smd.
ii  slurm-smd-libnss-slurm           26.05.4-nebius-1                          amd64        Slurm NSS plugin
ii  slurm-smd-libpam-slurm-adopt     26.05.4-nebius-1                          amd64        Slurm's pam_slurm_adopt module
ii  slurm-smd-slurmctld              26.05.4-nebius-1                          amd64        Slurm controller daemon.
ii  slurm-smd-slurmd                 26.05.4-nebius-1                          amd64        Slurm compute node daemon
ii  slurm-smd-slurmdbd               26.05.4-nebius-1                          amd64        Slurm database daemon.
ii  slurm-smd-slurmrestd             26.05.4-nebius-1                          amd64        Slurm REST API daemon
```

Therefore, `slurm-smd-libpam-slurm-adopt` is produced by the package repository but is not installed in the current base image. When installed, the module is placed under the PAM module directory:

`/usr/lib/x86_64-linux-gnu/security/pam_slurm_adopt.so`

It is not expected under `/usr/lib/x86_64-linux-gnu/slurm/`.

Since only worker SSH uses this module, should we keep the version-pinned installation in worker/slurmd.dockerfile or should we open a PR in other repositories?

asteny: @aydinmirac Somebody need to create PR to ml-containers repo to add package which you need to this list https://github.com/nebius/ml-containers/blob/main/ansible/inventory/group_vars/all.yaml#L5 After that we can build new base image with Slurm components

rdjjke: Yes, I was wrong, the package is not yet installed. But it would be more conventional to install it to the base Slurm image, as Pasha suggested above

aydinmirac: Thanks, understood. I will follow your instructions then. I will create a PR adding `slurm-smd-libpam-slurm-adopt` and fix the dockerfile here when the PR is merged.

